PR #7, opened 2024-05-14
add ci-tooling and config files

stavros-k:
Supersedes #4 

- removed the need of the additional go tool for checking container's health.
- now for each test file in each changed app, a new job will spin. that has some benefits:
  - parallel runs, (no more waiting an hour or even more for apps like minecraft or immich)
  - clean workspace for each test (so we can avoid parsing compose files to identify host paths used to clean them before the next run)

> - There are 2 python scripts that are probably hideous to look at. But I figured python would be easier for others to chime in and improve things later.
> - I've placed a max of 10 jobs in parallel to avoid hammering github :D. 

Currently the ci.py script does 2 modifications during its runtime:
- Copies the top-level library to the app-level based on the app.yaml `lib_version` field.
- Copies the top-level macros to the app-level.

If we add those 2 steps in a release script we might want to extract them and re-use them


stavros-k:
Please if anyone got a spare moment and want to take over/refactor this. Please do :)
I'm sure there are a ton of better ways to do all this with python-isms

.github/scripts/ci.py
```diff
@@ -0,0 +1,346 @@
...
+        if not line:
+            continue
+        try:
+            failed_containers.append(json.loads(line))
+        except json.JSONDecodeError:
+            print_stderr(f"Failed to parse container status output:\n {line}")
+            sys.exit(1)
+
+    actual_failed = []
+    for container in failed_containers:
```
Qubad786, line 203: Are we sure they won't be running at this point?

.github/scripts/ci.py
```diff
@@ -0,0 +1,346 @@
...
+        # Skip containers that are exited with 0 (eg init containers), but not restarting (during a restart exit code is 0)
+        if all(
```
Qubad786, line 205: Why use `all` ?

.vscode/extensions.json
```diff
@@ -0,0 +1,8 @@
+{
```
Qubad786, line 1: We should add these to gitignore ?

.github/scripts/ci.py
```diff
@@ -0,0 +1,346 @@
...
+    separator_start()
+    subprocess.run(cmd, shell=True)
+    separator_end()
+
+
+def get_failed_containers():
+    # Outputs one container per line, in json format
+    cmd = f"{get_base_cmd()} ps --all --format json"
+    print_cmd(cmd)
+    failed = subprocess.run(cmd, shell=True, capture_output=True).stdout.decode("utf-8")
```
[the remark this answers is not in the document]
stavros-k: Initially I had a --filter, but that wasn't very effective. So failed was making sense before. Forgot to update it! Fixing now!

.github/scripts/ci.py
stavros-k: docker compose ps -all will show all containers, both running and exited, stopped, completed, etc.

I observed some empty lines sometimes, which is why I skipping those.

.github/scripts/ci.py
stavros-k: So, to mark a container healthy we need to considering the following:

1. If its exited, the exit code is 0 and the healthcheck is empty. (example here is an init container that runs sometihng and then exits.)
2. It's state is not restarting. (At this point the Health state can have a misleading value)
3. And if there is a healthcheck, the state is healthy.

If any of the above is not true, we consider the container unhealthy.

.vscode/extensions.json
stavros-k: I don't have any hard opinions here.
I just added them because I found setting vscode for jinja with docker a bit annoying.

I do wanna try pycharm though, see if that's better in terms of syntax highlighting, etc.

.github/scripts/ci.py
stavros-k: Most of if not all logic in this function would be nuked if there wasn't this issue
https://github.com/docker/compose/issues/10596

Qubad786: I think that we should name the variable appropriately as they are not failed containers at that point?

.github/scripts/ci.py
Qubad786: There is no need to use `all` as we are not iterating over anything?
We can just do `if c[cond1] and c[cond2] and c[cond3]` ?

.vscode/extensions.json
Qubad786: IDE specific settings should be in gitignore i think

.github/scripts/ci.py
Qubad786: If you want to chain conditions you can do
```
if (cond1 or cond1 != something) and (cond2 == 'somethingelse' or cond2 == cond1):
```

.vscode/extensions.json
stavros-k: Fair, Removed!

.github/scripts/ci.py
stavros-k: Fixed!

.github/scripts/ci.py
stavros-k: Fixed!

Qubad786 (approved):
Thanks!

PR #8, opened 2024-05-14
Add minio enterprise

stavros-k:
Supersedes #6


Qubad786 (approved):
This LGTM but this is being merged into `ci-fixs`, can you clarify what we are doing there?

stavros-k:
> This LGTM but this is being merged into `ci-fixs`, can you clarify what we are doing there?

I've had rebased it to that branch and changed the target to it for cleaner diff.
Now that `ci-fixs` is merged it points to `master`

PR #9, opened 2024-05-14
add library

stavros-k:


Qubad786 (commented):
@stavros-k let's not use mutable objects as default values for args as that leads to unintended behavior

Qubad786:
@stavros-k is this up for review again?

Qubad786 (approved):

PR #14, opened 2024-06-10
Add prowlarr

stavros-k:


ix-dev/community/prowlarr/questions.yaml
```diff
@@ -0,0 +1,340 @@
...
+    schema:
+      type: dict
+      attrs:
+        - variable: web_port
+          label: WebUI Port
+          description: The port for Prowlarr WebUI
+          schema:
+            type: int
+            default: 30027
+            min: 9000 # TODO: do we still need this?
```
kmoore134, line 86: I would assume 9000 is no longer a restriction. We should probably lift it and only restrict specific ports we already know are in use by TrueNAS out of box. I.E. 443/80, etc.

stavros-k: Yea that's what I thought, I should probably update the type here and use the new `port` $ref, that will filter used ports too.

Qubad786 (approved):

PR #18, opened 2024-06-12
add readarr

stavros-k:


Qubad786 (commented):
We have conflicts here and i think we should also factor in the ix values changes here considering they are already being made for other apps?

stavros-k:
> We have conflicts here and i think we should also factor in the ix values changes here considering they are already being made for other apps?

Yes, will update this and the other apps I've shared that are ready for review.
Will ping you when those are ready again shortly!

Qubad786 (approved):

PR #27, opened 2024-06-18
add actual budget

stavros-k:


ix-dev/community/actual-budget/ix_values.yaml
```diff
@@ -0,0 +1,5 @@
+consts:
```
Qubad786, line 1: We don't have image defined in ix_values for this ?

stavros-k: Whoops, thanks for catching that!

Qubad786 (approved):

PR #46, opened 2024-07-03
NAS-129729 / 24.10 / Properly mark github action as success if nothing has been published

Qubad786:
## Problem

When there are no published apps, update catalog action would error out because it still tries to commit (part of it still commits where only date hashes are changed and published)

## Solution

Skip the commit generation if nothing has changed in the repository.

### DIFF
--- a/.github/workflows/update_catalog.yaml
+++ b/.github/workflows/update_catalog.yaml
@@ -25,16 +25,22 @@ jobs:
       - name: Publish catalog
         run: |
           /bin/bash -c "PWD=${pwd}; /usr/local/bin/apps_catalog_update publish --path $PWD"
+      - name: Check untracked files existence
+        run: echo "CHANGES=$(git -C ${pwd} --no-pager status --porcelain | wc -l)" >> "$GITHUB_ENV"
+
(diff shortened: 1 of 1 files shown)


.github/workflows/update_catalog.yaml
```diff
@@ -25,16 +25,24 @@ jobs:
       - name: Publish catalog
         run: |
           /bin/bash -c "PWD=${pwd}; /usr/local/bin/apps_catalog_update publish --path $PWD"
+      - name: Check untrack files exist or not
+        id: untrack_files
+        run: |
+          echo ::save-state name=exit_code::$(git -C $PWD --no-pager status --porcelain | wc -l)
```
stavros-k, line 31:
```suggestion
          echo "exit_code=$(git -C $PWD --no-pager status --porcelain | wc -l)" >> $GITHUB_STATE
```

Should we update it to this?

https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/#examples

Also can't remember if `$GITHUB_STATE` is reflected on the `.outputs`, but `$GITHUB_OUTPUT` should be.

stavros-k (approved):

PR #62, opened 2024-07-09
add fscrawler

stavros-k:


ix-dev/community/fscrawler/app.yaml
```diff
@@ -0,0 +1,33 @@
+app_version: '2.9'
```
Qubad786, line 1: App version does not really reflect the tag we are using - is that intended?

stavros-k, line 1:
```suggestion
app_version: '2.10'
```

stavros-k: Fixed!

Qubad786 (approved):

PR #75, opened 2024-07-16
add node-red

stavros-k:


ix-dev/community/node-red/app.yaml
```diff
@@ -0,0 +1,32 @@
+app_version: 3.1.10
```
Qubad786, line 1: Just curious how the app version is being deduced here

Qubad786 (approved):

ix-dev/community/node-red/app.yaml
stavros-k: If I recall correctly the "default" image was with node 16, and we used that as a "base" to get the app_version.

PR #77, opened 2024-07-16
add netbootxyz

stavros-k:


Qubad786 (approved):

ix-dev/community/netbootxyz/questions.yaml
```diff
@@ -0,0 +1,450 @@
...
+            required: true
+            $ref:
+              - definitions/port
+        - variable: host_network
+          label: Host Network
+          description: |
+            Bind to the host network. It's recommended to keep this disabled.
+          schema:
+            type: boolean
+            default: true
```
rehevkor5, line 127: It's a bit odd for the default to be `true` when the description says:

> It's recommended to keep this disabled.

Was that intentional? (Disabling it seems more secure anyway, since you must explicitly list the ports that the bridge should map... but maybe netbootxyz has some kind of issue in that situation?)

PR #86, opened 2024-07-23
adds library funcs for metadata and gpu

stavros-k:


library/1.0.0/metadata.py
```diff
@@ -0,0 +1,73 @@
...
+        if not path.startswith("/"):
+            utils.throw_error(
+                f"Expected [portal.path] to start with /, got [{portal['path']}]"
+            )
+
+        result.append(
+            {
+                "name": name,
+                "scheme": scheme,
+                # TODO: Default to something else?
```
stavros-k, line 66: https://github.com/truenas/middleware/pull/14057#discussion_r1687884797

library/1.0.0/metadata.py
```diff
@@ -0,0 +1,72 @@
+from . import utils
+
+
+def get_header(app_name: str):
+    return f"""# Welcome to TrueNAS SCALE
```
Qubad786, line 5: You can use `textwrap` here to fix the formatting but i think we should/can probably revisit this once at least we have the functionality in and clean it up appropriately

library/1.0.0/metadata.py
Qubad786: Wildcard default is good and handled in middleware ✅

migration_helpers/resources.py
```diff
@@ -2,19 +2,30 @@
...
+    if resources.get("limits", {}).get("cpu", ""):
+        result["limits"].update(
+            {"cpus": transform_cpu(resources.get("limits", {}).get("cpu", ""))}
+        )
+    if resources.get("limits", {}).get("memory", ""):
+        result["limits"].update(
+            {"memory": transform_memory(resources.get("limits", {}).get("memory", ""))}
+        )
+
+    for gpu in gpus.items() if gpus else []:
```
Qubad786, line 24: Curious about `gpus` format here - can you share it please?

Qubad786: So now it won't migrate nvidia ones if i am not wrong?

stavros-k: Correct, one way to solve this is to pass information about nvidia gpus in the migration script.
if there was previously `x` number of nvidia gpus, we can pick the first `x` nvidia gpus available at the time of migration. But in this case the UUIDs values etc  are needed in the migration script

Qubad786: I think just passing through one nvidia gpu regardless of the count is fine if a nvidia gpu earlier was configured

Qubad786: @stavros-k the migration can provide list of resources available on that end so the migration can go through seamlessly.
@sonicaj something like the following can work perhaps?
```
{'nvidia_gpus': ['uuid1', 'uuid2']}
```

Qubad786: It can be passed through as additional context so the migration can take an informed decision - @stavros-k would there be other cases which the config migration might require?

stavros-k: I can't think of something else currently, neither found something else missing!

stavros-k: > ```
> 
> {'nvidia_gpus': ['uuid1', 'uuid2']}
> 
> ```

The pci_slot would be needed too. Otherwise the new schema won't validate

sonicaj (approved):

PR #89, opened 2024-07-30
Add equivalent of emptyDir for docker

stavros-k:
Context:

Finding a replacement for `emptyDir` of k8s. This helps reducing permissions on the container, by presenting a writable directory where normally elevated perms would be needed.

Docker volumes (and anonymous volumes) when created are owned by the user that runs docker daemon (root).
While anonymous are great for temporary directories, which are cleaned easily after container is stopped, there is not way to change their permission neither on creation, nor by using an "init" container.

Using `named` volumes, we can share the volume with an "init' container and apply whatever permissions needed.
But this creates an issue where its no longer "temporary".

This change handles just that, `temporary` storage type is created as a named volume, and on each start an init container makes sure it is empty and applies permissions.


Qubad786 (approved):
We should be bumping library version as well but i guess we can start doing it from now as no one is really consuming this right now

stavros-k:
Yea once it starts being more consumed I'll be bumping it

PR #96, opened 2024-08-02
Add nextcloud

stavros-k:
Additionally fixes https://github.com/truenas/charts/issues/2123

Instead of having to install packages during runtime of the container, we do that now before the container even starts.
A hash of the inlined dockerfile is used to ensure that any change will cause a rebuild of the image.
Additionally, if the hashed image is already present on the host, the cached image will be re-used.


Qubad786 (approved):
Nice work 👏

PR #97, opened 2024-08-05
lock concurrency on catalog update

stavros-k:
Currently when 2 PRs are merged at the same time, update catalog will most likely fail.
Catalog update performs 2 commits, which between them the second catalog update can attempt to commit.

Solution:
Lock the catalog update concurrency to a single run at a time.
So newer catalog update workflows will be queued.

---

Also fixes a typo that leads to a CI failure here
https://github.com/truenas/apps/actions/runs/10248348970/job/28349443068

### DIFF
--- a/.github/workflows/update_catalog.yaml
+++ b/.github/workflows/update_catalog.yaml
@@ -1,9 +1,12 @@
 name: catalog_update
 
+concurrency:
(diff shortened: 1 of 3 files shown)


Qubad786 (approved):
Nice catch 👍

PR #98, opened 2024-08-06
Add github action to manually trigger update catalog workflow

sonicaj:


.github/workflows/manual_update_catalog.yaml
```diff
@@ -0,0 +1,45 @@
+name: catalog_update
```
Qubad786, line 1: Probably change this as well

Qubad786 (approved):

PR #99, opened 2024-08-06
add wg-easy

stavros-k:
We previously had a custom pre-stop script (https://github.com/truenas/charts/blob/a6a5a16af17407303798d70301e5b90c622dd7fd/library/ix-dev/charts/wg-easy/templates/_wgeasy.tpl#L70-L85)

But now wg-easy handles the cleanup automatically, so we dont have this custom script here. `https://github.com/wg-easy/wg-easy/pull/862`


ix-dev/stable/wg-easy/app.yaml
```diff
@@ -0,0 +1,39 @@
...
+  uid: 0
+  user_name: root
+screenshots:
+- https://media.sys.truenas.net/apps/wg-easy/screenshots/screenshot1.png
+sources:
+- https://github.com/WeeJeWel/wg-easy
+- https://github.com/truenas/charts/tree/master/charts/wg-easy
+title: WG Easy
+train: stable
+version: 1.0.1
```
Qubad786, line 39: Any particular reason for `1.0.1` here ?
`1.0.0` should work as well

Qubad786 (approved):

ix-dev/stable/wg-easy/app.yaml
stavros-k: Thanks for the catch!

PR #108, opened 2024-08-08
add zerotier

stavros-k:
App failure is expected (No API Key)


Qubad786 (commented):
We have failing tests, is that known?

stavros-k:
> We have failing tests, is that known?

Yes, we need api key to make it pass.

Qubad786 (approved):

PR #132, opened 2024-08-20
nextcloud: add ocrmypdf pacakge

stavros-k:
Closes https://github.com/truenas/charts/issues/2748


ix-dev/stable/nextcloud/questions.yaml
```diff
@@ -50,6 +50,28 @@ questions:
...
+                      description: ocrmypdf
+        - variable: tesseract_languages
+          label: Tesseract Language Codes
+          description: |
+            The tesseract languages to install for OCRmypdf.</br>
+            See a list of language codes here https://tesseract-ocr.github.io/tessdoc/Data-Files-in-different-versions.html.</br>
+            Typing a wrong language code will block the container from starting.</br>
+            Only takes effect if ocrmypdf is selected above.
+          schema:
+            type: list
```
stavros-k, line 63: I've tried to use a `show_if` with either `in` or `rin` operators,
eg `[["ocrmypdf", "in", "apt_packages"]]` but it seems to be expected `string` instead of a list.

Is there any other filter that I can use here?

> I want to show this section when `ocrmypdf` is selected in the `apt_packages`

stavros-k: A bit more clarification:

the `tesseract_languages` field should `show_if` 
the static value of `ocrmypdf` is selected in the `apt_packges` variable

Qubad786 (approved):

PR #146, opened 2024-08-23
Add webdav

stavros-k:


Qubad786 (approved):

claudiosv:
I don't think this is working as expected. In the original chart, the webdav httpd conf for each share is templated as `Alias /{{ .name }} "/{{ include "webdav.shares.prefix" $ }}/{{ .name }}"`. In this PR, it is `Alias /{{ share.name }} "{{ values.consts.shares_prefix }}/{{ share.name }}"`. Notice the lack of a prefix forward slash. I think this is causing non of my shares to appear (error upon navigating to /mysharename/).

stavros-k:
> I don't think this is working as expected. In the original chart, the webdav httpd conf for each share is templated as `Alias /{{ .name }} "/{{ include "webdav.shares.prefix" $ }}/{{ .name }}"`. In this PR, it is `Alias /{{ share.name }} "{{ values.consts.shares_prefix }}/{{ share.name }}"`. Notice the lack of a prefix forward slash. I think this is causing non of my shares to appear (error upon navigating to /mysharename/).

Thanks, fixed in #389

PR #281, opened 2024-08-27
add wg-easy migration

stavros-k:


ix-dev/stable/wg-easy/questions.yaml
```diff
@@ -128,7 +128,7 @@ questions:
             Bind to the host network. It's recommended to keep this disabled.
           schema:
             type: boolean
-            default: false
+            default: true
```
Qubad786, line 131: In the description we say it is recommended to keep this disabled but we are defaulting to it being enabled..that looks a bit off?

stavros-k: Fixed it! Thanks for the catch

Qubad786 (approved):

PR #334, opened 2024-08-31
qBittorrent: Use correct downloads volume configurations

joren485:
This fixes a typo in the qBittorrent docker compose file, where the qBittorrent config volume configuration is used for the downloads volume.


stavros-k (approved):

stavros-k:
Thanks!

PR #335, opened 2024-08-31
FreshRSS: Use correct extensions volume config

joren485:
Similar to #334, this fixes the reuse of a volume path variable.

After #334, I quickly looked for similar problems and found this one. Please note, I did not test this fix, but from looking at the template, I believe `values.storage.extensions` is correct (as it is not used anywhere else).


stavros-k (approved):
thanks

PR #363, opened 2024-09-04
Add ix-app + migration

stavros-k:
Diff for library is funny here..

Actual changes can be seen here https://github.com/truenas/apps/pull/363/files#diff-a824e355d3c4450e68f6b30070a53b36aa90bd3a244179ba5663a46af298b4ca (resources.py file)


ix-dev/stable/ix-app/migrations/migration_helpers/kubernetes_secrets.py
```diff
@@ -0,0 +1,15 @@
+def get_value_from_secret(secrets={}, secret_name="", key=""):
```
yocalebo, line 1: kwargs should be initialized to `None` (it's the paradigm we follow in middleware repo).

Inside your function it'd look like this:
```
def get_value_from_secret(secrets=None, secret_name=None, key=None):
    secrets = secrets if secrets else dict()
    secret_name = secret_name if secret_name or ""
    key = key if key else ""

Qubad786 (approved):
LGTM, finally!

PR #395, opened 2024-09-12
sabnzbd: update default suggested web_port to 30055 to match original chart

salekseev:
THis is to match https://github.com/truenas/charts/blob/master/community/sabnzbd/1.3.1/questions.yaml#L97C1-L98C1


stavros-k:
Thanks!

PR #405, opened 2024-09-12
nextcloud: add gpu questions

stavros-k:
Fixes https://github.com/truenas/charts/issues/2078

I couldn't find any additional dependencies needed gpu is used.
We can come back to this if we see any issues when consuming one.

(Library should pick up the gpu's from questions and map them automatically)


frederickjansen:
Thanks, I'll try this out once 24.10 is released.

PR #406, opened 2024-09-12
Add rust desk

stavros-k:


ix-dev/community/rust-desk/app.yaml
```diff
@@ -0,0 +1,31 @@
+app_version: 1.1.11.1
```
Qubad786, line 1: Should it be `1.1.11-1` instead ?

Qubad786 (approved):

PR #416, opened 2024-09-14
Fix paperless-ngx wrong consumption dir env variable

aaronjdale:
Basic paperless-ngx setup leaves env variable PAPERLES_CONSUMPTION_PATH empty. Suspect that it's just a typo here.

Screenshot 2024-09-13 at 10 26 42 PM


stavros-k (approved):
Thanks!

PR #480, opened 2024-09-22
set erroneous true to false

itsbhanusharma:
I'm not sure if this is the correct way to do this, but there seems to be a weird issue where subquestions are only shown when host network check is selected on frigate setup. And due to it binding to host network, these port setting are not respected  e.g. my frigate app is currently available on port 8971 which is frigate default and not 30058 which is what is set in the config. I strongly suspect this is due to the erroneous true set for subquestions. If this is not the right way to do this, please do the needful so frigate could be configured properly.


stavros-k (approved):
Nice catch. Thanks

PR #624, opened 2024-10-07
qbittorrent4: keep old version of qbittorrent

dextercai:
Keep the old version to meet some special needs

### DIFF
--- a/ix-dev/community/qbittorrent4/README.md
+++ b/ix-dev/community/qbittorrent4/README.md
@@ -1,0 +1,3 @@
+# qBittorrent4
+
+The [qBittorrent](https://www.qbittorrent.org/) project aims to provide an open-source software alternative to µTorrent.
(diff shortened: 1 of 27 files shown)


stavros-k:
Can you please explain the use case?
Also if this is to happen, it won't be a seperate app, but either just keeping the last v4 version in the dropdown selector during installation or an image selector. Depending on whether upstream are going to release images for the v4.

dextercai:
Current qbitorrent app do not have an image selector. I got your idea. I will create another PR later.

stavros-k:
Does upstream going to push more releases for v4?
if not we will just restore the last v4 app version and keep it available for installation.

Image selector will only be accepted if there is going to be v4 updates on the image.

dextercai:
As the official website of qbit, v4.6.7 will be the last minute version of 4. 

Ref: https://www.qbittorrent.org/news

dextercai:
I know this question may not be appropriate in this PR, but please tell me how to debug in the truenas app list. I don't seem to find relevant documentation. @stavros-k

stavros-k:
There is no easy way to test or debug in a truenas system currently.

PR #629, opened 2024-10-07
v2 lib

stavros-k:
V2 lib aims to 

- Have a standard API to use within templates.
- Validate and fail as early as possible in development with clear messages
- Automate some fields that are coming from known structures, like resources, network, storage etc
- Have tests so we can ensure we don't break things in the future
- Minimize the need to use jinja templating as much as possible, in order to avoid its quirks.
- Provide a cleaner way to define apps
- be extendable

So far there are few things that have been added, but few bigs are missing.

- [x] Storage
- [x] Dependencies (predefined) eg postgres/mariadb/redis/perms
- [ ] Networks (Will be done in a later PR)

---

Currently the imports are duplicated, once with a relative import and one without inside a try/except block.
That is because in jinja it wont work without the relative import, but pytest wont work with relative imports.
If there is a way to fix that, please let me know!


stavros-k:
Minus the storage (for now), this is how a compose file will look like for a single container

```
{% set renderer = ix_lib.base.render.Render(values) %}

{% set c1 = renderer.add_container(values.consts.app_container_name, "image") %}

{# TODO: additional storages #}
{% do c1.ports.add_port(values.network.web_port, values.network.web_port) %}

{% do c1.environment.add_user_envs(values.app.additional_envs) %}
{% do c1.environment.add_env("HOST", "0.0.0.0") %}
{% do c1.environment.add_env("PORT", values.network.web_port) %}

{% do c1.healthcheck.set_test("curl", {"port": values.network.web_port, "path": "/health"}) %}

{{ renderer.render() | tojson }}
```

stavros-k:
I think this is fine for initial review, and potentially merge and start using it on _new_ apps for now.
To make sure everything works smoothly, before we start using it on existing apps.

Qubad786 (approved):

PR #663, opened 2024-10-15
adguard: healthcheck endpoint is not working

stavros-k:
During the initial setup adguard redirects everything to `/install`
Which made me believe that the `/health-check` endpoint works (https://github.com/AdguardTeam/AdGuardHome/blob/41cce6259709d450b3c6c310cedd9610f4f82a69/internal/next/websvc/path.go#L8)

But it seems that it will exist in a `next` version?!

---

Switch back to using `/` for healthcheck


stavros-k:
Added the change in #662 to avoid conflicts

Issue #705, opened 2024-10-26
RSync daemon doesn't seem to have SYS_CHROOT permissions


jsnyder:
I installed App Version: 1.0.0 on ElectricEel-24.10-RC.2 and mostly left defaults including a UID of 0 and pointed it at a dataset on my main pool.  After setup when I attempt to connect to it I see this at the client rsync:

```
@ERROR: chroot failed
rsync error: error starting client-server protocol (code 5) at main.c(1863) [sender=3.2.7]
```

I also can't chroot within the docker container:
```
# chroot /tmp
chroot: cannot change root directory to '/tmp': Operation not permitted
```

In docker inspect it doesn't look like the capabilities have been added:
```
            "CapAdd": null,
            "CapDrop": [
                "ALL"
            ],
```

The portainer install I have looks like it got capabilities set:
```
            "CapAdd": [
                "CHOWN",
                "DAC_OVERRIDE",
                "FOWNER",
                "KILL",
                "SETGID",
                "SETUID",
                "SETPCAP",
                "SETFCAP"
            ],
            "CapDrop": [
                "ALL"
            ],
```

It looks like it may just be lacking a cap_add line below here:
https://github.com/truenas/apps/blob/a9d775c33a58eca17a36023bf78ba6afd61342a1/ix-dev/community/rsyncd/templates/docker-compose.yaml#L49

Similar to:
https://github.com/truenas/apps/blob/a9d775c33a58eca17a36023bf78ba6afd61342a1/ix-dev/community/portainer/templates/docker-compose.yaml#L66

PR #706, opened 2024-10-26
rsyncd: fix to add capabilities in compose template

jsnyder:
See a description in Issue #705.

I'm not entirely sure how to switch to another repo to get apps from yet, but I did try making the same change locally in `/mnt/.ix-apps/app_configs/rsyncd/versions/1.0.14/templates/docker-compose.yaml` then made a config change to the app so it redeployed and it appears to have the correct capabilities now and I can use the rsync daemon on the version of Electric Eel mentioned in the issue.

I did see that the repo says no PRs right now, but figured this should be pretty straightfoward and I may as well submit a single line change.

### DIFF
--- a/ix-dev/community/rsyncd/app.yaml
+++ b/ix-dev/community/rsyncd/app.yaml
@@ -41,4 +41,4 @@ sources:
 - https://hub.docker.com/r/ixsystems/rsyncd
 title: Rsync Daemon
 train: community
-version: 1.0.14
+version: 1.0.15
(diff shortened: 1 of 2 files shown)


stavros-k (approved):
This is a dejavu. I do remember doing the same thing couple days ago but cant remember the app :D

Please also bump the app.yaml `version` 

Thanks

jsnyder:
Thanks! Been a busy few days since.

PR #758, opened 2024-10-31
update apps to 1.1.5

stavros-k:


stavros-k:
See https://github.com/truenas/apps/pull/763

PR #766, opened 2024-10-31
bump lib to v2.0.6

stavros-k:


Qubad786 (approved):
PR title and actual library version don't match

PR #767, opened 2024-10-31
jellyfin: match validation rule to charts repo

stavros-k:
Fixes 
https://forums.truenas.com/t/jellyfin-migration-failed-during-upgrade-to-electric-eel-what-to-do-next/22384?u=djp-ix


stavros-k:
See https://github.com/truenas/apps/pull/770

Issue #837, opened 2024-11-07
Paperless-ngx - Add Tika and Gotenberg


tina-junold:
Hi,

I would like to see the Paperless-ngx add having additional two containers running, Tika and Gotenberg ([docs](https://docs.paperless-ngx.com/configuration/#optional-services)). 

I'm willing to help with a pull request if necessary. The docker-compose.yml of the paperless-ngx project looks like:

```yaml
    #[...]
    environment:
      PAPERLESS_TIKA_ENABLED: 1
      PAPERLESS_TIKA_GOTENBERG_ENDPOINT: http://gotenberg:3000
      PAPERLESS_TIKA_ENDPOINT: http://tika:9998

  gotenberg:
    image: gotenberg/gotenberg:latest
    restart: unless-stopped

    # The gotenberg chromium route is used to convert .eml files. We do not
    # want to allow external content like tracking pixels or even javascript.
    command:
      - "gotenberg"
      - "--chromium-disable-javascript=true"
      - "--chromium-allow-list=file:///tmp/.*"

  tika:
    image: docker.io/apache/tika:latest
    restart: unless-stopped
```

tina-junold:
@stavros-k This is my first commit in such a big project. Please forgive me if i have not followed some processes correctly.

stavros-k:
Hello, thanks for wanting to help!

We have this warning on the top of the readme (https://github.com/truenas/apps?tab=readme-ov-file#%EF%B8%8F-contributions) because everything is still going under heavy changes, as you will notice now, all your changes are outdated as I've just updater paperless to v2 lib.

You can probably work with the v2 lib now. Tho its still undocumented but there are tests and few other apps if you want to see examples.

Have you figured yet how to test/render locally?
I'm using the following ( you might need to install some python packages)

```sh
./.github/scripts/ci.py --app paperless-ngx --train community --test-file basic-values.yaml --wait=true
./.github/scripts/ci.py --app paperless-ngx --train community --test-file basic-values.yaml --render-only=true
```

freundTech:
Are you still working on this @tina-junold? If not I'd like to pick this up, as I could also use tika and gotenberg support.

tina-junold:
@freundTech yes i am:

https://github.com/truenas/apps/pull/1096

PR #850, opened 2024-11-08
#837 | Added Tika and Gotenberg to Paperless-ngx

tina-junold:
# Not quite sure with the syntax for tika user_name since the container has no username for tika. Also not sure if quotes for user are correct.


stavros-k:
In case you missed my comment here (https://github.com/truenas/apps/issues/837#issuecomment-2464601187).
I'll close this one, as a lot had changed. Feel free to open a new one if you plan to work on this!

There are now some more steps outlined here if you want to read.
https://github.com/truenas/apps/blob/master/CONTRIBUTIONS.md

Thanks!

PR #913, opened 2024-11-13
qbittorrent: handle better the path

### LINKED ISSUE #917: qbittorrent 1.1.1 can't find downloaded files
After updating qbittorrent to 5.0.1_1.1.1, it doesn't start properly due to apparently being unable to access the downloads directory. The directory in question is located on an existing dataset and has a correct ACL entry for UID 568.

When starting 1.1.1, the container remains as "starting" and I'm unable to open a shell to check anything further. The web UI works fine, and reports "Missing Files" for all transfers.

I have not changed any part of the configuration. The save location setting in the web UI is still `/downloads`.

stavros-k:
Handle better the path. So it will be consistent across versions.
Also replace deprecated variables.

Ended up adding a config container, that updates the config file.

Fixes #917


chencaidy:
Sorry, I tried install a new instance, , that document also outdated and not working for new user.

stavros-k:
> Sorry, I tried install a new instance, env not change path, that document also outdated and not working for new user.
> 
> I think it will be a knotty problem.
> 
> If user who upgrade from v4, the path and all files metadata will keep in /downloads. but new user default path is /config/Downloads.

yea. the variable is for the host side.
I'm looking into alternative (cli flag)

chencaidy:
> > Sorry, I tried install a new instance, env not change path, that document also outdated and not working for new user.
> > I think it will be a knotty problem.
> > If user who upgrade from v4, the path and all files metadata will keep in /downloads. but new user default path is /config/Downloads.
> 
> yea. the variable is for the host side. I'm looking into alternative (cli flag)

The offical docker image seems like solved this problem.
https://hub.docker.com/r/qbittorrentofficial/qbittorrent-nox
图片

Maybe need some compare between qbittorrentofficial and onedr0p's image.

PR #981, opened 2024-11-20
rsyncd : allow uid/gid > 65535

mpnico:
UID/GID can be > 65535, for example in some AD domain situations. This PR allows UID/GID to be up to max int32 (4294967295) in rsyncd module UI


stavros-k:
Please see https://github.com/truenas/apps/blob/7b87ec57d3c69ce733a09c6735d6dbc28b388df1/CONTRIBUTIONS.md?plain=1#L26-L27

Thanks

mpnico:
Sorry, and thank you for pointing this out. I'll make another PR the right way.

PR #990, opened 2024-11-21
rsyncd : allow uid/gid > 65535

mpnico:
UID/GID can be > 65535, for example in some AD domain situations. This PR allows UID/GID to be up to max int32 (4294967295) in rsyncd module UI


stavros-k:
@Qubad786  Does the `max` support big numbers like this?

Qubad786 (approved):

Qubad786:
@stavros-k yes it does

PR #998, opened 2024-11-22
Docs workflolw

stavros-k:
Closes https://ixsystems.atlassian.net/browse/NAS-132695

One thing to note is that there is a chance that the comment will not run when PRs are opened from forks.
We will have to re-check

Result:

image


kmoore134 (approved):

.github/scripts/message.py
```diff
@@ -0,0 +1,97 @@
...
+                    message += f"- `{file}`\n"
+                message += "\n"
+            if len(changes[train]["apps"][app]["modified"]) > 0:
+                message += "Modified files:\n"
+                for file in sorted(changes[train]["apps"][app]["modified"]):
+                    message += f"- `{file}`\n"
+                message += "\n"
+            message += "\n---\n\n"
+
+    if message != "":
```
Qubad786, line 79: could do
```
if message:
```

.github/scripts/message.py
```diff
@@ -0,0 +1,97 @@
...
+    except json.JSONDecodeError:
+        print(f"Failed to decode JSON from {file}", file=sys.stderr)
+        exit(1)
+
+
+trains_to_check = ["test", "stable", "enterprise"]
+account_to_notify = ["@truenas/docs-team"]
+
+
+def process(changed_files=[], added_files=[]):
```
Qubad786, line 20: We should never use mutable args as default args

PR #1074, opened 2024-11-29
Add ParsonLabs Music to Application Catalog

WillKirkmanM:
home-w
home-te
explore-page
album_showcase_wca
library-showcase
music-album-md
pfp-showcase
genres-showcase
music-video-home
plm-complete
genre-selection
music-video
home-plm
updated-lyrics
updated-artist
ArtistAlbumSearch
profile
search
lyrics-showcase
music-artist-showcase
music-in-your-library
music-playlist


stavros-k (changes requested):
Hello, this does not seem to work at all.
I cannot even render the template.

Image is recommended to use a versioned tag and not latest.
Port is static so users cannot edit it. Also it would be nice if internal container port is also changed via env variable to match the external one.
So users can also use host network without chances of port conflicts.

certificate is not used anywhere

container name is different in the labels section than the one that is used to create the container.

values are statically set in the compose file instead of using the values file.

For readme I'd suggest having a small description like the rest of the apps,
and have a link to your github for the rest.  (as the rest of the apps)

app.yaml file should be the same format as the rest, as it is modified from automated tools and will create big diffs on subsequent automated PRs

questions.yaml seems to have a user/group `group` defined at the top but is not used anywhere.
Is the container running as root or rootless?
Based on the compose it seems to run as rootless, but uses cannot edit that. as you reference it from the consts.

WillKirkmanM:
Thanks for the review, will fix.

stavros-k:
Hello, is there any progress here?
Are you still going to work on this?

Thanks

WillKirkmanM:
Yes I will fix it today

stavros-k:
Due to being stale for too long and lots of things in the lib API changed. I'll close this for now.
You can pick some of the newer apps as example like `umami` that use all the latest apis
(Mainly around port usage) and adapt.
Once you are ready, feel free to re-open

PR #1091, opened 2024-12-02
update contributions doc file

stavros-k:


CONTRIBUTIONS.md
```diff
@@ -59,6 +59,52 @@ The above command will also do the following:
...
+```
+
+## Testing on TrueNAS System
+
+There is no easy way currently to test on TrueNAS.
+As long as it works on your local machine, it should work on TrueNAS.
+There are some exceptions for things like devices etc, that might be different.
+In such case, let the reviewer know.
+
+`questions.yaml` is run though some validation during CI, but it still needs some manual check from the reviewer.
```
Qubad786, line 75: typo here
`though` --> `through`

Qubad786 (commented):
There's just a small typo

Qubad786 (approved):

PR #1164, opened 2024-12-12
fix(nextcloud): Rename network.nginx.use_different_access_port to network.nginx.use_different_port

JulesTriomphe:
This PR renames `network.nginx.use_different_access_port` to `network.nginx.use_different_port` to fix the use of the different port as it is not picked up at the moment.

It also adds the domain with the different port to `NEXTCLOUD_TRUSTED_DOMAINS`.


stavros-k (approved):
Nice catch. Thanks!

PR #1203, opened 2024-12-20
library: explicitly disallow /mnt mounting

stavros-k:


library/2.1.5/validations.py
```diff
@@ -135,6 +135,13 @@ def valid_fs_path_or_raise(path: str):
     return path
 
 
+def allowed_fs_host_path_or_raise(path: str):
+    disallowed_paths = ["/mnt"]
```
anodos325, line 139: Might make sense to disallow `/` as well.

library/2.1.5/validations.py
```diff
@@ -135,6 +135,13 @@ def valid_fs_path_or_raise(path: str):
...
+    if path in disallowed_paths:
```
yocalebo, line 140: Gotta be a bit more strict here. We should also prevent straight up `/`.

`/mnt/ in ['/mnt'] == False`

library/2.1.5/validations.py
anodos325: Also `/var/db` and `/data`

anodos325: Oh, and `/audit` :)

stavros-k: `/` is already blocked in both container and host paths. (check above func)
Will add the rest as well

library/2.1.5/validations.py
stavros-k: it is passed already with `rstrip("/")`. I'll add rstrip here as well as failsafe

library/2.1.5/validations.py
```diff
@@ -135,6 +135,13 @@ def valid_fs_path_or_raise(path: str):
     return path
 
 
+def allowed_fs_host_path_or_raise(path: str):
+    disallowed_paths = ["/mnt", "/audit", "/var/db", "/data"]
+    if path.rstrip("/") in disallowed_paths:
```
yocalebo, line 140: We have too many disallowed paths. It'll be easier going about it from the other direction. i.e.
```
if not path.startswith('/mnt') or path.rstrip('/') == '/mnt':
    raise error

yocalebo (changes requested):

yocalebo (approved):

PR #1265, opened 2025-01-02
remove psutil dependency

stavros-k:
Not bumping versions yet, as it will cause conflicts with other open PRs.
Will bump later this week all apps with a new library version


yocalebo (approved):
It's a shame that you have to copy and paste that function so many times.

PR #1284, opened 2025-01-03
Update z2m for SLZB-06M

domoritz:
The SLZB-06M has a different adapter not listed.


TheJulianJES:
The `ezsp` adapter type is [deprecated](https://www.zigbee2mqtt.io/guide/adapters/emberznet.html) and you should use the newer `ember` type. You might need to install a newer firmware on your Zigbee coordinator though, as EmberZNet 7.4.x or higher is required by `ember`.

domoritz:
Ah, thanks for the note. No need for this then.

Issue #1296, opened 2025-01-05
Can't reach docker containers via IPv6 with default settings


stuchlyf:
Since the `add_port` method in the `Ports` class uses `"0.0.0.0"` as it's default, you can't access the port with IPv6 if you don't explicitly add the IPv6 mapping.

And since some apps don't support customization on ports, IPv6 isn't supported by those at all.
An example for this is PiHole which by default binds the WebUI port and the DNS port to `"0.0.0.0"` but not to `"[::]"` so I can't access it and there is also no way to add this port mapping to the existing ones.
![image](https://github.com/user-attachments/assets/0bf7b2a0-c89c-45cc-bc1e-71212153da06)

My suggestion for this problem would be to add another case in the `add_port` method which adds the port to anyhost in IPv4 and IPv6 if the host IP wasn't specified.

stuchlyf:
I've created a PR showing my suggestion: #1297

PR #1297, opened 2025-01-05
Add mappings to IPv4 and IPv6 when no explicit `host_ip` is given

stuchlyf:
see #1296


stavros-k:
Hello, I'm not opposed on doing something like this.
Tho I'll make some changes, to also allow setting multiple host_ips per port, later on.
Will take a closer look within the week

PR #1351, opened 2025-01-12
Added Imaginary as an optional container alongside NextCloud

ChapterSevenSeeds:
Gives the user the option to install [Imaginary](https://docs.nextcloud.com/server/latest/admin_manual/installation/server_tuning.html#previews) alongside their NextCloud installation.

A couple of things:
- The installation uses the default port of 9000 which stays completely local to the Docker network. 
- The NextCloud automatic configuration doesn't yet support configuring the `preview_imaginary_url` setting in `config.php`. As such, I left a note in the description informing the user of what they need to change if they wish to use Imaginary. 
- I'm not entirely sure of the convention regarding the proper Docker user and group. The Imaginary image I am using, found [here](https://github.com/nextcloud/all-in-one/blob/main/Containers/imaginary/Dockerfile), sets the user to ID 65534. Since I don't think there exists any need to change this user ID or the relevant permissions, I did nothing with the user/group settings for the container in the corresponding `docker-compose.yaml`. If this is incorrect, please let me know and I will happily make any necessary changes.


stavros-k:
> * The NextCloud automatic configuration doesn't yet support configuring the `preview_imaginary_url` setting in `config.php`. As such, I left a note in the description informing the user of what they need to change if they wish to use Imaginary.

You can actually do something like this
https://github.com/stavros-k/containers/blob/2751f36b10c4ebd256ef57046be37bf0e0122732/apps/nextcloud-fpm/configure-scripts/occ-imaginary.sh#L4

> * I'm not entirely sure of the convention regarding the proper Docker user and group. The Imaginary image I am using, found [here](https://github.com/nextcloud/all-in-one/blob/main/Containers/imaginary/Dockerfile), sets the user to ID 65534. Since I don't think there exists any need to change this user ID or the relevant permissions, I did nothing with the user/group settings for the container in the corresponding `docker-compose.yaml`. If this is incorrect, please let me know and I will happily make any necessary changes.

Considering imaginary does not store/access any day, we can just use this ID directly. (I prefer explicitly setting it)

---

That being said, I'm thinking more and more to do a big facelift for nextcloud app.
Using a pattern like the script I linked above.
Tho, if I do that, that means all users will have a complete image with all the dependencies.
While larger to fetch, will shave time from apt installing packages at runtime.
Also means that plugins that we add support for, will be only managed from the UI, eg
if you disable imaginary from the truenas UI, it will also remove it from nextcloud on each restart.

And the last bit, is that will be most likely a breaking change, meaning a backup/restore of the db and a re-install will be needed.

I'm not 100% sure I'll go forward with this change tho. Still contemplating

ChapterSevenSeeds:
Oh that's interesting. Would that script be run during the startup of the NextCloud container? If so, are there other scripts that run similarly in this repository that I can use for reference?

Though I guess this won't matter if you decide to proceed with the facelift you described. Ultimately, with something as customizable as NextCloud, I've had a few thoughts here and there about ditching the app from the TrueNAS catalog and instead doing my own homegrown Docker Compose, saving me from being bottlenecked through the customization exposed through the TrueNAS UI. However, with how far this app in the catalog has come over the past few years, I've really liked having a lot of the configuration and setup hell hidden from my eyes.

ix-dev/stable/nextcloud/app.yaml
```diff
@@ -56,6 +58,11 @@ run_as_context:
   group_name: root
   uid: 0
   user_name: root
+- description: Imaginary runs as user 65534.
+  gid: 65534
+  group_name:
+  uid: 65534
+  user_name:
```
stavros-k, line 65:
```suggestion
- description: Imaginary runs as non-root user.
  gid: 65534
  group_name: imaginary
  uid: 65534
  user_name: imaginary
```

Those can't be empty, lets set them to something relevant atleast

ix-dev/stable/nextcloud/questions.yaml
```diff
@@ -92,6 +92,22 @@ questions:
...
+          schema:
+            type: dict
+            attrs:
+              - variable: enabled
+                label: Enabled
+                description: Enable Imaginary
+                schema:
+                  type: boolean
+                  default: false
+                  required: true
```
stavros-k, line 110:
```suggestion
```

Otherwise it will not allow to continue unless you enable it.

ix-dev/stable/nextcloud/templates/docker-compose.yaml
```diff
@@ -174,6 +174,13 @@
   {% do cron_container.environment.add_user_envs(values.nextcloud.additional_envs) %}
 {% endif %}
 
+{% if values.nextcloud.imaginary.enabled %}
+  {% set imaginary_container = tpl.add_container(values.consts.imaginary_container_name, "imaginary_image") %}
+  {% do imaginary_container.add_caps(["SYS_NICE"]) %}
```
stavros-k, line 179:
```suggestion
  {% do imaginary_container.set_user(65534, 65534) %}
  {% do imaginary_container.add_caps(["SYS_NICE"]) %}
```

Now I'm thinking about this, given this is usually the nobody user, it might also work with 568.
Can you give it a shot?

ix-dev/stable/nextcloud/app.yaml
ChapterSevenSeeds: Setting the user to 568 worked just fine.

ChapterSevenSeeds:
@stavros-k Thanks for the feedback and pointers. Please see my most recent changes as they address all your comments.

stavros-k (approved):

stavros-k:
had to remove the "unsetting of imaginary url" because this would probably break any user that had it running externally.

But once/if the nextcloud facelift happens, it will be from the get go with add/remove.
Thanks!

stavros-k (approved):

PR #1354, opened 2025-01-13
Set Lyrion Music Server to v9.0.2

### LINKED ISSUE #1348: lyrion-music-server installs development version instead of latest release version
lyrion-music-server installs v9.1.0 which according to docker hub is
"the development version⁠, with new features, and potentially less stability, currently v9.1.0"

I think for mainstream usage in truenas the "latest release version, currently v9.0.1" would be more appropriate.

hartzell:
Unfortunately, LMS pushes a version tag for the develop branch, even though it's not released.

This will make automated updating problematic.

Closes #1348

### DIFF
--- a/ix-dev/community/lyrion-music-server/app.yaml
+++ b/ix-dev/community/lyrion-music-server/app.yaml
@@ -1,4 +1,4 @@
-app_version: 9.1.0
+app_version: 9.0.2
 capabilities:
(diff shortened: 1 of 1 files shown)


stavros-k:
I'd prefer we close this and file an upstream issue to hopefully get something in the tag that we can differentiate.
What do you think?

PR #1355, opened 2025-01-13
portainer: add http port

### LINKED ISSUE #1307: Portainer: Add option in TrueNas WebUI to use http (instead of forcing https)
Hello,
I've setup a reverse proxy so I can do all my SSL termination on the reverse proxy host and liberate some resources on the endpoints.
It seems that the portainer app is launched with the --http-disabled command. 
I'd like for the user to be able to decide whether http is disabled or not.
Thank you

stavros-k:
Closes #1307

### DIFF
--- a/ix-dev/community/portainer/app.yaml
+++ b/ix-dev/community/portainer/app.yaml
@@ -49,4 +49,4 @@ sources:
 - https://github.com/portainer/portainer
 title: Portainer
 train: community
-version: 1.3.9
+version: 1.3.10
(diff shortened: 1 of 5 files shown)


Qubad786 (approved):

185245725:
this is not working at all as of right now, the http port never gets published or exposed, with or without host ips. portainer logs are showing that the http server is enabled, but the truenas app shows no ports.

TrueNAS Community Edition 25.04.1
Portainer Community Edition
App Version v2.31.0
Version v1.2.3

PR #1375, opened 2025-01-15
Small tweaks to RomM 

gantoine:
Hey ya'all, I'm one of the main devs of [RomM](https://github.com/rommapp/romm/), which was recently added to TrueNas in #1356. First off, thanks @stavros-k for adding the app, really appreciate it!

I've made some tweaks below to better align this template with a typical setup. ~~namely removing the need for an external Redis container, since the RomM image ships with [Valkey](https://valkey.io/) built-in. Also added secret keys for IGDB, Mobygames and SGDB, none of which are required but most users will set at least one.~~

I'm also curious why Postgres was used over MariaDB, as support was just added in `3.7.0` and while stable, is technically considered "experimental".

⚠️ As I don't use TrueNAS myself, I'm not familiar with the syntax and cannot test these changes. Please take them as suggestions and modify as necessary to get them working ⚠️

### DIFF
--- a/ix-dev/community/romm/questions.yaml
+++ b/ix-dev/community/romm/questions.yaml
@@ -1,15 +1,15 @@
 groups:
   - name: Romm Configuration
-    description: Deploy RomM by selecting from the following options, where in doubt use the defaults.
+    description: Deploy RomM by selecting from the following options, but when in doubt use the defaults.
   - name: User and Group Configuration
     description: Configure User and Group for Romm
   - name: Network Configuration
     description: Configure Network for Romm
   - name: Storage Configuration
     description: |
-      RomM uses two persistent volumes, one for resources (cover art, screenshots), and another for internal Redis data.
-      You will want to mount two other folders/volumes for the ROM library and the config.yml file.
-      You may want to mount an additional folder/volume for saves, states, and other adjacent files.
+      RomM uses two internal persistent volumes: one (resources) for cover art and screenshots, and another (valkey_data) for Valkey.
+      You'll want to mount two others (library and config) for the ROM library and the config.yml file.
+      You may want to mount an additional one (assets) for saves, states, and other adjacent files.
   - name: Labels Configuration
     description: Configure Labels for Romm
   - name: Resources Configuration
--- a/ix-dev/community/romm/questions.yaml
+++ b/ix-dev/community/romm/questions.yaml
@@ -7,9 +7,8 @@ groups:
     description: Configure Network for Romm
   - name: Storage Configuration
     description: |
-      RomM uses two internal persistent volumes: one (resources) for cover art and screenshots, and another (valkey_data) for Valkey.
+      RomM uses two internal persistent volumes: one (resources) for cover art and screenshots, and another (assets) for saves, states, and other adjacent files.
       You'll want to mount two others (library and config) for the ROM library and the config.yml file.
-      You may want to mount an additional one (assets) for saves, states, and other adjacent files.
   - name: Labels Configuration
     description: Configure Labels for Romm
   - name: Resources Configuration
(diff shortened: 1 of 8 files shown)


stavros-k:
> Hey ya'all, I'm one of the main devs of [RomM](https://github.com/rommapp/romm/), which was recently added to TrueNas in #1356. First off, thanks @stavros-k for adding the app, really appreciate it!
> 
> I've made some tweaks below to better align this template with a typical setup, namely removing the need for an external Redis container, since the RomM image ships with [Valkey](https://valkey.io/) built-in. Also added secret keys for IGDB, Mobygames and SGDB, none of which are required but most users will set at least one.
> 
> I'm also curious why Postgres was used over MariaDB, as support was just added in `3.7.0` and while stable, is technically considered "experimental".
> 
> ⚠️ As I don't use TrueNAS myself, I'm not familiar with the syntax and cannot test these changes. Please take them as suggestions and modify as necessary to get them working ⚠️

stavros-k:
Hello,

> Hey ya'all, I'm one of the main devs of [RomM](https://github.com/rommapp/romm/), which was recently added to TrueNas in #1356. First off, thanks @stavros-k for adding the app, really appreciate it!

🚀🚀

> I've made some tweaks below to better align this template with a typical setup, namely removing the need for an external Redis container, since the RomM image ships with [Valkey](https://valkey.io/) built-in. Also added secret keys for IGDB, Mobygames and SGDB, none of which are required but most users will set at least one.

I'm fine with adding fields for those secrets for easier setup, but why remove the redis container?
Containers meant to do as little as possible, and not ship all-in-one images. Afaik Romm does not start valkey if an external is connected. Also removing fields (such as redis_password), will break the schema. It requires migration and currently this is not supported yet.

If its about redis vs valkey, don't worry I plan change the underlying image at some point.
 
> I'm also curious why Postgres was used over MariaDB, as support was just added in `3.7.0` and while stable, is technically considered "experimental".

I prefer postgres when is available, because I find it much better than other db engines. Also its much more "developed" as one of our deps. ie having upgrades dealt automatically for newer postgres versions etc.

ix-dev/community/romm/questions.yaml
```diff
@@ -46,27 +51,50 @@ questions:
...
             default: ""
-            required: true
+            required: false
+            private: true
+        - variable: igdb_client_secret
+          label: IGDB Client Secret
+          schema:
+            type: string
+            default: ""
+            required: false
```
stavros-k, line 81: required false can be removed (thats the default)

gantoine:
> Containers meant to do as little as possible, and not ship all-in-one images.

I fully agree! Though we've found that the fewer external services users need to setup, the more likely they were to succeed in running the app. I think that might be due to our user base skewing a bit more towards gamers than self-hosters, at least compared to other apps.

> Afaik Romm does not start valkey if an external is connected. 

That's correct, and I personally run Valkey outside RomM for that reason. If you feel running Redis/Valkey as a separate service is better for TrueNAS users I'm all for it. 👍🏼 

> Also its much more "developed" as one of our deps. ie having upgrades dealt automatically for newer postgres versions etc.

Sounds great, if it makes setup and maintenance easier for ya'll then we'll make sure to keep it working it in future versions 👍🏼 

I'll revert some of these changes in a bit and let you take it from there.

stavros-k (approved):

stavros-k:
> I fully agree! Though we've found that the fewer external services users need to setup, the more likely they were to succeed in running the app. I think that might be due to our user base skewing a bit more towards gamers than self-hosters, at least compared to other apps.

Sure and I'm happy you offer opt-out of embedded services!
Luckily users here dont have to set those up!



> I did notice `base_path` was missing in the `consts`, so I set it to `/romm`. Internally all the folders should be mounted to `/romm/resources`, for example. We could also omit `base_path` and not set the env var, leaving it to the default which is _also_ `/romm`.

Nice catch. I tend to explicitly set some of those, for 2 reasons.
A change upstream of defaults won't "break" it, and also by setting it, will prohibit users from changing themselves, breaking all mounts as well.

stavros-k:
Thanks!

PR #1388, opened 2025-01-17
DRAFT: App for Octoprint

phillipkremer:
First draft of an app for Octoprint.

Validation with different value files via ci.py works fine. 
I just cannot validate that the questions file works as intended.
For example, does questions generate items for not shown entrys ?

I'm a little confused about running the app as different user.
Octoprint dosen't support that, so I didin't implement it.
The RunAs User/Group dialog in the questions stayed as is, couse other apps are configured the same, when they don't implement it.

Any feedback/validation would be appreciated.


ix-dev/community/octoprint/templates/docker-compose.yaml
stavros-k, line 1: Please use ANY other app than the ix-app.

ix-dev/community/octoprint/templates/docker-compose.yaml
```diff
@@ -0,0 +1,71 @@
...
+{% endfor %}
+
+{# Configs Configuration #}
+{% do c1.configs.add("mknod_devices.sh", script(values), "/mknod_devices.sh", "0755") %}
+
+{# Network Configuration #}
+{% do c1.ports.add_port(values.network.web_port, 80) %}
+
+{# Healthcheck #}
+{% do c1.healthcheck.use_built_in() %}
```
stavros-k, line 55: I dont see any built in here https://github.com/OctoPrint/octoprint-docker/blob/master/Dockerfile

ix-dev/community/octoprint/templates/docker-compose.yaml
```diff
@@ -0,0 +1,71 @@
...
+{% if values.octoprint.auto_tty_enable %}
+  {% do c1.add_device_cgroup_rule("c 188:* rwm") %}     
+{% endif %}
+{% if values.octoprint.auto_video_enable %}
+  {% do c1.add_device_cgroup_rule("c 81:* rwm") %}
+{% endif %}
+
+{# Devices#}
+{% for device in values.octoprint.devices %}
+  {% do c1.devices.add_device("/dev/"+device, "/dev/"+device) %}
```
stavros-k, line 45: This is way too unsafe.

ix-dev/community/octoprint/templates/docker-compose.yaml
```diff
@@ -0,0 +1,71 @@
...
+{% do c1.environment.add_env("MJPG_STREAMER_INPUT", "-r " + values.octoprint.video_resolution) %}
+{% do c1.environment.add_user_envs(values.octoprint.additional_envs) %}
+
+{# Device CGroup Rules #}
+{% if values.octoprint.auto_tty_enable %}
+  {% do c1.add_device_cgroup_rule("c 188:* rwm") %}     
+{% endif %}
+{% if values.octoprint.auto_video_enable %}
+  {% do c1.add_device_cgroup_rule("c 81:* rwm") %}
+{% endif %}
```
stavros-k, line 41: Any docs about those?

ix-dev/community/octoprint/templates/docker-compose.yaml
```diff
@@ -0,0 +1,71 @@
...
+
+{# Network Configuration #}
+{% do c1.ports.add_port(values.network.web_port, 80) %}
+
+{# Healthcheck #}
+{% do c1.healthcheck.use_built_in() %}
+
+{# Security Context Configuration #}
+{% do c1.clear_caps() %}
+{% do c1.remove_security_opt("no-new-privileges") %}
```
stavros-k, line 59: why does it need to escalate privileges after starting?

ix-dev/community/octoprint/ix_values.yaml
```diff
@@ -0,0 +1,9 @@
+images:
+  image:
+    repository: octoprint/octoprint
+    tag: latest
```
stavros-k, line 4: use tagged versions

ix-dev/community/octoprint/app.yaml
```diff
@@ -0,0 +1,38 @@
...
+keywords:
+- 3D
+- printer
+lib_version: 2.1.11
+lib_version_hash: 6ccf6e3fd289c920ef530a4280b8ea13f8488e63928f68e2bbaf6e3dea917c81e
+maintainers:
+- email: 
+  name: 
+  url: 
+name:
```
stavros-k, line 24: why are those empty?

ix-dev/community/octoprint/app.yaml
```diff
@@ -0,0 +1,38 @@
...
+  uid: 0
+  user_name: root
+screenshots:
+- https://camo.githubusercontent.com/ab1cc2e6f5329affb50d39a62d6bbe7995eca6dd6dfcaa0e923cec2706890fb6/68747470733a2f2f6f63746f7072696e742e6f72672f6173736574732f696d672f73637265656e73686f742d726561646d652e706e67
+sources:
+- https://hub.docker.com/r/octoprint/octoprint
+- https://github.com/OctoPrint/OctoPrint
+title: Octoprint
+train: community
+version: 1.10.3
```
stavros-k, line 38: 1.0.0

ix-dev/community/octoprint/templates/docker-compose.yaml
stavros-k: Which ever you pick, make sure you use the new port configuration style (see it-tools for example)

stavros-k (changes requested):
Hello, left some comments

ix-dev/community/octoprint/templates/docker-compose.yaml
phillipkremer: I changed it, so that only tty and video devices are able to be connected.

ix-dev/community/octoprint/templates/docker-compose.yaml
phillipkremer: 188 are USB/RS232 converters and 81 are webcam devices. Do you need any other documentation ?

ix-dev/community/octoprint/templates/docker-compose.yaml
phillipkremer: Removed it

ix-dev/community/octoprint/templates/docker-compose.yaml
phillipkremer: Changed it

ix-dev/community/octoprint/templates/docker-compose.yaml
phillipkremer: Changed it

ix-dev/community/octoprint/ix_values.yaml
phillipkremer: Changed it

ix-dev/community/octoprint/app.yaml
phillipkremer: I wasn't sure if I should use the standart truenas entrys or if I enter my information.
Which do you prefer ?

ix-dev/community/octoprint/app.yaml
phillipkremer: Changed it

ix-dev/community/octoprint/templates/macros/mknod_devices.sh
```diff
@@ -0,0 +1,45 @@
...
+    mknod /dev/video2 c 81 2
+    mknod /dev/video3 c 81 3
+    mknod /dev/video4 c 81 4
+    mknod /dev/video5 c 81 5
+    mknod /dev/video6 c 81 6
+    mknod /dev/video7 c 81 7
+    mknod /dev/video8 c 81 8
+    mknod /dev/video9 c 81 9
+    echo "Created video devices"
+fi
```
stavros-k, line 41: I'm still not sure why we need to make those devices.
If a user mounts a device it will be present there. Why you need to create it?
And I also doubt the cgroup rules are needed. Considering it already runs as root.
Are there any upstream docs that specify that you need to do this in the container?

phillipkremer: My intention was, that users are able to add devices in two ways:
(a) Look up device to add in truenas shell and add it via devices. In this case cgroup rules are not needed and will not be applied. Also no devices will be mknoded.
(b) Check auto add devices. In this case cgroup rules are needed to make the devices availiable to the container and devices will be mknoded. 
I added so many devices, cause it may be the case, that more than one printer is connected to the host. But realistically nobody will use octoprint that way. Will reduce the number of devices which will be mknoded. 
I pretty much got all the information i needed from this: 
https://www.reddit.com/r/octoprint/comments/klxzpr/docker_mapping_printer_to_usb_port/

stavros-k: I prefer to only have the (a) option only.
if users want to automagically add everything, they can use the yaml option and use privileged which I dont recommend.
Mapping a device once is not a big deal.

stavros-k:
Note: see latest changes here https://github.com/truenas/apps/pull/1480
(Just some small addition on the UI side)

ix-dev/community/octoprint/templates/macros/mknod_devices.sh
phillipkremer: The downside of this option is, that you cannot hotplug (or turn off) your printer, while octoprint is running. 
I would prefer option (b) cause it provides the best ux in my opinion. Frankly I would use yaml option with c group rules if only option (a) is implemented in an available app. Downsite would be, you had to go everytime in into app shell and mknod the devices. Privileged isn't required. Any second thoughts ? Otherwise I would remove option b.

stavros-k: I've talked internally, and option `b` won't be accepted for the time being.

phillipkremer: Allright

Craftzman7:
For anyone wondering, this seems to be continued in #1509

PR #1435, opened 2025-01-23
Immich update app.yaml to v1.125.1

bykof:


stavros-k:
This does not update the app.
Also there is no need to do it, it is done automatically.

Thanks

Issue #787, opened 2024-11-03
Docker Allow IP address binding


franco-martin:
According to the [official docs](https://docs.docker.com/engine/network/#published-ports) we should be able to bind a container to a specific IP address.
This is super useful for applications that should only be exposed to a particular network and for reverse proxying from one network to another.

fardog:
This would be an extremely helpful addition for those of us that want to bind a service to another interface, say for isolating services on a VLAN. Until this is available, I'm using custom docker-compose files for most of my services—which works fine for me and I'm happy enough with the option—but it would be nice to be able to use the apps catalogue.

stavros-k:
Hello, yes this is something that I want to add as well.
Most of the work is done on apps side, but needs some bits on the middlware/ui as well.

fardog:
That's fantastic; glad to hear it. Thanks!

Spunky17:
Not sure if another request is needed for this or if it falls here but binding to another containers network stack would be useful as well; such as in the case of sabnzbd or transmission binding to another containers running VPN.  If needed I can open an additional request.

PR #1443, opened 2025-01-23
minio: add option to set host ip

Emiliaaah:
This PR adds options to Minio to let users configure on which IP the app should be listening on the host.

#### known issues

This doesn't let you bind to ports that are already in use (even if they are listening on a completely different ip).
So if the Truenas webui is listening on `192.168.1.1:443` you're not able to let Minio listen on port `443` on for example `192.168.1.2`. This issue seems to be related [this line](https://github.com/truenas/middleware/blob/1b71ee8eb2df72dd97609ad1db584cc8c1f29215/src/middlewared/middlewared/plugins/apps/schema_validation.py#L137) in the [truenas/middleware](https://github.com/truenas/middleware) apps schema validator

### DIFF
--- a/ix-dev/stable/minio/questions.yaml
+++ b/ix-dev/stable/minio/questions.yaml
@@ -106,6 +106,15 @@ questions:
     schema:
       type: dict
       attrs:
+        - variable: api_ip
+          label: API IP
+          description: The IP for the MinIO API.
+          schema:
+            type: string
+            default: 0.0.0.0
+            required: true
+            $ref:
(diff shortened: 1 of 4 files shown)


Emiliaaah:
I fixed a small naming inconsistency in one of the descriptions

stavros-k:
Hello, this will be done for all apps https://github.com/truenas/apps/issues/787
The schema will change and migrations will happen. I can't accept this now as it will create edge cases in migrations.
Thanks

PR #1453, opened 2025-01-25
esphome: set two platformio env vars to allow esphome to run as non-root

FeatherKing:
ive tested compiling several device types with these settings. It would be a bit more clean if we could override permissions on /piolibs in the container. The platformio builds fail because they cant read that folder. When i set the env var PLATFORMIO_GLOBALLIB_DIR to any other dir, builds work. I assume it doesnt find that local cache in the container and downloads a new copy.

This is working though. I added a new test case for running as root (which also worked). If there is a way to force uid 568 on /piolibs im happy to update the PR.

### DIFF
--- a/ix-dev/community/esphome/app.yaml
+++ b/ix-dev/community/esphome/app.yaml
@@ -27,4 +27,4 @@ sources:
 - https://github.com/esphome/esphome
 title: ESPHome
 train: community
-version: 1.0.0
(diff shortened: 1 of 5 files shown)


domoritz:
Awesome. Thanks for the quick fix!

stavros-k (approved):

PR #1480, opened 2025-01-28
apps: allow selecting host-ip for ports with new config format

stavros-k:
New apps and new ports on existing apps that added using the new config format, can now select ip to bind that port.
When 25.04 is released, migrations for existing apps/ports will follow. 

Host IP features will also work on 24.10 as well.
For the apps/ports that need migration (that will come later), 25.04 will be **required**

Relates to https://github.com/truenas/apps/issues/787, but not fully completes it yet.


kmoore134 (approved):

FeatherKing:
@stavros-k can we pass the host ip as an env var using this variable?
(related to https://github.com/truenas/apps/discussions/1455)

Issue #1512, opened 2025-01-30
Firefly III uses the same port as Plex by default


TimvanDijkNL:
Firefly III and Plex use the same port by default: 30027

Of course the port can be changed, but I was under the impression that all the catalog apps had unique ports assigned to them ?

stavros-k:
Plex uses 32400 by default.

But yes there are some apps that might have duplicate ports between them.
It will be fixed when I go over all ports while doing #787 

Thanks

PR #1517, opened 2025-01-31
apps: fix default ports

stavros-k:
Changed default ports to ensure unique default ports per app.
Added a basic script to check for duplicates and fail CI if found, which also gives the next 5 free ports

Closes https://github.com/truenas/apps/issues/1512


yocalebo (changes requested):
listdir is painfully slow. scandir is the preferred method

yocalebo (approved):

Issue #1538, opened 2025-02-03
Unable to import company backup in Invoice Ninja


JacobDB:
I've been trying to get my old Invoice Ninja 4 install migrated in to TrueNAS Scale, which has required me to get the data in to an importable format. I managed to get the latest v4 installed locally, and migrated the data in to a local v5 install, then imported that in to their hosted platform, and now exported that, so I have a properly formatted company backup from the latest version of Invoice Ninja ready to restore.

Every time I run the restore process, it errors out. There seems to be some kind of permissions issue going on. I see the files being uploaded, but when they're extracted to `/tmp`, nothing is output. below is the error I'm seeing.

```
2025-02-03 19:06:59] production.ERROR: Backup file does not exist, or is corrupted. {"exception":"[object] (App\\Exceptions\\NonExistingMigrationFile(code: 0): Backup file does not exist, or is corrupted. at /app/app/Jobs/Company/CompanyImport.php:454)
[stacktrace]
#0 /app/app/Jobs/Company/CompanyImport.php(318): App\\Jobs\\Company\\CompanyImport->unzipFile()
#1 /app/vendor/laravel/framework/src/Illuminate/Container/BoundMethod.php(36): App\\Jobs\\Company\\CompanyImport->handle()
#2 /app/vendor/laravel/framework/src/Illuminate/Container/Util.php(43): Illuminate\\Container\\BoundMethod::Illuminate\\Container\\{closure}()
#3 /app/vendor/laravel/framework/src/Illuminate/Container/BoundMethod.php(95): Illuminate\\Container\\Util::unwrapIfClosure()
#4 /app/vendor/laravel/framework/src/Illuminate/Container/BoundMethod.php(35): Illuminate\\Container\\BoundMethod::callBoundMethod()
#5 /app/vendor/laravel/framework/src/Illuminate/Container/Container.php(694): Illuminate\\Container\\BoundMethod::call()
#6 /app/vendor/laravel/framework/src/Illuminate/Bus/Dispatcher.php(126): Illuminate\\Container\\Container->call()
#7 /app/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(170): Illuminate\\Bus\\Dispatcher->Illuminate\\Bus\\{closure}()
#8 /app/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(127): Illuminate\\Pipeline\\Pipeline->Illuminate\\Pipeline\\{closure}()
[...]

PR #1540, opened 2025-02-03
invoice-ninja: fix storage mounts

stavros-k:
~I was started implementing with another image and upstream released a new one before the initial PR was merged and it seems I never updated the storage location for the new image.
https://github.com/invoiceninja/dockerfiles/blob/24d1f9cd2ec8062369447e1c73bf93dbba35e462/debian/docker-compose.yml#L13~

~This fixes the path~

---

Waiting for this to finalize, to make any changes needed.

https://github.com/invoiceninja/dockerfiles/pull/698

### DIFF
--- a/ix-dev/community/invoice-ninja/app.yaml
+++ b/ix-dev/community/invoice-ninja/app.yaml
@@ -63,4 +63,4 @@ sources:
 - https://github.com/invoiceninja/dockerfiles
 title: Invoice Ninja
 train: community
-version: 1.0.15
+version: 1.0.16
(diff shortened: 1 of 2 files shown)


stavros-k:
See https://github.com/truenas/apps/issues/1538#issuecomment-2688891273

kreyastus:
Hi @stavros-k , I've installed Invoice Ninja, but I'm having trouble with anything that relies on persistent storage.  It seems like this PR might have helped, as the "octane" version of Invoice Ninja is using /app/storage rather than /var/www/html/storage - would you mind taking another look at this?  The "worker" container seems to be writing to /app/storage fine, but as it's the "invoice-ninja" container that serves web traffic, the user just gets a 404 when trying to access the file.

kreyastus:
@stavros-k Found a fix and mounting the storage folder as an additional storage with /app/storage as the path it works perfectly now.

PR #1551, opened 2025-02-05
discription for Server IP extended

schmieeed:
I tried to get it working over an our and then I found out the mqtt:// prefix is missing


stavros-k (changes requested):
Please only modify files under `ix-dev/` directory.

stavros-k:
See #1563

PR #1560, opened 2025-02-06
Add Monero Node (monerod) app

Keeqler:
Hi,

We at [MAGIC Grants](https://magicgrants.org) are adding a Monero app.

Monero is a popular cryptocurrency. Monerod is the Monero node software, which is included in this initial version. Monero-wallet-rpc is a programmatic wallet software that we have also included in this initial version.

We have been able to run the programs with the rendered Docker Compose file. We figure that now is a good time to check on this initial app to ensure that the PR matches expectations, and if so, then we will continue to add additional Monero-related programs to the app. Top of mind are a Monero block explorer and monero-lws (a server program that allows for more efficient lightweight wallets). Tor and i2p connectivity, likely through dedicated separate apps that can talk to this one, are also high priority for us.

You can reach me out at artur@magicgrants.org.

Thanks!


stavros-k:
Hey, sorry for the delay, Will review hopefully in the next few days.

Just tried to finish some migrations, so I won't mix up old/new api's and make sure we don't finish this app with old apis and require migrations.

stavros-k:
I'll be reviewing it tomorrow, is it fine if I push some changes (mostly some API changes).?

stavros-k:
Hello, I've reworked it fully as it was not matching code-style of this repo nor it matched the UI of all the other apps which users are familiar with.

Also I have a left a lot of comments/"fixme" left here and there.
But it seems that I can't push the changes as you either used an organization for the fork or you didnt check the checkbox that allows maintainers to push changes.

I'll open a new PR in this repo so you can check it out.

Keeqler:
Thanks for the rework! I'm taking a look at it.

We figured it would be more convenient putting the wallet and onion into one, but breaking them into multiple apps is totally fine if you prefer.

stavros-k:
Lets split them up? It will clean up the UI making it easier to spin up each one

Keeqler:
I made the requested changes and removed all the wallet-related code, I will create another PR for the wallet when I have it.

I invited you to our forked repo in case you need to do any more changes.

Thanks!

stavros-k:
Okay 
- flattened the UI a bit, as it didnt make much sense to had that nesting now with single container.
- cleaned up some files.

Can you confirm that p2p is static inside the container? ie for rpc port, we can change it but p2p does not look like its changable.

(which is why I added the internal_p2p_port const.)

Besides this, I think its good to be merged.

stavros-k:
Ah and two more things, 
- should we rename the app (id) to `monerod`? (`name:` and directory name are the ones that need to change), `title: ` can still be anything.
- is there a link for prefered icon? I can upload that on our bucket

SamsungGalaxyPlayer:
On the last point, the Monero M logo should suffice:

https://www.getmonero.org/press-kit/symbols/monero-symbol-on-white-480.png
https://www.getmonero.org/press-kit/symbols/monero-symbol-on-white-800.png
https://www.getmonero.org/press-kit/symbols/monero-symbol-on-white-1280.png

Keeqler:
> Can you confirm that p2p is static inside the container? ie for rpc port, we can change it but p2p does not look like its changable.

The p2p port is actually changeable on monerod, but since `18080` is the default port it's ok to use a const.

> should we rename the app (id) to monerod?

Yes, `monerod` is fine.

Also, what are the options we have to point the Monero Wallet app to the Monero Node app RPC port so they can communicate? Using host mode and pointing Wallet to the machine's IP would be easy, but host mode is often a bad practice.

stavros-k:
> > Can you confirm that p2p is static inside the container? ie for rpc port, we can change it but p2p does not look like its changable.
> The p2p port is actually changeable on monerod, but since `18080` is the default port it's ok to use a const.

What is the flag? I like to keep host port and container port in sync so when host network is used, it wont cause confusion or bind a used port.

> Also, what are the options we have to point the Monero Wallet app to the Monero Node app RPC port so they can communicate? Using host mode and pointing Wallet to the machine's IP would be easy, but host mode is often a bad practice.

Maybe something like that?
https://github.com/truenas/apps/blob/f3e19c4421d00f657442378e83249cafa2b99982/ix-dev/dev/truenas-webui/templates/docker-compose.yaml#L11
It will resolve to the docker interface on the host, port will default to what we have default here, but users will be able to override it if they have changed it on this app.

Keeqler:
> What is the flag?

`--p2p-bind-port`

> Maybe something like that?
...

I will try that. Thanks!

stavros-k (approved):

stavros-k:
Thanks! 
And sorry for not making the review all at once, lots of files and was finding more every time I looked at them :D

PR #1581, opened 2025-02-09
open-speed-test: minor description correction

swensorm:
Removed copied Zigbee to MQTT text from open-speed-test description 🙈 .  Noticed this when browsing all the cool new apps!

Thanks!

P.S. Another suggestion; Firefly-III should probably be in the new `financial` category, but didn't know if there's implications with just changing it.


stavros-k (approved):
Thanks!

PR #1586, opened 2025-02-10
tailscale: add reset toggle

### LINKED ISSUE #1528: Tailscale app refusing to start after changing advertised routes
**Description of issue**
Tailscale app will get stuck in a start/halt-loop after changing advertised routes.

**Logs**
Unfortunately lost by me, but they stated that "tailscale up" command cannot be used to change settings without the "--reset" flag if you are not using the same "tailscale up" variables as previous.

**Steps to reproduce**
1. Install Tailscale app
2. Set to authorize with tskey
3. Set to advertise routes
4. Start Tailscale app
5. Edit setting and remove advertised routes
6. Restart Tailscale container

**My solution**
1. Stop looping container.
2. Add "--reset" under "Extra Arguments" in app settings for Tailscale app.
3. Start Tailscale container.
4. Confirm successful authentication and start of services in logs.

**Suggestions**
The app should always pass the "--reset" variable with "tailscale up" to ensure that changed settings are accepted.

stavros-k:
Closes #1528


stavros-k:
Reset is a bit weird, but its just resetting unspecified options to default values.
Fixed description as well

Qubad786 (approved):

PR #1550, opened 2025-02-05
Migrate apps to support host-ip per port

stavros-k:

PR #1606, opened 2025-02-12
Migrate apps to support host-ip per port (part 2)

stavros-k:
Part 2 of #1550 
Mostly because github started hidding files be default for review.


stavros-k:
This can be reviewed and merged into #1550

PR #1607, opened 2025-02-12
lms: use stable instead of dev version

### LINKED ISSUE #1348: lyrion-music-server installs development version instead of latest release version

stavros-k:
Closes #1348

While currently there isn't a `XXX-stable` tag. 
We can switch back to 9.0.2 tag which is "stable" and once upstream releases a new `XXX-stable` tag, renovate will only pick those.

That being said, from what I understand `-stable` tags will appear on 9.1.x and later versions (once it hits stable).
Until then stable versions will be ignored (ie 9.0.3)


Qubad786 (approved):
I assume there are no issues with us downgrading the app version for the people who are already on this

stavros-k:
Shouldn't be any issues afaik.

PR #1648, opened 2025-02-17
esphome: remove ESPHOME_DASHBOARD_USE_PING override

domoritz:


ix-dev/community/esphome/templates/docker-compose.yaml
```diff
@@ -10,7 +10,6 @@
 {% do c1.healthcheck.set_test("curl", {"port": values.network.web_port.port_number, "path": "/version"}) %}
 {% do c1.set_command(["dashboard", values.consts.config_path, "--port", values.network.web_port.port_number]) %}
 
-{% do c1.environment.add_env("ESPHOME_DASHBOARD_USE_PING", true) %}
```
stavros-k, line 13: Lets make it configurable? defaulting to true?

domoritz: I think if we can make it work without it, the default should be false. People can override it with an environment variable.

stavros-k (approved):

PR #1649, opened 2025-02-17
Updated instances of wordpress to WordPress

johndcoy:
All instances of "WordPress" should be properly cased.


stavros-k:
Hello, this is touching generated files. And there is already a PR fixing this #1647

PR #1667, opened 2025-02-19
esphome: bump to 2025.2

domoritz:


domoritz:
Ah, happening already in https://github.com/truenas/apps/pull/1664. Although, I don't know why it just bumps the patch version.

stavros-k:
The app **package** has no changes that would require bump on minor or major.
This version is following the changes done in the template/ui.

stavros-k:
fyi, updates happen automatically. 
dont manually open PRs for it.
if you notice no PRs opened automatically, open an issue so we can fix that instead

PR #1704, opened 2025-02-24
Add Monero Wallet RPC app

Keeqler:


stavros-k:
Moved `ssl_enabled` under `use_remote_node = true`. At least thats how it is setup to behave on questions schema.

stavros-k:
Should I use the same icon as `monerod`?

Keeqler:
> Should I use the same icon as monerod?

Yes please!

PR #1744, opened 2025-02-27
Monerod: fixes, new question descriptions

Keeqler:


ix-dev/community/monerod/templates/docker-compose.yaml
```diff
@@ -70,9 +80,13 @@
 {% do perm_container.add_or_skip_action("bitmonero", values.storage.bitmonero, perm_config) %}
 
 {% if not values.network.host_network %}
-  {% do monerod.add_port(values.network.monerod_p2p_port) %}
-  {% do monerod.add_port(values.network.monerod_rpc_port) %}
-  {% do monerod.add_port(values.network.monerod_restricted_rpc_port) %}
+  {% if values.network.monerod_p2p_port.port_number %}
```
stavros-k, line 83: Can you please clarify what we are doing here?

ix-dev/community/monerod/templates/docker-compose.yaml
```diff
@@ -15,14 +14,18 @@
   {% endfor %}
 {% endfor %}
 
+{% set p2p_port_number = values.network.monerod_p2p_port.port_number or values.consts.default_monerod_p2p_port_number %}
+{% set restricted_rpc_port_number = values.network.monerod_restricted_rpc_port.port_number or values.consts.default_monerod_restricted_rpc_port_number %}
```
stavros-k, line 18: Why have the default here? The port number will always be defined from the user making the default here unneeded

ix-dev/community/monerod/templates/docker-compose.yaml
```diff
@@ -15,14 +14,18 @@
...
+
+{% do monerod.healthcheck.set_test("curl", {"port": restricted_rpc_port_number, "path": "/get_height"}) %}
+
 {# This way we get rid of fixuid #}
 {% do monerod.set_entrypoint(["monerod", "--non-interactive"]) %}
 
 {% set commands = namespace(x=[
-  "--rpc-bind-port=%d" | format(values.network.monerod_rpc_port.port_number),
```
stavros-k, line 22: Does removing this, stop binding the port completely?
Also what does this port does? Is it not needed?

ix-dev/community/monerod/templates/docker-compose.yaml
Keeqler: We found that if by whatever reason the user wants to set the a port bind mode to "None", an error will be thrown because `values.network...port_number` won't be defined when using it in `commands`.

ix-dev/community/monerod/templates/docker-compose.yaml
Keeqler: I thought this would throw an error similar to when using `.port_number` in `commands`, I just realized it won't and I'm now removing the `if` statements. Thanks.

ix-dev/community/monerod/templates/docker-compose.yaml
stavros-k: What if we leave it there, but we default to bind_mode: "" ?
That way, users can still expose it if they want easily.

ix-dev/community/monerod/templates/docker-compose.yaml
stavros-k: Hmm, now that I'm thinking it, I need to fix/update that show_if everywhere.

So, port_number is relevant in cases:
- bind_mode = published
- bind_mode = expose, AND internal app port is changeable
- bind_mode = "" (none), AND internal app port is changeable

So, port_number is IRrelevant in cases:
- bind_mode = expose AND internal app port is NOT changeable
- bind_mode = "" (none) AND internal app port is NOT changeable

Proper fix would be:

- Remove show_if competely if the internal port is changeable
- Set show_if to bind_mode = published when port is NOT changeable

(I'll make the PR to change it everywhere, btw don't worry)

stavros-k: https://github.com/truenas/apps/pull/1754/files

stavros-k: Pushed a commit for this here as well, let me know what you think.

ix-dev/community/monerod/questions.yaml
```diff
@@ -563,7 +570,7 @@ questions:
                 description: CPUs limit for Monero Node.
                 schema:
                   type: int
-                  default: 2
```
stavros-k, line 566: this will hard fail to deploy if the machine that is deployed does not have at least 4 cores.
docker does not allow exceeding

ix-dev/community/monerod/templates/docker-compose.yaml
Keeqler: Looks good! Thanks.

ix-dev/community/monerod/templates/docker-compose.yaml
Keeqler: Yes, thanks for the help!

I believe that if published, rpc also needs to bind to 0.0.0.0 since it is 127.0.0.1 by default?

stavros-k: publish won't change where the **internal** binding happens, will just add the port mapping under `ports:` in docker-compose.

stavros-k (approved):

ix-dev/community/monerod/templates/docker-compose.yaml
stavros-k: if we need a flag to change that, we can add it like so

`flag=%s | format("0.0.0.0" if values.network.rcp_port.bind_mode != "" else "127.0.0.1")`

Keeqler: Done in #1758

PR #1747, opened 2025-02-27
Changed UID of Odoo user

Nightbringer01:
For some reason the UID of odoo user in the docker container changed it's UID. changed the UID to 100 for the config

### DIFF
--- a/ix-dev/community/odoo/app.yaml
+++ b/ix-dev/community/odoo/app.yaml
@@ -20,7 +20,7 @@ run_as_context:
 - description: Odoo runs as non-root user.
   gid: 101
   group_name: odoo
-  uid: 101
+  uid: 100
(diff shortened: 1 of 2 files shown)


stavros-k:
Odoo 17 does not have this change.
Odoo 18 does, but it will be a major bump anyway, as every odoo version is not upgradable from another.

Nightbringer01:
yeah figured to have this pushed to the 18.x branch. not sure if the plan is to merge this onto the main branch though. 
Figured it might help just figuring out what the problem was.

stavros-k:
Will need to first go over release notes see if we need to change other things first.
Thanks tho

stavros-k:
Linked to the other PR, closing this for now
Thanks

PR #1785, opened 2025-03-05
Add MariaDB as App

tina-junold:
I was curious about the fact that MariaDB was not in the Apps section while it is as dependency. For the fact that I'd like to see it as app, I added it based on the Postgres configuration.


ix-dev/community/mariadb/ix_values.yaml
```diff
@@ -0,0 +1,16 @@
+images:
+  image:
+    repository: mariadb
+    tag: 11.7
+  v1104_image:
+    repository: mariadb
+    tag: 11.4
+  v1011_image:
+    repository: mariadb
+    tag: 10.11
```
stavros-k, line 10: Is there really a reason to run 11.4 instead of 11.7? 
I think only major versions matter right?

ix-dev/community/mariadb/questions.yaml
```diff
@@ -0,0 +1,392 @@
...
+      attrs:
+        - variable: tcp_port
+          label: TCP Port
+          description: The port for MariaDB TCP
+          schema:
+            type: int
+            default: 3306
+            required: true
+            $ref:
+              - definitions/port
```
stavros-k, line 101: See newer apps like umami for new style ports,
changes requierd in compose and test values as well

ix-dev/community/mariadb/ix_values.yaml
tina-junold: while 11.7 is latest/stable, 11.4 is lts, that's why i added it

stavros-k: I think we should go with major versions only.
Given its community catalog, I don't think LTS makes much sense here.

ix-dev/community/mariadb/questions.yaml
tina-junold: you mean the whole network block with web_port (should it be named web port, since its not http), bind_mode, port_number and host_ips?

tina-junold:
@stavros-k One additional question: should i add in questions.yaml an auto_update block since "MARIADB_AUTO_UPGRADE" is not part of the configuration yet.

ix-dev/community/mariadb/questions.yaml
stavros-k: the name can stay tcp! but the rest should change (the data structure)

stavros-k:
> @stavros-k One additional question: should i add in questions.yaml an auto_update block since "MARIADB_AUTO_UPGRADE" is not part of the configuration yet.

Lets add a toggle yes, I think its safe to defaut to true (we do that on dependencies without issues so far)

tina-junold:
Are there any examples for the Auto-Upgrade toggle?

ix-dev/community/mariadb/questions.yaml
tina-junold: updated

ix-dev/community/mariadb/ix_values.yaml
tina-junold: i removed lts

ix-dev/community/mariadb/ix_values.yaml
```diff
@@ -0,0 +1,13 @@
+images:
+  image:
+    repository: mariadb
+    tag: 11.7
+  v10_image:
+    repository: mariadb
+    tag: 10
```
stavros-k, line 7:
```suggestion
    tag: 10.11
```
Lets pin it, and the bot will open PRs for newer versions, so we can do some "smoke test" on ci. :)

stavros-k:
> Are there any examples for the Auto-Upgrade toggle?

Just a boolean field (similar to host_network)
And set the env in compose file

ix-dev/community/mariadb/templates/docker-compose.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+{% do c1.healthcheck.set_test("mariadb", {"port": values.network.tcp_port.port_number}) %}
+
+{% if perm_container.has_actions() %}
+  {% do perm_container.activate() %}
+  {% do c1.depends.add_dependency(values.consts.perms_container_name, "service_completed_successfully") %}
+{% endif %}
+
+{% do tpl.portals.add_portal({"port": values.network.tcp_port.port_number}) %}
+
```
stavros-k, line 35:
```suggestion
```
Lets remove it, considering there is no webui!

ix-dev/community/mariadb/app.yaml
```diff
@@ -0,0 +1,30 @@
+app_version: '11.7'
+capabilities: []
+categories:
+- database
+description: The MariaDB Server is one of the most popular open source relational
```
stavros-k, line 5: Maybe use the caption from the "About" section in their github?
https://github.com/MariaDB/server

I find the "most popular" bit misleading lol
> postgres disagrees :D

stavros-k: We can have first sentence here and full on the readme

tina-junold: done

stavros-k (approved):

PR #1825, opened 2025-03-11
remove migration from k8s

stavros-k:
This will be merged into #1550. 
Once #1550 is merged migrations from k8s are not going to work.


stavros-k:
Done in #1550

PR #1830, opened 2025-03-11
Fix unifi controller to open the discover port

Mephi00:
The current community app for the unifi controller can not adopt new unifi devices.
According to the [Required Ports Reference](https://help.ui.com/hc/en-us/articles/218506997-Required-Ports-Reference) the UDP Port 10001 is required for 'Device discovery during adoption.'

This PR adds the missing port to the community app. During a local test I was able to discover a unifi switch.


stavros-k (approved):

stavros-k:
Thanks, make sure to change the bind mode after you update, as the default is not to publsuh

PR #1863, opened 2025-03-14
Add JDownloader app

### LINKED ISSUE #1573: app Suggestion: JDownloader
It would be very interesting to have this download manager in Truenas

https://jdownloader.org/download/index

jondycz:
Added JDownloader app based on [jaymoulin/jdownloade](https://hub.docker.com/r/jaymoulin/jdownloader)

Used cloudflared as the base app to create this one. `lib_version` and `lib_version_hash` in `app.yaml` as well as the icon need to be updated.

icon link: https://jdownloader.org/_media/knowledge/wiki/jdownloader.png

fixes #1573


stavros-k:
Any reason not to use https://hub.docker.com/r/jlesage/jdownloader-2 ?
You can copy handbrake and you will need slight adjustments only.

btw you should use the new style ports (see Umami or zigbee2mqtt).
This affects questions, test values and docker-compose.

Thanks

jondycz:
The image I used is headless and is controlled through the MyJdownloader web portal. The image you suggest is a desktop GUI running through noVNC. I used both and in my opinion the headless one offers better user experience.

Could you please elaborate on what lib_version and lib_version_hash is and provide icon URL? Thanks.

stavros-k:
You can copy/paste existing lib_version and hash from other apps that use 2.1.17 (current).

Version is the one that you want to use from the `library` directory, currently only 1 version exists.
Hash is the hash of that library directory (hash exists in the hashes.yaml)

stavros-k:
Let me know once you update PR to use the new port style, so I can start reviewing.
Thanks

stavros-k:
https://media.sys.truenas.net/apps/jdownloader/icons/icon.png

ix-dev/community/jdownloader/app.yaml
```diff
@@ -0,0 +1,33 @@
...
+- media
+- download
+- files
+lib_version: 2.1.16
+lib_version_hash: dac15686f882b9ce65b8549a3d5c0ed7bafe2df7a9028880d1a99b0ff4af1eff
+maintainers:
+- email: dev@ixsystems.com
+  name: truenas
+  url: https://www.truenas.com/
+name: jdownloader
```
stavros-k, line 20: should we name it `jdownloader2`?

ix-dev/community/jdownloader/templates/docker-compose.yaml
```diff
@@ -0,0 +1,47 @@
...
+{% do c1.healthcheck.disable() %}
+{% set cmd = namespace(x=[]) %}
+{% if values.ci %}
+  {% do cmd.x.append("--hello-world") %}
+{% else %}
+  {% do cmd.x.extend(values.jdownloader.additional_args) %}
+  {% do cmd.x.append("run") %}
+{% endif %}
+
+{% do c1.set_command(cmd.x) %}
```
stavros-k, line 27: From a quick look on the upstream image, overriding command will make the container not start. as the image uses command instead of entrypoint.

ix-dev/community/jdownloader/templates/docker-compose.yaml
```diff
@@ -0,0 +1,47 @@
...
+    {% do tpl.funcs.fail("JDownloader already inserts [%s] as an argument, please remove it from [jdownloader.additional_args]" | format(arg)) %}
+  {% endif %}
+{% endfor %}
+
+{% set c1 = tpl.add_container(values.consts.jdownloader_container_name, "image") %}
+{% set perm_container = tpl.deps.perms(values.consts.perms_container_name) %}
+{% set perms_config = {"uid": values.run_as.user, "gid": values.run_as.group, "mode": "check"} %}
+
+{% do c1.set_user(values.run_as.user, values.run_as.group) %}
+{% do c1.healthcheck.disable() %}
```
stavros-k, line 18: Can't we curl/wget any api endpoints?

stavros-k: Maybe this? `/jd/version`
https://my.jdownloader.org/developers/#tag_206

ix-dev/community/jdownloader/templates/test_values/basic-values.yaml
```diff
@@ -0,0 +1,25 @@
...
+  limits:
+    cpus: 2.0
+    memory: 4096
+
+ci: true
+
+cloudflared:
+  tunnel_token: ""
+  additional_args: []
+  additional_envs: []
```
stavros-k, line 11: probably these needs updating

ix-dev/community/jdownloader/templates/test_values/basic-values.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+network:
```
stavros-k, line 13: missing the api port?

jondycz:
Hey, I was just about to work on the rewrite to the "new style" but you seemed to review it before the change. As you stated, the new style also affects the test values. Should I still pursue the new style of are we gonna stick with the old one then?

stavros-k:
You can proceed with the change, rest of the review stands regardless of the port change

ix-dev/community/jdownloader/app.yaml
jondycz: I am not sure. The original one has been replaced by the v2 in 2011 and i doubt anyone is using that version to this day. Even the official website doesn't mention the original version. So it is really up  to the maintainers. https://en.wikipedia.org/wiki/JDownloader

ix-dev/community/jdownloader/templates/docker-compose.yaml
jondycz: Perhaps. Or uptime. Will use your suggestion.

jondycz: Though, it seems like it is required to call these endpoints through their cloud service called MyJdownloader and calling it directly is unsupported and deprecated:

```
Communication
API Endpoint: https://api.jdownloader.org
To avoid problems with NAT and Portforwarding, the communication between the frontends and JDownloader is routed through our Connection-Server. IF available, the frontend will try to establish a direct connection between the frontend and JDownloader after validating the account credentials and initializing the encryption tokens.
Pro Tip: It's possible to access the JDownloader API directly (Bypass our server) by enabling the so called 'Deprecated API' in the Advanced Options.
```

Perhaps this is something that could be added later on?

ix-dev/community/jdownloader/app.yaml
stavros-k: In that case, we either have `jdownloader` everywhere or `jdownloader2`/`JDownloader 2` everywhere, not mixed

ix-dev/community/jdownloader/templates/docker-compose.yaml
stavros-k: Isn't those endpoints for the localAPI?

ix-dev/community/jdownloader/templates/docker-compose.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+{% set reserved_args = [] %}
+{% if values.jdownloader.additional_args != values.jdownloader.additional_args|unique|list %}
+  {% do tpl.funcs.fail("JDownloader [jdownloader.additional_args] contains duplicate values [%s], please remove them" | format(values.jdownloader.additional_args|join(', '))) %}
+{% endif %}
+{% for arg in values.jdownloader.additional_args %}
+  {% if arg in reserved_args %}
+    {% do tpl.funcs.fail("JDownloader already inserts [%s] as an argument, please remove it from [jdownloader.additional_args]" | format(arg)) %}
+  {% endif %}
+{% endfor %}
```
stavros-k, line 11: Those should go away as well, there aren't args for the user to set

ix-dev/community/jdownloader/app.yaml
jondycz: I mean most sense makes naming it after the upstream docker image, doesn't it? Should i then change all the descriptions to not include the version number in the name?

ix-dev/community/jdownloader/questions.yaml
```diff
@@ -101,6 +101,15 @@ questions:
...
       attrs:
+        - variable: direct_connection_port
+          label: Direct Connection Port
+          description: The port to enable JDownloader Direct Connection
+          schema:
+            type: int
+            default: 3129
+            required: false
+            $ref:
+              - definitions/port
```
stavros-k, line 112: Thats the old port config style,
The new one uses a dictionary.

ix-dev/community/jdownloader/app.yaml
stavros-k: I think having the `2` there would be better. AFAIK people "known" it as v2

jondycz:
> https://media.sys.truenas.net/apps/jdownloader/icons/icon.png

will we also need to rename the icon url to reflect the jdownloader2 name?

ix-dev/community/jdownloader/ix_values.yaml
```diff
@@ -0,0 +1,9 @@
+images:
+  image:
+    repository: jaymoulin/jdownloader
+    tag: 2.1.2
+
+consts:
+  jdownloader_container_name: jdownloader
+  perms_container_name: permissions
+  internal_direct_connection_port: 3129
```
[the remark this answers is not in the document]
stavros-k: yes

stavros-k:
Did some changes to clean it up.

But, I'm not sure I understand how this headless image works.
I need to have an account and have port forward in order to access the application? 
Do I understand it wrong?

jondycz:
You need to have an account indeed. Exposed container ports are only necessary when using direct connection, but you still need an account for that.

stavros-k:
> You need to have an account indeed. Exposed container ports are only necessary when using direct connection, but you still need an account for that.
> 
> You use my.jdownloader.org to control all your instances

Hmm,In the case of this image(headless) you are forced to have an account right?
> There are people accepting that for a self hosted service? :D 

Anyway, I remember using JD2 couple years before (with the jlesage image) and I don't remember needing an account to use it. What's the difference there?

jondycz:
Yes. It's a requirement. 

If you wish to also connect this instance to Home Assistant for example, you will also need to have the account. So for any API calls and integrations, it's required. The image you used is only good for people who are used to the desktop experience and do not want or need any advanced features or automations.

stavros-k:
To be fair jlesage's image also support direct connection
https://github.com/jlesage/docker-jdownloader-2?tab=readme-ov-file#direct-connection

---

I don't mind which one will be added, but I'm trying to make sure we are adding the right one. 
If most people just want the desktop experience, we should go with the jlesage image and you will also have the ability to setup your integrations.

jondycz:
Both have about the same amount of pulls.

The "GUI" of jlesage image is just noVNC, which in my opinion is an ugly solution, and doesn't work well on mobile devices. I also couldn't find an easy way to provide credentials to the jlesage image to enable MyJdownloader (the web panel) - which is just an env variable in the jaymoulin image. The root requirement by the jlesage image is another thing...

Also the "direct access port" cannot be changed from the original based on the docs of jlesage and needs to stay 3129 (that's true for both images)

I suggest you try both if you want to make absolutely sure. But I think the jaymoulin image is lighter (less cluttered) and more easily configurable.

I'm also not sure if home assistant can use direct connection, and even if that is enabled, the port is sent to the MyJdownloader web service. So you can't make it work locally without internet connection anyway.

jondycz:
Yes, you  can access the noVNC GUI without an account. It's basically the same as downloading the desktop app, but... Running on a server. Which to me seems kind of weird to run a "window" app on a server and share that with noVNC. But yes, you're right that it is a use case someone might find useful. In my opinion if you want to run the window version, you should run it on your PC and not on a server. But that's just me. And yes, even the "window" version allows you to connect it to the cloud and control through phone, home assistant etc.

Upon re-reding the docs, the port on the host can be any port.

Again, I feel like servers should run headless software. Running window apps and then transmitting the window through noVNC seems like a hacky solution to me. That's what desktops are for, to run window apps. 

Both images got pros and cons. Either way I'd always choose a rootless option when available even if it means sacrificing rarely used feature. Should we make a poll on discord?

stavros-k:
Wait, how can I access the headless one with a desktop app without connecting to an account?

jondycz:
> Wait, how can I access the headless one with a desktop app without connecting to an account?

You can't. The non-headless version is just a desktop version running in a browser through noVNC.

AdrianJesusSilva:
sorry to comment in a closed PR, but i have problems with the app, i put the configuration as it is asked and JD start without a problem (as far the logs said), but i don't see the instance in the my.jdownloader, any of you have this situation ?

AdrianJesusSilva:
sorry to comment in a closed PR, but i have problems with the app, i put the configuration as it is asked and JD start without a problem (as far the logs said), but i don't see the instance in the my.jdownloader, any of you have this situation ?

jondycz:
> sorry to comment in a closed PR, but i have problems with the app, i put the configuration as it is asked and JD start without a problem (as far the logs said), but i don't see the instance in the my.jdownloader, any of you have this situation ?

Mine works. I didn't do anything other than specify the required parameters (user, pass, instance name) and download path. 🤷

PR #1886, opened 2025-03-18
adjust category for openspeedtest

RemyDG0:
OpenSpeedTest is currently in its own category « network » but an appropriate category « networking » suits also and contains other similar apps 
this update avoids confusion


stavros-k (approved):

RemyDG0:
@stavros-k 
The category has not been updated in the TrueNas app page (at least on my system) did I missed something ?

RemyDG0:
It is now updated.

PR #1916, opened 2025-03-22
add: Tvheadend

### LINKED ISSUE #272: Community Chart Suggestion: Tvheadend
Hi, 

it would be great to get Tvheadend as an official app, esp since when running an SatIP system there seems to be no good free alternative existing to it. I used to run it from some other catalog and some network fiddling is needed to be able to run it with proper SatIP capabilities but once done it runs flawless.

URL: https://tvheadend.org/

Thanks in advance!

jondycz:
Possible issues:
- the docker repository of this image doesnt use semver, neither does the project itself.
- to add dvb support, i got inspired by the controller support of the kasm app, which works by mounting the device as volume

icon: https://cdn.jsdelivr.net/gh/homarr-labs/dashboard-icons/svg/tvheadend.svg

closes #272


jondycz:
Why would you remove gpus? Tvheadend can also transcode streams for better device compatibility.

stavros-k:
> * the docker repository of this image doesnt use semver, neither does the project itself.

eww

> * to add dvb support, i got inspired by the controller support of the kasm app, which works by mounting the device as volume

Does it need to be writeable or read only is enough?

jondycz:
> Does it need to be writeable or read only is enough?

I don't know, I set it like this:
```
devices:
      - /dev/dri:/dev/dri
      - /dev/dvb:/dev/dvb
```

stavros-k (approved):
lgtm
Thanks

PR #1936, opened 2025-03-25
Add Monero LWS app

Keeqler:
This is a Monero light wallet server that can be used with apps such as MyMonero and Edge wallet.

### DIFF
--- a/ix-dev/community/monero-lws/README.md
+++ b/ix-dev/community/monero-lws/README.md
@@ -1,0 +1,3 @@
+# Monero Wallet RPC
+
+[Monero Wallet RPC](https://www.getmonero.org/) is RPC server that allows interaction with Monero wallets programmatically.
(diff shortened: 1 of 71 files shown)


Keeqler:
Will disabling healthchecks pass CI?

LWS requires a connection with a Monero node to be able to accept requests, that wouldn't be possible in CI.

stavros-k:
> Will disabling healthchecks pass CI?

As long as the container does not `exit 1`, yes.
Is there a specific endpoint to hit for heathcheck? (when it is run normally, with a connection)
Or `/` is enough?

stavros-k:
Do we have any icons for this?

SamsungGalaxyPlayer:
The normal Monero icon can be used, thanks!

stavros-k:
Looks good to me.
I need clarification on two things:
- Versioning scheme
- Healthcheck endpoint (should we use `/` or `/get_height` like in `monerod`?)

Keeqler:
`0.3` is the version of the app itself and `0.18` is the version of `monerod` that LWS is built against.

We should use `/` for healthchecks.

stavros-k (approved):
LGTM

PR #1961, opened 2025-03-30
add: ArchiSteamFarm

### LINKED ISSUE #1813: Community App Suggestion: ArchiSteamFarm [app-request]
I'd like to suggest adding ASF, an app used by millions of steam users.

Docs: https://github.com/JustArchiNET/ArchiSteamFarm/wiki/Docker

The installation for the end user should be pretty simple, only requiring users to specify an IPC Password and a port for the web interface.

The installation script would need to additionally create an IPC.config file with the following contents: 
```
{
	"Kestrel": {
		"Endpoints": {
			"HTTP": {
				"Url": "http://*:1242"
			}
		}
	}
}
```

The IPC port chosen by the user can be then mapped to the internal port of 1242

jondycz:
This app by default does not set the IPC password, meaning the app is not password protected by default. By mounting the config folder to the host, it is possible to set the password manually in the config file.

Moreover, the app does not use commonly used env vars to set User and Group IDs, only supports setting the user using their custom ASF_USER variable, which then is used to set both user and group ([reference](https://github.com/JustArchiNET/ArchiSteamFarm/blob/816914dd1068ebbfc68a49f684383d02937ea742/Dockerfile#L106))

Couldn't find an SVG logo, the best alternative to use is probably webP: https://cdn.jsdelivr.net/gh/homarr-labs/dashboard-icons/webp/archisteamfarm.webp

closes #1813


stavros-k:
> Moreover, the app does not use commonly used env vars to set User and Group IDs, only supports setting the user using their custom ASF_USER variable, which then is used to set both user and group ([reference](https://github.com/JustArchiNET/ArchiSteamFarm/blob/816914dd1068ebbfc68a49f684383d02937ea742/Dockerfile#L106))

That is a build time thing, and user is not affected by that variable on runtime.

stavros-k:
Also its not usable rootless either
https://github.com/JustArchiNET/ArchiSteamFarm/issues/2197

jondycz:
> Also its not usable rootless either [JustArchiNET/ArchiSteamFarm#2197](https://github.com/JustArchiNET/ArchiSteamFarm/issues/2197)

Actually you can run as non-root, but requires passing a --user flag 

The docs mention that using this approach, the user is required to handle the permissions themselves - which we can do with our own TRUENAS permissions container.

https://github.com/JustArchiNET/ArchiSteamFarm/wiki/Docker#volume-permissions

stavros-k:
> > Also its not usable rootless either [JustArchiNET/ArchiSteamFarm#2197](https://github.com/JustArchiNET/ArchiSteamFarm/issues/2197)
> 
> Actually you can run as non-root, but requires passing a --user flag
> 
> The docs mention that using this approach, the user is required to handle the permissions themselves - which we can do with our own TRUENAS permissions container.
> 
> https://github.com/JustArchiNET/ArchiSteamFarm/wiki/Docker#volume-permissions

Thanks, didnt see that. Should work now :)

stavros-k:
> Usage of the /healthcheck endpoint is pretty smart. Didn't know it exists.
> 
> Also, could you clarify what you changed about the startup.sh script and the storage mounts? I'm on a phone so I can't properly do file diffs, but you mount a /config and then copy it to /app/config? Is that necessary?

The config is mounted there because currently the permissions container won't touch any directories that contain data.
Mounting the config on the /app/config, makes the directory non-empty and permissions won't apply.
Its more of a "hack" D:

stavros-k (approved):

jondycz:
I also see you converted the config into an object and modify the URL from within your templating engine. Does this mean that we can actually modify any JSON based configurations on startup, even configurations that already exist on the host? 

An example I can see is for example adding the ability to set the IPC password from the trueNAS GUI instead of requiring users to edit the configuration files.

stavros-k:
> I also see you converted the config into an object and modify the URL from within your templating engine. Does this mean that we can actually modify any JSON based configurations on startup, even configurations that already exist on the host?
> 
> An example I can see is for example adding the ability to set the IPC password from the trueNAS GUI instead of requiring users to edit the configuration files.

Yep you can have an yaml object and modify it on demand, and then `tojson`.
Note that any file mounted with `configs.add` will be reset on container restart, so user mods will not persist

jondycz:
Well, there seems fo be more issues with the storage. I deleted my custom app in favor of this one, kept the IX volume and left default settings. After install, it tries to launch, but it shows status stopped. When I force it to start, it just immediately stops without any errors thrown.

I tried starting with an empty directory, which worked. Then u copied over my config files and chowned it to 568, same as before. App couldn't start. There were no errors about failing to do the docker compose up command like there usually is. Any idea?

stavros-k:
Logs? What do they say

jondycz:
I can't really open the logs when the container is immediately stopped. And app_lifecycle.log also didn't log anything

stavros-k:
You can have a shell open and spam right after you hit start

```sh
sudo docker logs ix-archisteamfarm-archisteamfarm-1 -f
```

jondycz:
I managed to open the config in the mere second it runs.

```
2025-03-31 14:57:28.066851+00:00 Copying IPC config  

2025-03-31 14:57:28.067978+00:00 cp: cannot create regular file '/app/config/IPC.config': Permission denied
```

jondycz:
I take it back. It's not caused by copying the files. You can simulate it by installing the app and then restarting. It fails even if the IPC config created by the setup exists. So on each reboot, the file needs to be deleted first manually in order to start the container.

stavros-k:
Check v1.0.1 in ~5min
https://github.com/truenas/apps/pull/1971

PR #1983, opened 2025-04-01
photoprism: run rootless

### LINKED ISSUE #1956: Photoprism: allow running rootless
Why does Photoprism need to run as root with the ability to override permissions and kill processes? The default user in the configuration flow is the `apps` user, but the manifest states that it needs to run as root. Is this a mistake in the manifest or is the app actually running as root?

cc @stavros-k

stavros-k:
Closes #1956


stavros-k:
Closing for now, as this will break other parts of the app like gpu transcoding

PR #2078, opened 2025-04-14
Add Bitcoin Node app

Keeqler:
Icon
image


ix-dev/community/bitcoind/questions.yaml
```diff
@@ -0,0 +1,874 @@
...
+          description: |
+            Enter the local IP of this machine if you are running the Arti TrueNAS App.
+          schema:
+            type: ipaddr
+            required: true
+            # Is this the right way to do OR?
+            show_if:
+              - ["tor_outgoing_enabled", "=", true]
+              - ["clearnet_outgoing_via_tor_enabled", "=", true]
+              - ["tor_incoming_enabled", "=", true]
```
Keeqler, line 119: I want `tor_ip` to show up only if one of these 3 values are `true`. Is this the right way to do it?

stavros-k: Great question, I'll let you know soon

stavros-k:
Is my assumption that rpc uses the default port if `rpcbind` flag is not passed? (ie always binds a port)

Keeqler:
`-mempoolfullrbf` has been removed in [v29](https://bitcoincore.org/en/releases/29.0/) since full RBF is now a standard policy.
It can be removed along with the `full_rbf_enabled` question.

stavros-k (approved):
LGTM - Pending to check the OR on questions before merge

ix-dev/community/bitcoind/questions.yaml
stavros-k: I think this is where it is handled
https://github.com/truenas/middleware/blob/7283c398d3c90baf2665b9884a484bb76856723e/src/middlewared/middlewared/utils/__init__.py#L427

I've updated the questions, we will see when its live to confirm!

Issue #2086, opened 2025-04-15
outline update and configuration


elfensky:
I cannot update outline to 8.3+ 

The App UI limits the following values to 32 characters
- values.outline.secret_key
- values.outline.utils_secret
But it seems outline expects them to be 64. However, I was not able to find anything related to that in the docs, the example docker .env file still mentioned 32 bytes. 

` # a lot of python code prior to this, ending with
[EINVAL] values.outline.secret_key: Should be between 64 and 64
[EINVAL] values.outline.utils_secret: Should be between 64 and 64
`

A similar issues occurs when attempting to set up Github Authentication
- App Private Key field is limited to 1024 characters
however, the base64 encoded key (as mentioned in [outline's documentation](https://docs.getoutline.com/s/hosting/doc/github-GchT3NNxI9) is much longer, almost 2500 characters.

elfensky:
The full trace
` Error: Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/middlewared/job.py", line 509, in run
    await self.future
  File "/usr/lib/python3/dist-packages/middlewared/job.py", line 556, in __run_body
    rv = await self.middleware.run_in_thread(self.method, *args)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/middlewared/main.py", line 1367, in run_in_thread
    return await self.run_in_executor(io_thread_pool_executor, method, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/middlewared/main.py", line 1364, in run_in_executor
    return await loop.run_in_executor(pool, functools.partial(method, *args, **kwargs))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/middlewared/schema/processor.py", line 183, in nf
[...]

PR #2087, opened 2025-04-15
fix outline schema for updatability

elfensky:
This change should allow Outline to be updated again.
Fixes one of the problems mentioned in https://github.com/truenas/apps/issues/2086

change secret_key to 32 length
change utils_secret to 32 length


elfensky:
This is the wrong fix. They should indeed be 64 characters long. It's the initial setup field that's wrong, only allowing 32 characters

PR #2089, opened 2025-04-16
fix outline schema for updatability

elfensky:
change secret_key min_length to 32 length 
change utils_secret min_length to 32 length
so that updating from older versions is possible.
It could be set back to 64 in later releases, a couple updates in, once the majority has updated.

Bumped version to 1.0.4

This fixes Issue1 in https://github.com/truenas/apps/issues/2086 so you can update the app from an older version where those values were 32-characters long.


stavros-k:
Follow #2119

PR #2113, opened 2025-04-21
Warracker: Add uploads question to storage configuration

### LINKED ISSUE #2112: Warracker: Cannot deploy due to missing uploads mount in questions.yaml
The rendered questions form is missing field for "uploads" mount and thus an instance can not be created. Will create a PR in a couple of minutes to fix this issue.

jondycz:
closes #2112


stavros-k (approved):

stavros-k:
auto perms are (for now) only for postgres data

PR #2137, opened 2025-04-24
Add Electrs app

Keeqler:
Electrs is an Electrum Server implementation optimized for personal usage, which users can connect their Bitcoin wallet to.

Logo
electrs-logo


Keeqler:
Electrs will crash in CI since it depends on a Bitcoin node connection. Anything we can do?

ix-dev/community/electrs/templates/macros/entrypoint.sh
```diff
@@ -0,0 +1,13 @@
+{% macro entrypoint(values) -%}
+#!/bin/sh
+
+{% set cfg_file_path = "%s/config.toml" | format(values.consts.config_dir_path) %}
+
+mkdir -p "{{ values.consts.config_dir_path }}"
+touch "{{ cfg_file_path }}"
+
+sed -i "/auth=/d" "{{ cfg_file_path }}"
+echo 'auth="{{ values.electrs.bitcoind_rpc_user }}:{{ values.electrs.bitcoind_rpc_password }}"' >> "{{ cfg }}"
```
Keeqler, line 10: `{{ cfg }}` should rather be `{{ cfg_file_path }}` right?

stavros-k: yep
thanks

stavros-k:
> Electrs will crash in CI since it depends on a Bitcoin node connection. Anything we can do?

Its fine, I won't exclude it from CI, so we can at least inspect logs when updates/changes happen to see if its just the connection error (simillar to apps like tailscale)

stavros-k (approved):
Thanks

Keeqler:
Thanks!

PR #2169, opened 2025-04-29
Allow bind mode for DNS listening

Madko:
Bind mode was missing.
There is some kind of conflict now that there is some dnsmasq using port 53 on libvirt bridges. I guess that's related to incus integration.
Bind mode allows pi-hole to start its dnsmas on a specific interface. pi-hole container can now work on 25.04.


Madko:
Sorry. Should I do a new PR ?

stavros-k (approved):

PR #2197, opened 2025-05-04
Add Convertx

### LINKED ISSUE #1777: App request: ConvertX
It is a self hosted online file converter app, it looks like it has a simple ui and easy to use. You can use it to convert files without using those pdf to word converter type of sites, the app claims to support 1000+ file formats (i havent used it yet but it looks cool)

Image

The github page: [C4illin/ConvertX: 💾 Self-hosted online file converter. Supports 1000+ formats ⚙️](https://github.com/C4illin/ConvertX?tab=readme-ov-file)

The docker page: [c4illin/convertx - Docker Image | Docker Hub](https://hub.docker.com/r/c4illin/convertx)

jondycz:
closes #1777 

logo:
https://filebin.net/3i22rr9ywfdxwcp6


jondycz:
I'm confused. It fails due to:
`No write access to /.config/calibre using a temporary dir instead`

But the only mount point is `/app/data` - perhaps the app needs to run as root? That would be unfortunate.
Docs dont mention this, but perhaps we could add a config mount point for /.config? Though I am unsure whether ALL of the convertor utilities use this path. Probably not.

stavros-k:
> JWT secret isn't actually required. If not supplied, it generates a random UUID

I know, but if you leave it to use a random one, it will invalidate all login sessions on contaienr restart (or at least thats what I would expect)

stavros-k (approved):

jondycz:
And by not passing HTTP_ENABLED, the UI loads over http (app doesn't support https, only through a proxy) but doesn't allow for logging in, due to the variable setting the "secure" flag on the auth cookie

stavros-k:
> And by not passing HTTP_ENABLED, the UI loads over http (app doesn't support https, only through a proxy) but doesn't allow for logging in, due to the variable setting the "secure" flag on the auth cookie

thanks

jondycz:
> > JWT secret isn't actually required. If not supplied, it generates a random UUID
> 
> I know, but if you leave it to use a random one, it will invalidate all login sessions on contaienr restart (or at least thats what I would expect)

That is probably correct. Though if set as optional, users get to choose whether or not they desire the auth to persist on reboots. I'd actually prefer it that way.

jondycz:
All looks good to me now. Thanks @stavros-k! ❤️

PR #2206, opened 2025-05-05
arti: accept list of hidden services to host

Keeqler:


ix-dev/community/arti/questions.yaml
```diff
@@ -42,6 +42,56 @@ questions:
...
+                        type: ipaddr
+                        required: true
+                        show_if: [["is_app", "=", false]]
+                    - variable: port
+                      label: App Port
+                      schema:
+                        type: int
+                        required: true
+                        $ref:
+                          - definitions/port
```
stavros-k, line 86: Is this port going to be bound on the host? Or is it a remote port?

If its remote, you shouldnt use the $ref for port. As it will not allow setting a port that is already bound on the host.

stavros-k (approved):

ix-dev/community/arti/questions.yaml
Keeqler: Yes, remote. Sorry for the delay. Thanks!

PR #2217, opened 2025-05-06
Add Channels DVR

### LINKED ISSUE #615: Community Chart Suggestion: channels-dvr
https://hub.docker.com/r/fancybits/channels-dvr

https://github.com/truecharts/charts/tree/master/charts/stable/channels-dvr

Flexible, standalone DVR server⁠ that lets you schedule and watch recordings anywhere.
https://getchannels.com

jondycz:
Unfortunately channelsDVR doesnt use versions in tags.

logo: https://github.com/homarr-labs/dashboard-icons/blob/main/svg/channels-dvr.svg

closes #615


jondycz:
It seems like the docker image doesn't have a shell. I can see this in the failed run logs: `timeout: can't execute 'bash': No such file or directory`

Is there an alternative way to do healthcheck or am I just blind as always @stavros-k? Thanks for suggestions.

stavros-k:
> It seems like the docker image doesn't have a shell. I can see this in the failed run logs: `timeout: can't execute 'bash': No such file or directory`
> 
> Is there an alternative way to do healthcheck or am I just blind as always @stavros-k? Thanks for suggestions.

`http` check uses bash, but I checked and the container has `curl`, `wget` and `nc` (`netcat` in `set_test()` ) available.

stavros-k (approved):

PR #2248, opened 2025-05-11
Fix Newt user+group

HPPinata:
This is not exposed as a variable in the UI and also doesn't make sense to expose.

Without setting it here creation fails with: jinja2.exceptions.UndefinedError: 'dict object' has no attribute 'run_as'


stavros-k:
Fixed in #2242

PR #2256, opened 2025-05-12
nzbget: run rootless

stavros-k:


stavros-k:
Can't happen currently, while keeping backwards compat.
It only allows running as `1000` user, not any user. 
Doing so here, will break existing installs.

PR #2258, opened 2025-05-12
Add Pterodactyl Panel

jondycz:
Pterodactyl consists of 2 separate parts. Panel and Wings. Usually only one Panel instance runs, which acts as a main control center.

Then there is usually multiple wings, which listen for commands from the panel. This PR should hopefully add a working panel which can then be connected to wings either hosted elsewhere or on TrueNAS using a custom app. It makes sense to make this separate apps as users might want to only host the panel on TrueNAS and then install wings on dedicated servers elsewhere. (similar to newt and pangolin)

Most of the installation should be automatic, but there is a single catch. Upon first install, users have to run `php artisan p:user:make` within the pterodactyl container to create an admin user. It has an interactive step-by-step tutorial on user creation, just clicking enter, keyboard arrows and typing in an admin user name and password.

icon: https://github.com/homarr-labs/dashboard-icons/blob/main/svg/pterodactyl.svg

example docker compose: https://github.com/pterodactyl/panel/blob/1.0-develop/docker-compose.example.yml

I decided to opt out from providing https support in this app by default because it requires manual configuration of nginx and certbot which are running inside of the pterodactyl container. Users can follow the manual steps from official pterodactyl documentation to set this up (would require mounting nginx and certificate folder to the host). It is simpler to just hide pterodactyl behind NginxProxyManager and users still have the choice to set the SSL themselves. If there is some easy way to implement this, feel free to change the PR to support this.

Mail environment variables are also not needed as they can be configured through the webUI of pterodactyl. Or users can define those as "additional environment variables"

I'm sure there will be a lot of issues, this is just a first shot at this.


jondycz:
Honestly, not running this app as root causes so many inter-container read/write issues. Should i try with your temp storage trick @stavros-k? In case you want to change/add direct mount points, doing this would only be a waste of time and also adding more bloat code that will later be removed. And I am also unsure whether im skilled enough to not mess this up. I guess I would first check whether a user has added that path as an additional mount point and if not, mount the temp one?

jondycz:
Oh well, I guess this solves all the issues:

`Error: Can't drop privilege as nonroot user`

Unless there are more tricks I'm not aware of.

jondycz:
The permissions are a mess and should be reworked by someone more competent. Could you pretty please take a look when you have some more time @stavros-k? Except for the CAPS, this runs the exact setup (and also produces exact logs) to my custom "yaml" app.

stavros-k:
> Oh, and also the default template docker-compose from pterodactyl includes this as a startup command of the mariadb image. I'm not sure how important it is but i couldnt find any other dependency using a startup command: `--default-authentication-plugin=mysql_native_password`

I doubt it is needed (anymore), and if it was, it would fail always to auth to the database.

stavros-k:
While I _can_ make it create a user on startup, artisan commands are not idempotent, so it would then error on any subsequent starts (we can silence that error). But we would also have to override the entrypoint and command which is "meh".
And given this is an one time thing, I'll let the user handle it

stavros-k (approved):

PR #2263, opened 2025-05-13
Enforce nextcloud nginx port

AlphaDelta9:
Use nginx port instead host port for https access (host port ignored by nginx config)

### DIFF
--- a/ix-dev/stable/nextcloud/questions.yaml
+++ b/ix-dev/stable/nextcloud/questions.yaml
@@ -113,9 +113,8 @@ questions:
           description: |
             Nextcloud host to create application URLs</br>
             Examples: </br>
-            cloud.domain.com:30001</br>
-            cloud.domain.com (if you use port 443 externally)</br>
-            192.168.1.100:9001 (replace ip and port with your own)</br></br>
+            cloud.domain.com</br>
+            192.168.1.100 (replace with your own ip)</br></br>
             This will be appended to the trusted domains list, but changing that will not remove it from the list.</br>
           schema:
             type: string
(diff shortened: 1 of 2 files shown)


ix-dev/stable/nextcloud/questions.yaml
```diff
@@ -294,24 +293,26 @@ questions:
...
                   required: true
-              - variable: use_different_port
-                label: Use different port for URL rewrites
-                description: |
-                  If enabled, the URL rewrite will use [Access Port] defined below instead of the [Node Port].</br>
-                  Note that Nextcloud will still listen on the [Node Port]. (Default 9001)
-                schema:
-                  type: boolean
-                  default: false
+              # - variable: use_different_port
```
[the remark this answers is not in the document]
stavros-k: I also don't see what is the benefit of doing this?
If someone does not use a different external port, will be forced to enter the same port twice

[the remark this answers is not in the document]
AlphaDelta9: My bad, that field used by migration.
Yes, just realize need to handle the https port properly, not just use the port twice.

My point to make sure the host tooltip doesn't show misleading info, because previously I just confused why my https port notworking. (turns out the port cleaned by the template and use the rewrite port)

ix-dev/stable/nextcloud/templates/docker-compose.yaml
```diff
@@ -117,8 +117,8 @@
   {% do nc_env.x.append(("OVERWRITEPROTOCOL", "https")) %}
   {% do nc_env.x.append(("TRUSTED_PROXIES", ["127.0.0.1", "192.168.0.0/16", "172.16.0.0/12", "10.0.0.0/8"] | join(" "))) %}
   {% do nc_confs.append(("logformat.conf", use_x_real_ip_in_logs(), "/etc/apache2/conf-enabled/logformat.conf", "")) %}
-  {% if values.nextcloud.host and values.network.nginx.use_different_port %}
-    {% set host.x = "%s:%d"|format(values.nextcloud.host, values.network.nginx.external_port) %}
+  {% if values.nextcloud.host %}
+    {% set host.x = "%s:%d"|format(values.nextcloud.host, values.network.nginx.external_port if values.network.nginx.use_different_port else values.network.web_port) %}
```
stavros-k, line 121: sadly this change also requires migration. because users that dont have use_different_port checked and have already set OVERWRITEHOST. will receive an error that this variable is already set and wont be able to upgrade.

I'm still not 100% sure what this change is supposed to do, but so far it only sets the OVERWRITEHOST when cert is defined.

ix-dev/stable/nextcloud/questions.yaml
```diff
@@ -307,7 +306,7 @@ questions:
                 description: The external port for Nginx.
                 schema:
                   type: int
-                  default: 443
+                  default: 30027 # 443 used by host
```
stavros-k, line 309: Why is this default better?

stavros-k: 443 is the **external** port. ie a port in the firewall or some other proxy on the network, not always on the host.

stavros-k (approved):

PR #2268, opened 2025-05-13
nextcloud/stable: fix overwritehost env and add warning if external port + host field has missmatched ports.

AlphaDelta9:
Forget expose external port on portal


stavros-k:
Can you please explain what the result is with your changes?

stavros-k:
WebUI is currently not taking into account host (and in other apps too).
This needs to be fixed more centrally.

See `https-values` ci run, it should now render correctly.

stavros-k:
We can't just change examples and the host field format, as this would be a breaking change for existing installs.
I do however handle it more gracefully and added a warning as well

stavros-k:
btw, you can check out nextcloud in test train. Which hopefully will move to stable or community catalog once there is some feedback that it works good.
https://forums.truenas.com/t/nextcloud-test-train-feedback-request/37296

stavros-k (approved):

Issue #2303, opened 2025-05-16
App request / Add: zipline


tacoroumen:
About: https://github.com/diced/zipline
https://zipline.diced.sh/

Icon: https://github.com/diced/zipline/blob/trunk/public/zipline_small.png

The next generation ShareX / File upload server

(like Imgur but selfhosted)

tacoroumen:
Hey @stavros-k in my pull request should i include the library folder for the app as well or only the test_values and docker-compose from the templates folder?

PR #2312, opened 2025-05-18
Add Zipline 

tacoroumen:
Added zipline app #2303 

Icon still needs to be uploaded i have however set the icon url already to https://media.sys.truenas.net/apps/zipline/icons/icon.png

the Original icon can be found [here](https://github.com/diced/zipline/blob/trunk/public/zipline_small.png) or in the app request ticket #2303

### DIFF
--- a/cspell.config.yaml
+++ b/cspell.config.yaml
@@ -374,5 +374,6 @@ words:
   - zenika
   - zerotier
   - zigbee
+  - zipline
   - zksync
   - zoneedit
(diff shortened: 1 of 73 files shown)


tacoroumen:
Idk why it's failing, locally it runs fine. HealthCheck or Dependency issue?

jondycz:
The container doesn't have bash so you can't use tcp. You can also try using wget, curl or netcat

tacoroumen:
Thanks, wget did the trick.

stavros-k (approved):

PR #2334, opened 2025-05-21
Add LanCache Monolithic

stavros-k:


jondycz:
I think we should call this app just "Lancache" as that's what it's called on other NAS app catalogs.

https://unraid.net/community/apps?q=Lancache#r

jondycz:
There's only monolithic and DNS. I see no issues having LanCache and LanCache DNS but you do you. Perhaps there could also be an option in the questions.yaml to enable the LanCache DNS to have a full stack app?

Qubad786 (approved):

PR #2340, opened 2025-05-22
immich: upgrade to 133

stavros-k:


rabotond:
Hello,

a new input has been added in this version called "DB_STORAGE_TYPE" , shouldnt be it exposed in the template?

TheOnlyArtz:
> DB container variables are not accessible. 
> And I wouldnt recommend running DBs on HDDs.
> 
> 

What does that mean for users who don't have their DB stored on a SSD as users on Reddit reported crashes trying to boot an instance without the correct setting 

[Reference](https://www.reddit.com/r/immich/s/FUn9xOBevl)

Why can't we have access for the database container variables? Is there a limit from the Truenas architecture side of things?

gyscos:
The main difference is that it uses a different postgres config file:

https://github.com/immich-app/base-images/blob/1f8398155909759fdc16b2036848fb548dcf7e31/postgres/postgresql.hdd.conf
or
https://github.com/immich-app/base-images/blob/1f8398155909759fdc16b2036848fb548dcf7e31/postgres/postgresql.ssd.conf

with the difference being this added to the SSD (default) config:
```
effective_io_concurrency = 200
random_page_cost = 1.2
```

[Regarding the first parameter](https://postgresqlco.nf/doc/en/param/effective_io_concurrency/): 
> A value higher than needed to keep the disks busy will only result in extra CPU overhead

[Regarding random_page_cost](https://postgresqlco.nf/doc/en/param/random_page_cost/):
> Reducing this value relative to [seq_page_cost](https://postgresqlco.nf/doc/en/param/seq_page_cost/) will cause the system to prefer index scans; raising it will make index scans look relatively more expensive.

Neither should really cause crashes, but they will certainly affect performance. It would be very nice to be able to set this option.

gyscos:
> DB container variables are not accessible

I'm not sure what you mean by that.

Here, env vars are injected based on the config:
https://github.com/truenas/apps/blob/102d5d97e4733dae4c13e148bfaf1276736717ad/trains/community/immich/1.7.44/templates/library/base_v2_1_16/deps_postgres.py#L55-L63

The config is built here:
https://github.com/truenas/apps/blob/102d5d97e4733dae4c13e148bfaf1276736717ad/trains/community/immich/1.7.44/templates/docker-compose.yaml#L6-L11

Based on some answers to the questions here (example of one such field):
https://github.com/truenas/apps/blob/102d5d97e4733dae4c13e148bfaf1276736717ad/trains/community/immich/1.7.44/questions.yaml#L53

jondycz:
@gyscos There is really no reason to modify the library in order to make app specific exceptions.

You can set the environment variable in the docker-compose template using:

```
{% do postgres.container.environment.add_env("DB_STORAGE_TYPE", immich.database_storage_type) %}
```

Where the database storage type can be either a checkbox or a select box with the value of "HDD" and we can also make an if clause to only set this env if the variable equals HDD

TheOnlyArtz:
@stavros-k Please review #2352 .

We need this update ASAP as the new mobile update for Immich breaks on every version which is below 1.33

jondycz:
> @stavros-k Please review #2352 .
> 
> We need this update ASAP as the new mobile update for Immich breaks on every version which is below 1.33

The mobile app should be backwards compatible. Only issue with version mismatch is when your app is outdated and you update server to 1.133

TheOnlyArtz:
Thank you @stavros-k  🙏🏼

cole21771:
Let's get this moving please 🙏

PR #2352, opened 2025-05-23
Immich 133 exposing storage optimization

TheOnlyArtz:
Exposing the storage optimization to the end user to choose.
This storage option is critical as setting it wrong will break the Immich instance.

- Added a question with an enum (either "SSD" or "HDD")
- Changed the deps_postgres and settings "DB_STORAGE_TYPE" only when it's the immich postgres image.


TheOnlyArtz:
Followed @jondycz suggestion and didn't modify the postgres library

stavros-k:
As I've already said, it won't be exposed at the moment.
If this is allowed, it will just invalidate recommendation for SSD on apps/db data.

I've already started getting DMs in various platforms to customing postgres configuration for HDDs due to the discussion in #2340.

gyscos:
Could using a HDD for the DB be not recommended, but still supported? Sometimes SSDs simply are not available.

TheOnlyArtz:
I don't get it, why can't we support it?
It literally BREAKS for users who use an HDD for their database.

I store the database on a HDD because this is what I have available, you are right, SSDs are RECOMMENDATION, not mandatory.

Give the users freedom to do whatever they want, it's their hardware not yours so you you bother so much?

Not exposing this variable means incompatibility for thousands of users that don't have SSDs available for the database storage... 

@stavros-k

stavros-k:
> It literally BREAKS for users who use an HDD for their database.

That's not true, It just performance optimizations. It's not make or break.

> Not exposing this variable means incompatibility for thousands of users that don't have SSDs available for the database storage...

That's also not true. 

> Give the users freedom to do whatever they want

That's what custom-app gives.

> SSDs are RECOMMENDATION, not mandatory.

Which is why you have the ability to customize what you want via *custom app*.

---

This env var does not make or break the app. Its just performance optimizations, which you didn't had before.
How is it suddenly make or break?

TheOnlyArtz:
> > It literally BREAKS for users who use an HDD for their database.
> 
> That's not true, It just performance optimizations. It's not make or break.
> 
> > Not exposing this variable means incompatibility for thousands of users that don't have SSDs available for the database storage...
> 
> That's also not true. 
> 
> > Give the users freedom to do whatever they want
> 
> That's what custom-app gives.
> 
> > SSDs are RECOMMENDATION, not mandatory.
> 
> Which is why you have the ability to customize what you want via *custom app*.
> 
> ---
> 
> This env var does not make or break the app. Its just performance optimizations, which you didn't had before.
> How is it suddenly make or break?

I've already mentioned users on Reddit claiming this breaks the instance on incorrect setting.

kabatp:
So are we ignoring the manufacturer's suggested way to boost performance for "go buy SSD or create custom app"? In that case, there is no reason to use the catalog, and let's create all custom apps, as we can't be sure what changes will not be implemented in the future.

BTW, I am running the updated version of the app with the old version of the server, and I am not encountering any issues, so I guess the one to two days will be fine for users to wait, if any changes need to be made for this variable.

TheOnlyArtz:
Even if it won't break the instance for HDD users the SSD optimization is aggressive, it sets the `random_page_cost` to `1.2` which is way down from the default value of `4.0`

What it means HDD would be a major misconfiguration likely leading to poor database performance.

It would significantly underestimate the cost of random I/O. This would lead the PostgreSQL query planner to favor index scans and other random access patterns even when a sequential scan would be much more efficient in reality.

gyscos:
> This env var does not make or break the app. Its just performance optimizations, which you didn't had before.

That's not really accurate. Previous images didn't have a custom postgresql.conf and were relying on default values. The SSD config is what deviates from postgresql's default values, so for a HDD user, it brings new "de-optimizations" that were not there before.

PR #2429, opened 2025-05-31
add directus to the catalog of truenas apps

JerryBuilds:
## New Community App: Directus
I've been wanting to use TrueNAS to deploy Directus. I tried my best going through the [contributors guide](https://github.com/truenas/apps/blob/master/CONTRIBUTIONS.md). Here's what works: `./.github/scripts/ci.py --app directus --train community --test-file basic-values.yaml --wait=true`

But I wasn't able to test that my questions.yaml is correct. I couldn't find a way to test this, other than manually examining the YAML by hand. If there is a way, please let me know and I can do the verification. Ideally, if I could get the same screen on TrueNAS Scale when I am about to deploy the new app, that would be nice.

I also wasn't sure about the `icon` field in the `app.yaml`, and how that gets populated. And, I'm also not certain how to load images.

Additionally, do I have to commit the copied `templates/libary` files?

### DIFF
--- a/ix-dev/community/directus/README.md
+++ b/ix-dev/community/directus/README.md
@@ -1,0 +1,3 @@
+# Plex Auto Languages
+
+[Directus](https://github.com/directus/directus) is a real-time API and App dashboard for managing SQL database content.
--- a/ix-dev/community/directus/app.yaml
+++ b/ix-dev/community/directus/app.yaml
@@ -1,0 +1,32 @@
+app_version: 11.8.0
+capabilities: []
+categories:
+- media
+date_added: '2025-05-30'
+description: Directus is a real-time API and App dashboard for managing SQL database content.
+home: https://directus.io/
+host_mounts: []
+icon: https://media.sys.truenas.net/apps/directus/icons/icon.svg # change?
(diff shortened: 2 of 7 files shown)


jondycz:
> Another question -- if and once this gets merged into master, when will the update show up for users to deploy "Directus" on their TrueNas-Scale boxes?

When sonicaj updates the catalog changes. Usually happens within a couple of minutes or hours of your PR getting accepted.

stavros-k (approved):

stavros-k:
> I also wasn't sure about the `icon` field in the `app.yaml`, and how that gets populated. And, I'm also not certain how to load images.

I have to upload it. (Done)

> Additionally, do I have to commit the copied `templates/libary` files?

Yes

> But I wasn't able to test that my questions.yaml is correct. I couldn't find a way to test this, other than manually examining the YAML by hand. If there is a way, please let me know and I can do the verification. Ideally, if I could get the same screen on TrueNAS Scale when I am about to deploy the new app, that would be nice.

There isn't a way, copying an existing and modifying is enough in most cases.

Issue #2433, opened 2025-05-31
Permission Error in Gramps-Web


HubGon:
Hi,
as i am trying to Add the Environment Variable "GRAMPSWEB_VECTOR_EMBEDDING_MODEL" to Gramps-Web, i am getting a permission error for the internal path "/.cache". I got to solve it by mapping this container path to a host path and chown it to the "apps" user. Maybe the standard installation of this app can be modified to map this by default, so noone else get this problem?

Kind regards

jondycz:
Hey, I'll make a PR later today if Stavros doesn't beat me to it. 😅

I'm kinda new to the Gramps thing and you seem to be more experienced. Is there any community I can join of like-minded people or are you all on your own on this?

HubGon:
Hey,
in the standard installation of gramps web this step is not necessary, I think this problem is more related specifically to the installation through truenas apps. As I understand the docker compose template, the container user is set to 568, so the internal permission doesn't fit anymore. This is more a suggestion to fix it.

As for the question of a community, I don't know of one. I am just tinkering around. This error and solution is more the reading of the documentation and the container logs :)

Kind regards

jondycz:
Yes, your assumption is correct. We will need to create a temporary mount to /.cache so that we can chown the directory.

HubGon:
> Yes, your assumption is correct. We will need to create a temporary mount to /.cache so that we can chown the directory.

Maybe a permanent (hidden?) volume automatically mountet to the path instead of a temporary one? The /.cache folder is the place where the embedding model is placed. If it is only temporary the model would be downloaded every time when the app is started instead of just loading the last time downloaded one.

PR #2438, opened 2025-06-01
Gramps: Mount /.cache as a temp volume to allow usage of ML

jondycz:
Gramps web allows running Machine Learning models within the container. It stores the models in a ./cache folder, which is not accessible when not run as a root. This PR mounts the ./cache folder as a temp volume and lets our permissions container take care of fixing the permissions of the directory.

Things to consider:
What is the lifetime of the temporary volume? Is it cleared on every reboot or duing container updates? If it gets cleared on every container reboot, we might want a different approach for this solution as models can be sometimes quire large. Clearing it on container updates seems right to me as a new update might include updated models and might not dispose of the old ones properly resulting in larger space being used.

closes: #2433

### DIFF
--- a/ix-dev/community/gramps-web/app.yaml
+++ b/ix-dev/community/gramps-web/app.yaml
@@ -36,4 +36,4 @@ sources:
 - https://github.com/gramps-project/gramps-web
 title: Gramps Web
 train: community
-version: 1.0.2
+version: 1.0.3
--- a/ix-dev/community/gramps-web/app.yaml
+++ b/ix-dev/community/gramps-web/app.yaml
@@ -36,4 +36,4 @@ sources:
 - https://github.com/gramps-project/gramps-web
 title: Gramps Web
 train: community
-version: 1.0.3
+version: 1.0.4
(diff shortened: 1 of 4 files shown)


stavros-k:
If it stores large models there, we should expose it to the UI and default to temporary (i think jellyfin has this for logs or plex?).

Temporary vols are cleaned up on every app restart/update/edit etc

jondycz:
> If it stores large models there, we should expose it to the UI and default to temporary (i think jellyfin has this for logs or plex?).
> 
> Temporary vols are cleaned up on every app restart/update/edit etc

Alright. That seems like a better solution. It's already past bedtime and I got to go to work tomorrow, you can either wait for me to re-do it tomorrow or you can close this and do it your suggested way.

stavros-k:
No worries I'm off as well currently 😅

HubGon:
Hey guys,
thanks for your effort. The mentioned directory is also used for loading the model for the next startup. So if the directory is cleaned up every time, the model would be newly downloaded for every start of the app. Maybe a permanent mount would be more sufficient?

stavros-k (approved):

PR #2504, opened 2025-06-05
remove refs and add range

stavros-k:
Depends on #2501


amomchilov:
Hey Stavros,

Renaming the folder to change the version name causes a ton of church, which makes it pretty hard to navigate around this repo's history.

Would it be possible to store the version elsewhere, so the entire library doesn't have to be created on version bump?

Alternatively, if the version number needs to be part of the path for whatever reason, could that be done with a symlink instead? That way each version bump just requires changing the symlink destination, without renaming/moving any files.

stavros-k:
> Hey Stavros,
> 
> Renaming the folder to change the version name causes a ton of church, which makes it pretty hard to navigate around this repo's history.
> 
> Would it be possible to store the version elsewhere, so the entire library doesn't have to be created on version bump?
> 
> Alternatively, if the version number needs to be part of the path for whatever reason, could that be done with a symlink instead? That way each version bump just requires changing the symlink destination, without renaming/moving any files.

You can check history on the `ix-dev` directory, which is where any changes happen.
Trains dir is autogenerated **completely**

amomchilov:
Ahhhh, I see, thanks!

PR #2517, opened 2025-06-08
Add GPU settings to Romm

kabatp:
Add GPU resources to Romm


stavros-k (changes requested):
Please dont edit files under `trains` directory. only under `ix-dev`.
Thanks

kabatp:
Hello @stavros-k , do I need to edit anything other than questions.yaml to add GPU resources to the app? It looks to me that resources are added to the container automatically based on the questions, and I can't find anything else where to add them (tried to reverse engineer the Immich app)

stavros-k:
Can you share some docs that mention Romm using GPU? I cant find anything on their repo

kabatp:
As per this issue
https://github.com/rommapp/romm/issues/1499

and also 

https://docs.romm.app/3.7.3/Platforms-and-Players/EmulatorJS-Player/

RomM is using EmulatorJS for emulation which is using PPSSPP for PSP games. Which is using GPU resources.

https://forums.ppsspp.org/showthread.php?tid=12713

https://github.com/hrydgard/ppsspp

But I would agree that this could be consulted with RomM itself. Should I open issue in their repo or is there anything like "app ownership" in TrueNas Community catalogue?

stavros-k:
I'm just looking if we need to configure anything (assuming we can) for Romm, or it will pick up any GPUs automatically.
Is there any "success" reports for Romm image itself? maybe it lacks needed dependencies which will render the gpu's addition useless.

kabatp:
I tried to mess with the RomM as a custom app with passing non-NVIDIA GPUs with no success, then I had a chat with the RomM admin/support guy to find out I was trying to fix an issue that is on the other side of the RomM app.

The PPSSPP emulations require too much power that the browser is not able to offer, which means that regardless of hardware or GPU support, there is kinda no way to run the PPSSPP emulations via RomM itself with reasonable performance.

PR #2570, opened 2025-06-14
Add Many Notes Application

YTKme:
This is my first time trying to add an application, please let me know if I missed anything 🙏🏻

Add [Many Notes](https://github.com/brufdev/many-notes) Application

The icon logo and screenshot can be found in application GitHub
* [Logo](https://github.com/brufdev/many-notes/blob/main/art/logo.png)
* Screenshot
  * [Light](https://raw.githubusercontent.com/brufdev/many-notes/refs/heads/main/art/theme-light.png)
  * [Dark](https://raw.githubusercontent.com/brufdev/many-notes/refs/heads/main/art/theme-dark.png)


stavros-k:
Sorry but this is 100% untested. I does not even render successfully.

Also upstream docs suggest that a docker image have to be rebuilt in order to use bind mounts. (Which is the only mounts we support). If that's the case, then I don't think this will be accepted (due to requiring image rebuild)

YTKme:
> Sorry but this is 100% untested. I does not even render successfully.
> 
> Also upstream docs suggest that a docker image have to be rebuilt in order to use bind mounts. (Which is the only mounts we support). If that's the case, then I don't think this will be accepted (due to requiring image rebuild)

Thanks @stavros-k, I'll check on this and update

PR #2583, opened 2025-06-16
portainer: fix http port

stavros-k:


Qubad786 (approved):

185245725:
thanks, http published/exposed works now but webui button is missing if https not published/exposed

PR #2610, opened 2025-06-19
immich: update vectorchord

jondycz:
According to immich release notes, it is recommended to update the vectorchord version. (not required though)

https://github.com/immich-app/immich/releases/tag/v1.135.0

### DIFF
--- a/ix-dev/community/immich/app.yaml
+++ b/ix-dev/community/immich/app.yaml
@@ -43,4 +43,4 @@ sources:
 - https://github.com/immich-app/immich
 title: Immich
 train: community
-version: 1.9.3
+version: 1.9.4
--- a/ix-dev/community/immich/ix_values.yaml
+++ b/ix-dev/community/immich/ix_values.yaml
@@ -16,7 +16,7 @@ images:
     tag: v1.135.0-openvino
   pgvecto_image:
     repository: ghcr.io/immich-app/postgres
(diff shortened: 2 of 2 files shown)


jondycz:
blocked by upstream: https://github.com/immich-app/immich/issues/19313

> Right after we rolled out the VectorChord vector database extension, the library dropped a new version [0.4.2](https://github.com/tensorchord/VectorChord/releases/tag/0.4.0) with some significant improvements (classic timing!). We've updated our server to support it and made it the default in the docker-compose.yml file.

Would have thought that the following sentence means it was tested... Anyways issue has to do with their migration scripts.

jondycz:
Update:
> Sorry for the inconvenience. Setting parameters normally doesn't require superuser permission, so we didn't realize this migration would fail for non-superuser environments.

This issue only affects users who connect to the database as a non-superuser. Immich team is investigating ways to allow this even for non-superusers.

stavros-k:
There is no need to do that manually. I explicitly revert this update every day on the bot, until its stable.
Bot will handle it.
Thanks

jondycz:
@stavros-k so apparently this should be fixed in 1.135.2 but a new issue appeared 

https://github.com/immich-app/immich/issues/19374

stavros-k:
Shouldnt affect us, we use `immich` for user/db name statically

Issue #157, opened 2023-08-12
Community Chart Suggestion: Minetest


pabloeisenhut:
https://www.minetest.net/

jondycz:
Project has been renamed to "Luanti"

stavros-k:
If anyone has a working setup, please share. Even the "try out" docker they have on their github docs does not start for me.

jondycz:
```
2025-06-20` 12:44:37: ERROR[Main]: To run a Luanti server, you need to select a game using the '--gameid' argument.
2025-06-20 12:44:37: ERROR[Main]: Check out https://content.luanti.org for a selection of games to pick from and `download.
```
So, in order to run the server, we need to specify which "world" or "game" we want to run. LinuxServer.io does this by compiling the games thelselves with cmake, copying them to the games folder and then providing the --gameid argument.


https://github.com/linuxserver/docker-luanti/blob/73b9a3488d11dbc1ab20059c5db380ef8e1d2479/Dockerfile#L113

stavros-k:
And how I can find what the gameid is?
In the contentdb I cant find which is the gameid

jondycz:
> And how I can find what the gameid is? In the contentdb I cant find which is the gameid

The ID is the name of the game folder which you need to manually download and paste into your games folder.

stavros-k:
Okay, I'll see if adding an init container to fetch the game/world works.

jondycz:
They do have some API, so might be pretty nice UX to have an input field in the UI to select the map. Format could be `author/mapname` just copied from the URL.

You could then check the games folder if the `mapname` is there. If not, download it from contentdb

https://content.luanti.org/help/api/#package-queries

jondycz:
Actually this is way easier:

```
https://content.luanti.org/packages/${author}/${name}/download/`
https://content.luanti.org/packages/${author}/${name}/releases/${release}/download/`
```

PR #2614, opened 2025-06-20
add luanti

stavros-k:
Closes https://github.com/truenas/apps/issues/157

Currently not working, if anyone has any ideas what's the bare minimum to start it. please share


jondycz:
You're a crazy man. Truly, now the easiest way to set up luanti is through TrueNAS. 🤣

PR #2615, opened 2025-06-20
Immich: Update vectorchor to 0.4.2

gyscos:
As mentioned in https://github.com/immich-app/immich/releases/tag/v1.135.0, immich still supports vectorchord 0.3.0 but now defaults to 0.4.2.

Note that 0.4.3 is already available, and immich has started building images with it as well, but it hasn't been made the default version yet.


stavros-k:
See #2610

gyscos:
Ah just noticed it was part of https://github.com/truenas/apps/pull/2613 that got canned, so I guess it'll just come with the next update?...

stavros-k:
Once upstream fix lands. yes

PR #2644, opened 2025-06-25
Add Factorio

cemno:
I wanted to add Factorio server, but i still require some help setting everything up. Currently, when starting the test docker image, i get the following error:
```
Failed to print docker compose config
unknown shorthand flag: 'p' in -p
See 'docker --help'.
```
And here is the Factorio icon:
thumbnail

### DIFF
--- a/ix-dev/community/factorio/questions.yaml
+++ b/ix-dev/community/factorio/questions.yaml
@@ -1,16 +1,16 @@
 groups:
   - name: Factorio Configuration
-    description: Configure the Factorio
+    description: Configure Factorio
   - name: User and Group Configuration
-    description: Configure User and Group for the Factorio
+    description: Configure User and Group for Factorio
   - name: Network Configuration
-    description: Configure Network for the Factorio
+    description: Configure Network for Factorio
   - name: Storage Configuration
-    description: Configure Storage for the Factorio
+    description: Configure Storage for Factorio
   - name: Labels Configuration
-    description: Configure Labels for the Factorio
+    description: Configure Labels for Factorio
   - name: Resources Configuration
-    description: Configure Resources for the Factorio
+    description: Configure Resources for Factorio
 
 questions:
   - variable: factorio
@@ -85,7 +85,7 @@ questions:
       attrs:
         - variable: user
           label: User ID
-          description: The user id that the Factorio files will be owned by.
+          description: The user id that Factorio files will be owned by.
           schema:
             type: int
             min: 568
@@ -93,7 +93,7 @@ questions:
             required: true
         - variable: group
           label: Group ID
-          description: The group id that the Factorio files will be owned by.
+          description: The group id that Factorio files will be owned by.
           schema:
             type: int
             min: 568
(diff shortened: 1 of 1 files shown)


stavros-k:
> Failed to print docker compose config
> unknown shorthand flag: 'p' in -p
> See 'docker --help'.

You probably have old docker version on your system.

ix-dev/community/factorio/templates/test_values/basic-values.yaml
```diff
@@ -0,0 +1,33 @@
...
+network:
+  host_network: false
+  game_port:
+    bind_mode: published
+    port_number: 8080
+  messaging_port:
+    bind_mode: published
+    port_number: 8081
+
+run_as:
```
stavros-k, line 25: why 845? default is everywhere 568.

ix-dev/community/factorio/app.yaml
```diff
@@ -0,0 +1,43 @@
...
+- email: dev@ixsystems.com
+  name: truenas
+  url: https://www.truenas.com/
+name: factorio
+run_as_context:
+- description: factorio Server runs as root user.
+  gid: 845
+  group_name: factorio
+  uid: 845
+  user_name: factorio
```
stavros-k: I don't think thats true, you haven't done `c1.set_user()` and docker by default runs as root.

ix-dev/community/factorio/ix_values.yaml
```diff
@@ -0,0 +1,7 @@
+images:
+  image:
+    repository: factoriotools/factorio
+    tag: stable
```
stavros-k: please use versioned tag

ix-dev/community/factorio/questions.yaml
```diff
@@ -0,0 +1,359 @@
...
+            default: false
+  - variable: storage
+    label: ""
+    group: Storage Configuration
+    schema:
+      type: dict
+      attrs:
+        - variable: volume
+          label: Factorio Server config, mods and saves Storage
+          description: To keep things simple, the container uses a single volume mounted at /factorio. This volume stores configuration, mods, and saves. To change these settings or addings mods or savefiles, access the dataset manually and change the files to your liking.</br>
```
stavros-k: I dont think we need to have all this here. You can just have a link to upstream docs.

ix-dev/community/factorio/templates/test_values/basic-values.yaml
```diff
@@ -0,0 +1,33 @@
+resources:
+  limits:
+    cpus: 2.0
+    memory: 4096
+
+factorio:
+  
+factorio_custom_env:
+  additional_envs: []
+
```
stavros-k: why have a custom_env section? ALL other apps have the additional_envs in the main section.

ix-dev/community/factorio/questions.yaml
```diff
@@ -0,0 +1,359 @@
...
+    description: Configure Network for the Factorio server
+  - name: Storage Configuration
+    description: Configure Storage for the Factorio server
+  - name: Labels Configuration
+    description: Configure Labels for the Factorio server
+  - name: Resources Configuration
+    description: Configure Resources for the Factorio server
+
+questions:
+  - variable: factorio
```
stavros-k, line 16: anything in this object must also exist in the test values.

cemno:
Thanks a lot for the pointers! 
I was indeed running docker v27. I will rework the app later, it does seem like it will take me some time.

cemno:
The problem regarding the docker error was that I needed to install docker-compose-v2 on Linux. With v2 the command `docker compose` instead of `docker-compose` can be used.

I tried adapting the code to user 568 and passing it as `c1.set_user()`, but I get permission errors on my system. I tried setting the test ix-volume `/opt/tests` to user and group 568, but that didn't work. Do I have to "prepare" the folder or my system in any way if I don't want to run the image as root?

And I adapted the other changes, but I was wondering how environment values, that are added twice, are handled (host ip to bind and the ports are is entered in the network section, but could also be added manually)?

stavros-k:
> but I get permission errors on my system.

That shouldnt matter. The issue is that not all containers are created equally. Some can run as any user, some only as root, some only as a specific non-root user.

stavros-k:
> And I adapted the other changes, but I was wondering how environment values, that are added twice, are handled

You cant add twice any variable

stavros-k:
> host ip to bind and the ports are is entered in the network section

You had changed the structure, which would had broken the ip/port binding feature. fix it.
BIND var is only relevant if user selects host network.

stavros-k (approved):

cemno:
I tested the app real quick and it seems to be working, but one problem is the version not being the stable version. Currently, the latest server version (2.0.58) is being used, but that is ahead of the stable (game and server) version (2.0.55). To connect without changing the game version to "Experimental", the stable server version is needed. This is also the reason i previously entered the [stable tag](https://github.com/factoriotools/factorio-docker). Is there a reason a version tag is required? Maybe the stable version tag `stable-2.0.55` could be used instead? Only versions that are stable get this tag, but they also show the exact version number of the game.

stavros-k:
stable-X.Y.Z can be used yes (renovate-config.js needs to be updated tho)
but plain `stable` not accepted.

cemno:
Something like this?

```
    customVersioning(
      // stable-2.0.55
      "^stable-(?<major>\\d+)\\.(?<minor>\\d+)\\.(?<patch>\\d+)$",
      ["factoriotools/factorio"]
    ),
```

And would the versioning in the app.yaml has to changed manually as well? If needed i can do some changes and open a new pull request.

stavros-k:
> Something like this?
> 
> ```
>     customVersioning(
>       // stable-2.0.55
>       "^stable-(?<major>\\d+)\\.(?<minor>\\d+)\\.(?<patch>\\d+)$",
>       ["factoriotools/factorio"]
>     ),
> ```

Yes

> And would the versioning in the app.yaml has to changed manually as well? If needed i can do some changes and open a new pull request.

First time yes, both app and ix_values.yaml

PR #2685, opened 2025-07-03
Add truenas-zabbix-agent, a mix of zabbix-agent with some midclt sprinkled in

lingfish:
Just opening the PR for now, work not complete yet.

### DIFF
--- a/ix-dev/community/truenas-zabbix-agent/README.md
+++ b/ix-dev/community/truenas-zabbix-agent/README.md
@@ -1,0 +1,3 @@
+# Monitee-agent
+
+[Monitee-agent](https://github.com/Krillsson/monitee-agent) is a cross-platform system monitoring tool.
--- a/ix-dev/community/truenas-zabbix-agent/README.md
+++ b/ix-dev/community/truenas-zabbix-agent/README.md
@@ -1,3 +1,4 @@
-# Monitee-agent
+# truenas-zabbix-agent
 
-[Monitee-agent](https://github.com/Krillsson/monitee-agent) is a cross-platform system monitoring tool.
+[truenas-zabbix-agent](https://github.com/lingfish/truenas-app-zabbix-agent) combines the vendor Zabbix Agent with the
+TrueNAS API `midclt` to provide a simple way to deploy and manage a Zabbix Agent using Docker and Docker Compose.
--- a/ix-dev/community/truenas-zabbix-agent/app.yaml
+++ b/ix-dev/community/truenas-zabbix-agent/app.yaml
@@ -1,3 +1,5 @@
+annotations:
+  min_scale_version: 25.04
 app_version: latest
 capabilities: []
 categories:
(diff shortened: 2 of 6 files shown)


lingfish:
Hi @stavros-k -- how do I go about doing the below, whereby the zabbix config is static and supplied by the app?

```
    volumes:
      - ./conf:/etc/zabbix/zabbix_agentd.d
```

EDIT: meanwhile, I've just included the static file in my Docker image build.

stavros-k:
Thank you for your contribution and interest in adding this integration to our catalog. After review, we're unable to approve this submission at this time due to several security and quality considerations:
Primary concerns:

- This appears to be an unofficial third-party Zabbix integration rather than an official one
- The source repository lacks sufficient development history, documentation, and community validation to establish trustworthiness
- The container permissions requested present potential security risks that we cannot adequately assess given the limited repository credibility

stavros-k:
> how do I go about doing the below, whereby the zabbix config is static and supplied by the app?

For future reference you can see how some other apps mount configs, you can search `c1.configs.add(...)` and see examples

lingfish:
> Thank you for your contribution and interest in adding this integration to our catalog. After review, we're unable to approve this submission at this time due to several security and quality considerations: Primary concerns:

Hrm, ok. I had hoped for a bit more discussion before just being closed... a bit frustrating.

> * This appears to be an unofficial third-party Zabbix integration rather than an official one

Correct, because due to the decision to go to a WS based API, there is no easy way to use your API anymore; Zabbix server doesn't do websockets, so the only option was to use the API via `midclt`, called via `zabbix-agent`. The only way to do it, therefore, is using the `zabbix-agent` `UserParameter` method.

> * The source repository lacks sufficient development history, documentation, and community validation to establish trustworthiness

Correct, because it isn't particularly complicated. Yes, it's a custom image, but the associated docker/compose files are clear, simple, and public.  It uses the official Zabbix image, and simply installs `midclt` via `pip`.

> * The container permissions requested present potential security risks that we cannot adequately assess given the limited repository credibility

This is quite possibly a error on my part -- it needs no special or particular permissions, other than listening to a TCP socket.

Re the configs -- the situation is that the user isn't required to configure the app... it's a simple, single `UserParameter` line in a `zabbix-agent` config file that is static. Would it not be weirder/clunky to expect the user to setup a config volume, go to another repo and copy a file in?!?

stavros-k:
> * The container permissions requested present potential security risks that we cannot adequately assess given the limited repository credibility

Let me rephrase that. A container having access to the TN API Key, gets full access to the system. Not just metrics.
Coupled with the fact that neither the source repo or your account, don't have established trustworthiness, makes it very hard to accept it in the catalog. 

To be clear, I'm not saying that your intents are bad. But you can see why its hard to accept it.

It would be much better to create and share a guide on how to deploy zabbix agent on the forums (as a custom app)
And allow users to bundle in whatever dependencies needed. (You can use inline Dockerfile and configs in compose nowdays, you can render nextcloud/stable app to see the compose file structure for that).
That way they know exactly what's in the container that's built.

I'm sure that if anyone is looking to add midclt support on their zabbix agent, they already have one running, and adding another agent just for midclt, does not seem the right way to do it.
They would probably need 1 agent to collect everything.

PR #2732, opened 2025-07-07
nextcloud: fix IPv6 host truncation in trusted_domains

Xseventh:
When a user sets **Host** to `[xxxx:xxxx:xxxx:xxxx::xxxx]:yyyy`, only  
`[xxxx` was written to `config.php → trusted_domains`.  
This patch preserves the full address `xxxx:xxxx:xxxx:xxxx::xxxx`, so IPv6
access works out of the box.

I ran the following command and confirmed that
`NEXTCLOUD_TRUSTED_DOMAINS` contains the correct IPv6 value:

```bash
.github/scripts/ci.py --app nextcloud --train stable --test-file ipv6-values.yaml

### CODE REVIEW

**Comment by stavros-k** [comment]
It should now work much better! Waiting tests to pass first

**Review by stavros-k — APPROVED** [review]

**Comment by Xseventh** [comment]
Hi @stavros-k , sorry 🙇 for the oversight in the previous patch.

- Issue The template now writes the IPv6 literal without the surrounding brackets, e.g.
```
'trusted_domains' => [
  '127.0.0.1',
  'localhost',
  'nextcloud',
  'aaaa:bbbb:cccc:dddd',   // ← missing [ … ]
],
```

- Expected Nextcloud expects the bracketed form to match the browser’s Host header:
```
'trusted_domains' => [
  '127.0.0.1',
  'localhost',
  'nextcloud',
  '[aaaa:bbbb:cccc:dddd]',
],
```
I tried to adjust the Jinja logic but, after the large refactor that landed in 2.0.6, I’m not 100 % sure where the cleanest fix should live. My earlier idea was simply:
```
{% if host.x.startswith("[") and "]" in host.x %}
    {% do trusted_domains.x.append(host.x.split("]")[0] ~ "]") %}
{% else %}
    {% do trusted_domains.x.append(host.x.split(":")[0]) %}
{% endif %}
```
If someone more familiar with the new template structure could tweak this (or point me to the right helper macro), that would be fantastic. I’m happy to test whatever patch you propose.

Again, apologies for the extra churn, and thanks for all the reviews & merges!

PR #2748, opened 2025-07-09
bumping npm version due to chown permissions issues

rossampro:
Creating this PR to fix issue noted here https://github.com/NginxProxyManager/nginx-proxy-manager/issues/4629.

### DIFF
--- a/ix-dev/community/nginx-proxy-manager/ix_values.yaml
+++ b/ix-dev/community/nginx-proxy-manager/ix_values.yaml
@@ -1,7 +1,7 @@
 images:
   image:
     repository: jc21/nginx-proxy-manager
-    tag: 2.12.4
+    tag: 2.14.4
 consts:
   npm_container_name: npm
   data_path: /data
--- a/ix-dev/community/nginx-proxy-manager/app.yaml
+++ b/ix-dev/community/nginx-proxy-manager/app.yaml
@@ -1,6 +1,6 @@
 annotations:
   min_scale_version: 24.10.2.2
-app_version: 2.12.4
+app_version: 2.14.5
 capabilities:
 - description: Nginx Proxy Manager is able to change file ownership arbitrarily
   name: CHOWN
--- a/trains/community/nginx-proxy-manager/1.2.5/app.yaml
+++ b/trains/community/nginx-proxy-manager/1.2.5/app.yaml
@@ -1,6 +1,6 @@
 annotations:
   min_scale_version: 24.10.2.2
-app_version: 2.12.4
+app_version: 2.14.5
 capabilities:
 - description: Nginx Proxy Manager is able to change file ownership arbitrarily
   name: CHOWN
--- a/trains/community/nginx-proxy-manager/1.2.5/ix_values.yaml
+++ b/trains/community/nginx-proxy-manager/1.2.5/ix_values.yaml
@@ -1,7 +1,7 @@
 images:
   image:
     repository: jc21/nginx-proxy-manager
-    tag: 2.12.4
+    tag: 2.14.5
 consts:
   npm_container_name: npm
   data_path: /data


ix-dev/community/nginx-proxy-manager/ix_values.yaml
```diff
@@ -1,7 +1,7 @@
 images:
   image:
     repository: jc21/nginx-proxy-manager
-    tag: 2.12.4
+    tag: 2.14.4
```
CrazyWolf13, line 4: shouldn't this be 2.14.5 too?

CrazyWolf13 (changes requested):

ix-dev/community/nginx-proxy-manager/ix_values.yaml
jondycz: 2.14.5 doesn't exist. Newest tag is 2.12.5

Also, this is handled automatically with workflows, no need to manually bump the version.

stavros-k: basides that, there is no image published yet

jondycz: @stavros-k also make sure to read the changelog. Looks like using the s6 environment variable is no longer recommended unless you want bad things to happen. And it seems like a lot of people do use that variable https://github.com/NginxProxyManager/nginx-proxy-manager/releases/tag/v2.12.5

https://github.com/NginxProxyManager/nginx-proxy-manager/issues/4629#issuecomment-3050966633

stavros-k: We don't set that variable, but I can't really "forbid" it, as you can do anything you want with that var. And people might use it for other things other than disabling ownership.

jondycz: Oh well. Anyways, seems like the maintainer of NPM is struggling getting it right, looking for people to test and figure it out. Perhaps if you got some free time, you can throw them a stick or something (Convo in the second link)

rossampro: > 2.14.5 doesn't exist. Newest tag is 2.12.5
> 
> 
> 
> Also, this is handled automatically with workflows, no need to manually bump the version.

Thanks for sharing this information. If I'm understanding correctly, the newest version will be available when a new image is published to the docker registry?

rossampro: @jondycz the newest image was published to the docker registry. when does the workflow run on TrueNAS to pull it?

jondycz: @stavros-k so, does adding this environment variable by default make sense? Apparently without it the deployment takes as much as half an hour. I'm not sure if this env would have some kind of side effect or something. https://github.com/NginxProxyManager/nginx-proxy-manager/issues/4629#issuecomment-3056994374

I see NginxProxyManager doesn't utilize our permissions container so it might be necessary to NOT skip permission mechanisms of NPM?

stavros-k: Permissions container can only operate on external storage. 
Npm is chowning files in the container layer. (Which is why it takes so long due to the immutability)

jondycz: > Permissions container can only operate on external storage. Npm is chowning files in the container layer. (Which is why it takes so long due to the immutability)

So the most logical thing to do is setting the GID and UID to root and disabling chowning files?

stavros-k: or upstream could copy the files in a location that is bind mount and chown them there. 
this can be done on each startup by cleaning the target location and re-copying (so files are always fresh from the image)

Tho programs that use those files must also instructed to use the new location.

Doing file changes on a bind mount does not take that long, as it does not have to create new layer for each file.

PR #2772, opened 2025-07-10
Add portacker

Mostafa-Wahied:


stavros-k:
Hello, thanks for the PR. 

1. Is there a reason that you touched hashes of the library?
2. Is there a reason that questions.yaml has completely different structure and changed defaults than all the other apps?
3. Regarding middleware socket the same idea from here (https://github.com/truenas/apps/pull/2685) applies.

Thanks

mostafa-wahied:
Thank you for your response.

> 1. Is there a reason that you touched hashes of the library?

These hashes got updated after updating the used library to `2.1.42`. Let me know if i should take them off the changes.

> 2. Is there a reason that questions.yaml has completely different structure and changed defaults than all the other apps?

I can address that and make it follow existing patterns

> 3. Regarding middleware socket the same idea from here ([Add truenas-zabbix-agent, a mix of zabbix-agent with some midclt sprinkled in #2685](https://github.com/truenas/apps/pull/2685)) applies.

Could you please clarify a bit more on what does that imply; does that mean that in order for this app to be approved, the middleware mount needs to be not included?
How about the TrueNAS API key support?

I appreciate your response and clarification.

stavros-k:
> These hashes got updated after updating the used library to `2.1.42`. Let me know if i should take them off the changes.

Not sure why would the hash change, there was no changes in either 2.1.42 or 0.0.1 libs.

stavros-k:
> Could you please clarify a bit more on what does that imply; does that mean that in order for this app to be approved, the middleware mount needs to be not included?
> How about the TrueNAS API key support?

Both methods give full control of the system, and without any trustworthiness I can't accept that currently.

mostafa-wahied:
I can work on removing the middleware mount and making the API key off by default unless the user adds it as an env variable. Would that work?

mostafa-wahied:
Closed to start a fresh PR addressing the points made.

PR #2789, opened 2025-07-13
Add MakeMKV

### LINKED ISSUE #220: APP Request MakeMKV
i currently use this and works fine https://github.com/jlesage/docker-makemkv

Momi-V:
Closes #220.

Initial draft, still needs lots of refinement.
PR to avoid uncoordinated duplicate work.

`/dev` read access since MakeMKV needs the raw `/dev/sg*` device with unpredictable numbering in addition to the normal `/dev/sr*` by default. If I find an alternate option that'll be implemented.

### DIFF
--- a/ix-dev/community/makemkv/app.yaml
+++ b/ix-dev/community/makemkv/app.yaml
@@ -1,5 +1,19 @@
 app_version: v25.07.1
-capabilities: []
+capabilities:
+- description: MakeMKV is able to change file ownership arbitrarily
+  name: CHOWN
+- description: MakeMKV is able to bypass file permission checks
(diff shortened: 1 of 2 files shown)


stavros-k:
> Closes #220.
> 
> Initial draft, still needs lots of refinement. PR to avoid uncoordinated duplicate work.
> 
> `/dev` read access since MakeMKV needs the raw `/dev/sg*` device with unpredictable numbering in addition to the normal `/dev/sr*` by default. If I find an alternate option that'll be implemented.

Users can map each device (simillar to esphome and automatic ripping machine).
It should work fine.

PR #2820, opened 2025-07-18
emby: use stable versions

### LINKED ISSUE #2794: Emby use beta version
TrueNas uses the beta version of emby

Releases: https://github.com/MediaBrowser/Emby.Releases/releases

Latest stable version: 4.8.11.0
TrueNas use version: v4.9.1.5

jcassette:
Fixes #2794

### DIFF
--- a/.github/renovate-config.js
+++ b/.github/renovate-config.js
@@ -147,8 +147,8 @@ module.exports = {
       ["storjlabs/storagenode"]
     ),
     customVersioning(
-      // 1.2.3.4, but not 1.2.0.4 (3rd digit 0 equals beta)
-      "^(?<major>\\d+)\\.(?<minor>\\d+)\\.(?<patch>[1-9]\\d*)\\.(?<build>\\d+)$",
+      // 1.2.3.0, but not 1.2.0.0 (3rd digit 0 equals beta) nor 1.2.0.4 (4th digit non-zero equals beta too)
+      "^(?<major>\\d+)\\.(?<minor>\\d+)\\.(?<patch>[1-9]\\d*)\\.0$",
       ["emby/embyserver"]
     ),
     customVersioning(
--- a/ix-dev/stable/emby/app.yaml
+++ b/ix-dev/stable/emby/app.yaml
@@ -52,4 +52,4 @@ sources:
 - https://hub.docker.com/r/emby/embyserver
 title: Emby Server
 train: stable
-version: 1.3.10
+version: 1.3.11
--- a/ix-dev/stable/emby/ix_values.yaml
+++ b/ix-dev/stable/emby/ix_values.yaml
@@ -2,6 +2,9 @@ images:
   image:
     repository: emby/embyserver
     tag: 4.9.1.7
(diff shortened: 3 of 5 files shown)


stavros-k:
Source for this?

jcassette:
```
~/Dev/Emby.Releases$ git tag -l | egrep '^4\.[0-9]*\.[1-9][0-9]*'
4.0.1.0
4.0.2.0
4.0.3.0
4.1.1.0
4.2.1.0
4.3.1.0
4.4.1.0
4.4.2.0
4.4.3.0
4.5.1.0
4.5.2.0
4.5.3.0
4.5.4.0
4.6.1.0
4.6.2.0
4.6.3.0
4.6.4.0
4.6.5.0
4.6.6.0
4.6.7.0
4.7.1.0
4.7.10.0
4.7.11.0
4.7.12.0
4.7.13.0
4.7.14.0
4.7.2.0
4.7.3.0
4.7.4.0
4.7.5.0
4.7.6.0
4.7.7.0
4.7.8.0
4.7.9.0
4.8.1.0
4.8.10.0
4.8.11.0
4.8.3.0
4.8.4.0
4.8.5.0
4.8.6.0
4.8.7.0
4.8.8.0
4.8.9.0
4.9.1.0
4.9.1.1
4.9.1.10
4.9.1.11
4.9.1.2
4.9.1.3
4.9.1.4
4.9.1.5
4.9.1.6
4.9.1.7
4.9.1.8
4.9.1.9
```

stavros-k:
I don't get what this list shows? I can still see a `.9` for example is this not a stable release?.
Any source that says that `.0` is always a stable release AND all others are not?

jcassette:
Emby 4.8 is stable and 4.9 is beta, see https://emby.media/linux-server.html

That list shows the current filter:

> 1.2.3.4, but not 1.2.0.4 (3rd digit 0 equals beta)

So you can see that:
- all stable releases (4.8 and prior) end with .0
- releases ending with .1-11 are not stable (because 4.9) so they should be excluded

stavros-k:
> all stable releases (4.8 and prior) end with .0

Filter for `3rd digit != 0` == `stable` also works fine here. I don't see any invalid matches.

The issue is with 4.9.x.x not adhering to none of the rules
- `3rd digit != 0` == `stable`
- `4rd digit == 0` == `stable`

4.9.1.0 according to the above rules, it should be stable, yet its not.
https://github.com/MediaBrowser/Emby.Releases/releases/tag/4.9.1.0

So, it doesn't really fix it.
We need Emby official response on how we can distinguish stable from beta.
Either a versioning scheme or a suffix/prefix (ie `-stable`)

stavros-k:
The image selector will only work for new installs. Existing installs will fail to downgrade.
And I'd very much like to avoid having to have migration to remove that key (image selector) from the schema as soon as emby releases stable 4.9. I remember seeing a comment from emby recently that stable should be happening soon

stavros-k:
Closing this for now; until we get a conclusion

PR #2830, opened 2025-07-21
Add Hoppscotch

TigerGorilla2:
Hey, I would like some feedback on this.
There are still some [todos](https://github.com/TigerGorilla2/truenas-apps/blob/take-two/ix-dev/community/hoppscotch/todo.md)
and I have some questions:
- should i add "additional_storage" even tho the app doesn't use or expose it?
- app.yaml: are the keywords good? (is there a set to choose from?)
- app.yaml: is ixsystems as maintainers ok?
- should the app run as root? or apps? or something else? is there an app that I can use as a reference implementation?
- item.yaml: are the tags good? (is there a set to choose from?)
- are there too many settings exposed to users? or should the reset of jwt settings also be exposed?
- should there be a note about [telemetry](https://docs.hoppscotch.io/documentation/self-host/community-edition/telemetry) somewhere?

Some stuff to note:
- [using https on 443](https://docs.hoppscotch.io/documentation/self-host/community-edition/prerequisites#support-for-standard-http%2Fs-ports) should be supported but didn't work for me
- I added .github/scripts/selinux-ci.py for use on my fedora atomic
- questions.yaml is still split up for better overview and generate-questions.sh combines it


stavros-k:
Please combine questions file and remove extraneous files like the my.env, todo.md, comose file, scripts etc.
After that I can start reviewing and if you agree I can push changes to polish/finish this.

TigerGorilla2:
Files are removed and edits are now allowed :)

stavros-k:
> * should i add "additional_storage" even tho the app doesn't use or expose it?

Added

> * app.yaml: are the keywords good? (is there a set to choose from?)

It's fine, we can always update later.

> * app.yaml: is ixsystems as maintainers ok?

Yep, that's the only valid one at this point.

> * should the app run as root? or apps? or something else? is there an app that I can use as a reference implementation?

Ideally no, but peaking at one of their containers, seems everything is owned by root.
Luckily there is no storage tied to those containers, so if we discover it can be run as non-root later, we can easily update

> * item.yaml: are the tags good? (is there a set to choose from?)

Should be insync with keywords (in fact ci.py autogenerates the item.yaml)

> * are there too many settings exposed to users? or should the reset of jwt settings also be exposed?

I trimmed some of them, users can still sue additional envs to add them

> * should there be a note about [telemetry](https://docs.hoppscotch.io/documentation/self-host/community-edition/telemetry) somewhere?

Unless there is a way to expose a toggle in the UI, its up to the user to study what the app they are installing.

stavros-k:
That being said, I find their docs lacking. Like a lots of things.

- Is the oauth scope/tenant for each of the providers kinda static? Or supposed to be edited by the user? I currently have them hardcoded and not exposed
- Switched all containers to use port 80 as they mention somewhere in the docs that other ports will be removed in the future. (I find it also bad, as the internal ports does not really matter, but they impose higher privileges (NET_BIND_SERVICE)
- Does containers talk to each other? Or all happen on the browser. If they talk to each other, which envs sets the internal URLS? For example a container might not resolve the correct IP (LAN) when using sub.example.com, as it relies on the user to have correctly setup hairpin nat.
- Should we bundle an nginx container so we have to only expose less ports? (I think app bundle has to be separate.) (Is there a nginx config suggested by hoppscotch?)

stavros-k (approved):

stavros-k:
Merging for now and we can revisit

PR #2814, opened 2025-07-17
chore(deps): update updates-patch-minor

truenasbot:
This PR contains the following updates:

| Package | Update | Change |
|---|---|---|
| alexta69/metube | patch | `2025-07-01` -> `2025-07-27` |
| [amir20/dozzle](https://redirect.github.com/amir20/dozzle) | patch | `v8.13.6` -> `v8.13.8` |
| [automaticrippingmachine/automatic-ripping-machine](https://redirect.github.com/automatic-ripping-machine/automatic-ripping-machine) | patch | `2.17.2` -> `2.17.4` |
| [baserow/baserow](https://gitlab.com/baserow/baserow) | patch | `1.34.2` -> `1.34.5` |
| [brandawg93/peanut](https://redirect.github.com/Brandawg93/PeaNUT) | minor | `5.8.0` -> `5.10.0` |
| [chrislusf/seaweedfs](https://redirect.github.com/seaweedfs/seaweedfs) | minor | `3.94` -> `3.95` |
| collabora/code | patch | `25.04.3.2.1` -> `25.04.4.1.1` |
| dbeaver/cloudbeaver | patch | `25.1.2` -> `25.1.3` |
| [docker.elastic.co/elasticsearch/elasticsearch](https://www.elastic.co/products/elasticsearch) ([source](https://redirect.github.com/elastic/elasticsearch)) | patch | `9.0.3` -> `9.0.4` |
| dpage/pgadmin4 | minor | `9.5` -> `9.6` |
| elasticsearch | patch | `9.0.3` -> `9.0.4` |
| emby/embyserver | patch | `4.9.1.7` -> `4.9.1.11` |
| [fallenbagel/jellyseerr](https://redirect.github.com/fallenbagel/jellyseerr) | patch | `2.7.1` -> `2.7.2` |
| [filebrowser/filebrowser](https://redirect.github.com/filebrowser/filebrowser) | minor | `v2.40.1` -> `v2.42.0` |
| [fireflyiii/core](https://redirect.github.com/firefly-iii/docker) ([source](https://dev.azure.com/Firefly-III/_git/MainImage)) | patch | `version-6.2.20` -> `version-6.2.21` |
| [fireflyiii/data-importer](https://redirect.github.com/firefly-iii/docker) ([source](https://dev.azure.com/Firefly-III/_git/ImportToolImage)) | patch | `version-1.7.6` -> `version-1.7.8` |
| fosrl/newt | patch | `1.3.2` -> `1.3.4` |
[...]

PR #2837, opened 2025-07-25
Update Immich ML Images

Attack8:
This adds on to #2814, updating the Immich machine learning models with the app update


Attack8:
Fixed by 95d150d73cc69e783e81cfe56009bde425318257

PR #2845, opened 2025-07-28
nextcloud: remove imaginary compose dependency

### LINKED ISSUE #2843: nextcloud app - depends for imaginary and nextcloud are reversed
In the stable nextcloud app, when imaginary is enabled, a depends is added to the imaginary container that makes it so it won't start until nextcloud is running and healthy: https://github.com/truenas/apps/blob/6de11ec2c3f0177ee7a785c2ebce4c9ccfe407f7/trains/stable/nextcloud/2.0.10/templates/docker-compose.yaml#L187

This seems reversed to me. The way it's set up:
Nextcloud will start
then Imaginary will start

When shutting down
Imaginary will stop
Then Nextcloud will stop

Various nextcloud apps/features use imaginary, so if the imaginary service is not running, there can be errors. For example, if the `facerecognition` app is scanning photos for faces, and nextcloud shuts down, the background job will start erroring because the imaginary service is no longer available to generate smaller photos for image processing. This causes photos to be skipped for processing.

I think that nextcloud should have a depends on the imaginary container, not the other way around. Am I missing something, or some requirement of imaginary, that needs nextcloud to be running first?

stavros-k:
Closes #2843

### DIFF
--- a/ix-dev/stable/nextcloud/templates/docker-compose.yaml
+++ b/ix-dev/stable/nextcloud/templates/docker-compose.yaml
@@ -184,7 +184,6 @@
   {% set imaginary_container = tpl.add_container(values.consts.imaginary_container_name, "imaginary_image") %}
   {% do imaginary_container.add_caps(["SYS_NICE"]) %}
   {% do imaginary_container.set_user(568, 568) %}
-  {% do imaginary_container.depends.add_dependency(values.consts.nextcloud_container_name, "service_healthy") %}
   {% do imaginary_container.healthcheck.set_test("wget", {"port": 9000, "path": "/health"}) %}
 {% endif %}
 
--- a/ix-dev/stable/nextcloud/app.yaml
+++ b/ix-dev/stable/nextcloud/app.yaml
@@ -76,4 +76,4 @@ sources:
 - https://github.com/nextcloud/docker
 title: Nextcloud
 train: stable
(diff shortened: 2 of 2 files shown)


ix-dev/stable/nextcloud/templates/docker-compose.yaml
```diff
@@ -184,7 +184,6 @@
   {% set imaginary_container = tpl.add_container(values.consts.imaginary_container_name, "imaginary_image") %}
   {% do imaginary_container.add_caps(["SYS_NICE"]) %}
   {% do imaginary_container.set_user(568, 568) %}
-  {% do imaginary_container.depends.add_dependency(values.consts.nextcloud_container_name, "service_healthy") %}
```
dtczest, line 187: Since apps/services in nextcloud depend on imaginary and will not work properly unless it's running, I think the reverse of this should be added:
`{% do nextcloud_container.depends.add_dependency(values.consts.imaginary_container_name, "service_healthy") %}`

ManiacDC, line 187: Since apps/services in nextcloud depend on imaginary and will not work properly unless it's running, I think the reverse of this should be added:
`{% do nextcloud_container.depends.add_dependency(values.consts.imaginary_container_name, "service_healthy") %}`

PR #2854, opened 2025-07-29
immich: fix renovate issue

stavros-k:
Renovate was having trouble updating the ML images, after some testing on a fork, this change fixed the problem.
I can only assume that there is an issue in cache key creation or changing.


stavros-k:
Never mind. I still see the same issue on my fork. Seems very flakey

PR #2884, opened 2025-08-02
Adds PEP 723 compliant inline metadata

StreamOfRon:
This is a convenience PR to add PEP-723 metadata to the scripts so that they can be directly invoked with the proper dependencies using a tool such as `uv run`, without needing to directly manage the python dependencies.  Since they're based upon comment syntax, tools which do not implement this (or direct invocation of the script) remain unaffected.

Reference:
https://packaging.python.org/en/latest/specifications/inline-script-metadata/#inline-script-metadata
https://docs.astral.sh/uv/guides/scripts/#declaring-script-dependencies

I listed the Python version constraint as Python>=3.11 since that's what is on the current TrueNAS Scale stable release but I noticed a couple of things have been bumped to 3.13 -- let me know if I should make that change here in advance of an upcoming release.

### DIFF
--- a/.github/scripts/ci.py
+++ b/.github/scripts/ci.py
@@ -1,4 +1,12 @@
 #!/usr/bin/env python3
+
+# /// script
+# requires-python = ">=3.11"
+# dependencies = [
+#     "pyyaml",
+# ]
+# ///
+
 import subprocess
 import argparse
 import pathlib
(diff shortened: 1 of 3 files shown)


stavros-k:
All those scripts run on CI or local machines and dont depend on what python version truenas runs.
Current constrain is fine tho, I dont think we use any >3.11 python feature.

Thanks

PR #2894, opened 2025-08-04
Gramps Web: increase celery health check timeout

DavidMStraub:
Deploying Gramps Web, the app stayed in "Deploying" forever as the celery container reported it was still starting, even though the log showed it was ready.

I found out this was due to me enabling semantic search (off by default), which loads a PyTorch model on Celery initialization that takes a couple of seconds. On my machine, the health check command `celery -A gramps_webapi.celery inspect ping` took 11 seconds. Since only 5 seconds is allowed by default, this failed.

I now increased the timeout to 20 seconds, which should be enough on most machines.

If this solution is not desired, we could also try to come up with a simpler health check command, or with none at all, although the one used is indeed the most meaningful one.

### DIFF
--- a/ix-dev/community/gramps-web/templates/docker-compose.yaml
+++ b/ix-dev/community/gramps-web/templates/docker-compose.yaml
@@ -15,8 +15,7 @@
 {% set redis = tpl.deps.redis(values.consts.redis_container_name, "redis_image", redis_config, perm_container) %}
 
 {% do web.healthcheck.set_test("tcp", {"port": values.consts.internal_web_port}) %}
-{% do celery.healthcheck.set_timeout(20) %}
-{% do celery.healthcheck.set_custom_test("celery -A gramps_webapi.celery inspect ping") %}
+{% do celery.healthcheck.set_custom_test("pgrep -f 'celery worker' > /dev/null") %}
 {% do celery.set_command([
   "celery", "-A", "gramps_webapi.celery",
   "worker", "--loglevel=INFO", "--concurrency=2"


trains/community/gramps-web/app_versions.json
```diff
@@ -56,7 +56,7 @@
             ],
             "title": "Gramps Web",
             "train": "community",
-            "version": "1.1.7"
```
stavros-k: please dont touch files under `trains/` directory

DavidMStraub: Got it, sorry.

stavros-k:
> On my machine, the health check command `celery -A gramps_webapi.celery inspect ping` took 11 seconds.

Why does it take so much time to respond? Does it always take 11 seconds or only while model is loading?
If its only during model loading, we can leave it as is, and once everything is _really_ ready and respond in time, healthcheck should respond healthy.

If it always takes so much time, I think this is a bit too expensive for healthcheck. Healthcheck is meant to be a quick check that nothing went horribly wrong.

DavidMStraub:
> If it always takes so much time, I think this is a bit too expensive for healthcheck. Healthcheck is meant to be a quick check that nothing went horribly wrong.

Understood. Yes, it always takes so much time because the command spawns a new celery instance, so it has to do all the Flask app initialization work.

I'll think of a different command.

Is the health check compulsory, actually?

stavros-k:
> > If it always takes so much time, I think this is a bit too expensive for healthcheck. Healthcheck is meant to be a quick check that nothing went horribly wrong.
> 
> Understood. Yes, it always takes so much time because the command spawns a new celery instance, so it has to do all the Flask app initialization work.
> 
> I'll think of a different command.

Does celery listen on any port internally? Maybe a curl/wget to something like /ping should be enough.

> Is the health check compulsory, actually?

Healthchecks determine the status on the truenas webui (and in some cases used for other containers to wait other containers to become ready). If there are no healthchecks and the container is crashed, webui will show that everything is fine.
So, I'd say its best to have one.

DavidMStraub:
I changed to a simple `pgrep` now, what do you think?

stavros-k:
> I changed to a simple `pgrep` now, what do you think?

I suppose it should work.
if the process exits/fails container will anyway be restarted.
then if pgrep finds it should mean that is running.

Issue #2375, opened 2025-05-26
Homarr Docker widget not working


coinflipcoder:
Since [Homarr v1.21.0](https://github.com/homarr-labs/homarr/releases/tag/v1.21.0), there is a docker widget. This does not work, with or without the "mount docker socket" option checked. 
This is from the log:

`error tRPC Error with query on 'docker.getContainers'`
`error Failed to update Docker container status`

Meierschlumpf:
Hey, maintainer from Homarr here,  @coinflipcoder can you create an issue on our repository https://github.com/homarr-labs/homarr/issues/new/choose where we can assist you in getting it working?

stavros-k:
I'll close this issue, you can follow the upstream issue for updates. Thanks

PR #2906, opened 2025-08-06
Homarr change default to root:root

HubGon:
Hello everyone,
although in the description it is written, that the App Homarr is run as root, the min value of the uid and gid is set to 568, so by newly creating the app, the user is apps:apps. That is causing the problem, that Homarr can not access the docker.sock and no Docker statistics and be retrieved in the Dashboard.
Kind regards


stavros-k:
Hello, yes homarr runs as root even if you set a different UID there. 
What happens is homarr starts subprocesses as the non privileged user.

Lots of other containers work with unprivileged user and have proper docker socket.
I suspect that when starting the subprocesses with the unprivileged user, this user does get the correct privileges to access the socket. 
ie make this user be part of the docker group (in this case `999`)

I'd really prefer if homarr fixes this on their side instead.

stavros-k:
See https://github.com/truenas/apps/issues/2375

HubGon:
Hey,
thank you for clarification. I think this PR can then be closed without being pushed 😀
Kind regards

PR #2935, opened 2025-08-11
apps: use valkey

stavros-k:
- Replaces bitnami images (redis and 1 bitnami/git)
- Updates library to pass config via flags instead of envs (official redis/valkey does not support envs)
- bumps library on touched apps so it correctly setup redis
- updates metadata now that redis/valkey can run as a non root user fully


stavros-k:
Invoice ninja test failure is expected (#2924)

Qubad786 (approved):

PR #2938, opened 2025-08-12
add linkwarden

krojew:


krojew:
@stavros-k thank you for the review, but I have to ask why did you change linkwarden uid to 0? It doesn't need to run as root.

stavros-k:
> @stavros-k thank you for the review, but I have to ask why did you change linkwarden uid to 0? It doesn't need to run as root.

While it mostly works,
It fails to generate previews. Puppeteer needs root or at least the image is setup to work with root only.

krojew:
Ok, that makes sense.

stavros-k:
@krojew ugh when merged the target branch this closed automatically, can you reopen and target master branch now please?

stavros-k:
nvm fixed it.

PR #2957, opened 2025-08-13
Add Element web

### LINKED ISSUE #266: Community Chart Suggestion: Element web
https://github.com/element-hq/element-web

HubGon:
Hi,

this is a Draft for the Element-Web app.
The only configuration mentioned with env-variables is build into the draft (only 1). Everything else should be configured through a json file, located at "/app/config.json". [Docs](https://github.com/element-hq/element-web/blob/develop/docs/install.md#docker)
I tried to mount the file (like mentioned in the doc), but that's not possible with ix_volumes it seems. When i tried to mount the whole "/app" directory, there seem to be a permission error, since the app can not read the config anymore (a permission container doesn't seem to fix it). Maybe it is better user mount the file themself though host path. With host path it is possible to mount a single file.
Besides that there is no health-check endpoint, like mentioned in this [Issue](https://github.com/element-hq/element-web/issues/29548). Therefor i used the mentioned Workaround with the config.json. This is working fine too.
At last there is also a possibility to mount the "/module" directory to load modules into the app. Since this is not needed for basic working for the app, i did not implemented it, to keep the app simple to start with.

Icon for the app: https://element.io/assets-32bb636196f91ed59d7a49190e26b42c/5ef25c0d30ee3108da4c25e9/5f365d3197194f8c73b00112_logo-mark-primary.svg

Kind regards

Closes #266

### DIFF
--- a/ix-dev/community/element-web/README.md
+++ b/ix-dev/community/element-web/README.md
@@ -1,0 +1,3 @@
+# Element-Web
+
+[Element-Web](https://element.io/) is a glossy Matrix collaboration client for the web.
--- a/ix-dev/community/element-web/app.yaml
+++ b/ix-dev/community/element-web/app.yaml
@@ -1,0 +1,32 @@
+app_version: v1.11.109
+capabilities: []
+categories:
+- productivity
+changelog_url: https://github.com/element-hq/element-web/releases
+date_added: '2025-08-14'
+description: Element-Web is a glossy Matrix collaboration client for the web.
+home: https://element.io/
+host_mounts: []
+icon: https://media.sys.truenas.net/apps/element-web/icons/icon.svg
+keywords:
+- chat
+- element
+lib_version: 2.1.46
+lib_version_hash: 464cacce9ffec31b1819cd064a4d1099af6d232cb3d870da847a885e00ed12ce
+maintainers:
+- email: dev@ixsystems.com
+  name: truenas
+  url: https://www.truenas.com/
+name: element-web
+run_as_context:
+- description: Element-Web runs as any non-root user.
+  gid: 568
+  group_name: element-web
+  uid: 568
+  user_name: element-web
+screenshots: []
+sources:
+- https://github.com/element-hq/element-web
+title: Element-Web
+train: community
+version: 1.0.0
--- a/ix-dev/community/element-web/item.yaml
+++ b/ix-dev/community/element-web/item.yaml
@@ -1,0 +1,7 @@
+categories:
+- productivity
+icon_url: https://media.sys.truenas.net/apps/element-web/icons/icon.svg
+screenshots: []
+tags:
+- chat
+- element
--- a/ix-dev/community/element-web/ix_values.yaml
+++ b/ix-dev/community/element-web/ix_values.yaml
@@ -1,0 +1,7 @@
+images:
+  image:
+    repository: ghcr.io/element-hq/element-web
+    tag: v1.11.109
+
+consts:
+  element_web_container_name: element-web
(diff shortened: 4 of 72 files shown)


stavros-k:
I think we should have a setup container to set couple things in the config.json.
I think the default_server_config is required? Any other settings? I'll create the setup container later today probably.
If you can gather a list with required or very popular fields, it would be nice.

HubGon:
Hey,
the container is creating an example config for basic installation at start-up, which looks like this (this is starting and working fine):
```
{
    "default_server_config": {
        "m.homeserver": {
            "base_url": "https://matrix-client.matrix.org",
            "server_name": "matrix.org"
        },
        "m.identity_server": {
            "base_url": "https://vector.im"
        }
    },
    "disable_custom_urls": false,
    "disable_guests": false,
    "disable_login_language_selector": false,
    "disable_3pid_login": false,
    "force_verification": false,
    "brand": "Element",
    "integrations_ui_url": "https://scalar.vector.im/",
    "integrations_rest_url": "https://scalar.vector.im/api",
    "integrations_widgets_urls": [
        "https://scalar.vector.im/_matrix/integrations/v1",
        "https://scalar.vector.im/api",
        "https://scalar-staging.vector.im/_matrix/integrations/v1",
        "https://scalar-staging.vector.im/api"
    ],
    "default_widget_container_height": 280,
    "default_country_code": "GB",
    "show_labs_settings": false,
    "features": {},
    "default_federate": true,
    "default_theme": "light",
    "room_directory": {
        "servers": ["matrix.org"]
    },
    "enable_presence_by_hs_url": {
        "https://matrix.org": false,
        "https://matrix-client.matrix.org": false
    },
    "setting_defaults": {
        "breadcrumbs": true
    },
    "jitsi": {
        "preferred_domain": "meet.element.io"
    },
    "element_call": {
        "url": "https://call.element.io",
        "participant_limit": 8,
        "brand": "Element Call"
    },
    "map_style_url": "https://api.maptiler.com/maps/streets/style.json?key=fU3vlMsMn4Jb6dnEIFsx"
}
```
As i understand the mentioned default_server_config is for default picked server for registration and login, which can be changed manually in the web-ui (this app allows to connect to all matrix servers, by default). See officially hosted [element-web ](https://app.element.io/#/welcome) for comparison. In my opinion it is not critical to change that config for basic installation.

Anyway, there is also a recommended production ready config, also hosted on the official element-web host: [config](https://app.element.io/config.json).

If you really want to create a config by user input through ui, i think the following config-options are the most interesting ones:
default_server_config --> By default picked matrix server
disable_custom_urls --> Disabling enter own matrix server

[All other options](https://github.com/element-hq/element-web/blob/develop/docs/config.md#customisation--branding) are, in my opinion too much for initial setup, but you have to decide.

Kind regards

stavros-k:
Well, we can't really update the config on the fly.

Because all files are owned by root, we cant use the `/docker-entrypoint.d/*` to have scripts run startup to modify it (unless we run it as root)
Also because the config is located in `/app/config.json` where `/app` also contains other app files, we cant mount something there without either wiping the content or breaking future upgrades.
Therefor we cant use init container to share the config file, in order to modify it on another container with higher privileges.

So it will be "as is".

HubGon:
Hey @stavros-k ,

i have installed the app on my truenas and with the standard port of 30246 it is not working...
Just out of curiosity i also tried 8080 (of the basic values - which the CI ran perfectly fine with it), it is working with it. It seems it isn't working with any other port than 8080 (tried different ones in the basic-values locally).

#Edit: Locally other Ports work too with this implementation. Only the installation in TrueNAS is not working with other ports.

Kind regards

PR #2962, opened 2025-08-14
element-web: change hc path

stavros-k:


stavros-k:
Covered this in #2963

PR #2977, opened 2025-08-17
Add Mediawiki

### LINKED ISSUE #175: Community charts suggestion: MediaWiki

HubGon:
Hey,
this is a first draft for the implementation of Mediawiki as an App.
This app is passing the CIs and can be accessed in web, but have still some todos.

1. After starting the containers, the frontend have to be populated with the database credentials manually. This is working fine in principle. After filling in the credentials and starting the setup, the database (mariadb) seem to be setting itself to a read only state (at least this is the error message the setup is giving).

2. The official Procedere upstream require an initial setup of the wiki and afterwards binding a file named "LocalSettings.php". --> described in comments in the example [docker-compose](https://hub.docker.com/_/mediawiki/#-via-docker-compose).
I my opinion the users have to do it by themself. If you want the user to do it in the UI nevertheless (using a macro?), here is the description of which settings can be set:
https://www.mediawiki.org/wiki/Manual:LocalSettings.php
#Edit: To make it maybe a little bit more clear, the file is creating at the installation process and the user can download it. The user does not have to create the file from scratch. 

3. As i couldn't complete the setup in testing, i don't exactly know if there are other problems with this app. So i let the user, which is running the main container, at 33:33 until upper problem is resolved. Then one could test other users too.
It could be, that the user have to be fixed at 33:33, as the "LocalSetting.php" have to readable by the web interface (www-data).

Besides that, here are some points, that i stumbled over as i tried to make this draft:

1. The here used, official, Image has no Environments Variables (or at least has no variables that i have found).
https://github.com/wikimedia/mediawiki-docker
https://hub.docker.com/_/mediawiki/

2. There is another image provided by Wikimedia, which takes Enviroment Variables, but this is for Development Purposes only and should not be used in Production, as they stated.
https://github.com/wikimedia/mediawiki/blob/master/docker-compose.yml
https://github.com/wikimedia/mediawiki/blob/master/DEVELOPERS.md#development-environment

At last, here is the logo for the app: https://github.com/docker-library/docs/blob/master/mediawiki/logo.svg

Kind regards

Closes #175

### DIFF
--- a/ix-dev/community/mediawiki/questions.yaml
+++ b/ix-dev/community/mediawiki/questions.yaml
@@ -457,6 +457,8 @@ questions:
                         enum:
                           - value: mediawiki
                             description: mediawiki
+                          - value: mariadb
+                            description: mariadb
   - variable: resources
     label: ""
     group: Resources Configuration


stavros-k:
I dont see a point of having an app if you have to manually configure it yourself and go back and forth the UI to bind mount another file to complete the setup.

if it was a json or yaml file, we could have an init script to edit it inplace.
But php file? Nope, we cant programmatically do that, and I'm not going to use tools like `sed` as it creates a million other issues.

sorry but unless mediawiki provides a more docker oriented way to bootstrap the app. I can't accept this.

HubGon:
I didn't saw it that way, until now. I think you are right.
Should I close this PR or do you just to pause it, until maybe upstream is changing?

stavros-k:
I dont think mediawiki will make any changes _soon_ enough that this PR will still be relevant. I'll go ahead and close it for now.

Thanks

PR #2989, opened 2025-08-19
storj: remove the requirement of the authorization step

elek:
It's not required any more: https://forum.storj.io/t/identity-signing-identity-authorize-is-not-required-any-more/30581

See also: https://github.com/storj/docs/pull/586


stavros-k:
Thanks, will push a migration in a bit as well to remove the field from the existing installs schema and merge it.

stavros-k (approved):
Thanks

PR #3010, opened 2025-08-22
Update questions.yaml

HubGon:
Hey,

i think it was a small mistake with the default port for the newly added TubeArchivist App.
This should be the next free one.

Kind regards.


stavros-k:
Thanks

PR #3017, opened 2025-08-22
Add restic rest-server

flufsor:
Hi Stavros,

This PR adds support for Rest Server, a high-performance HTTP server that implements Restic’s REST backend API.

The only manual step required is updating the .htaccess file inside the container to enable authentication—this is also necessary in the original setup, so no additional complexity is introduced.

One thing I’d like your input on is the health check. Since the container doesn’t include Bash, and both curl and wget fail on non-2xx/3xx status codes, they’re not ideal for testing endpoints that require authentication (which always return 401). I’m currently using netcat, but that results in a http: TLS handshake error in the logs for each test, which isn’t ideal either.

I’m considering postponing native SSL support and possibly submitting a PR to allow configurable status code ranges for health checks using curl or wget. Let me know what you’d prefer.

Also, the tool doesn’t have a dedicated logo, but since it’s maintained by the Restic team, we could potentially use the Restic logo for visual consistency.

Let me know if I’ve missed anything!

### DIFF
--- a/ix-dev/community/restic-rest-server/templates/test_values/host-networking-true.yaml
+++ b/ix-dev/community/restic-rest-server/templates/test_values/host-networking-true.yaml
@@ -19,6 +19,9 @@ rest_server:
 
 network:
   host_network: true
+  web_port:
+    bind_mode: published
+    port_number: 8000
 
 run_as:
   user: 568


stavros-k:
> The only manual step required is updating the .htaccess file inside the container to enable authentication—this is also necessary in the original setup, so no additional complexity is introduced.

We can generate that file. See `distribution` app for example.

For healthchecks I'll check later.
Please also use the `--listen` flag to set the port to match the web_port.port_number.

stavros-k:
Hello, I'll review this first thing tomorrow!

flufsor:
Added the listen flag and the htpassword generation and some cleanup before you reviewed.

Thanks for the comments already!

PR #3025, opened 2025-08-25
Optional htpasswd authentication for garage-webui

KittyKatt:
Hello! Garage was recently added to Apps as an alternative to MinIO and I've been loving it. However, the way that the App is setup has no way to leverage the optional htpasswd authentication you can add to the garage webui container. This is kinda important as the webui allows you to do many modifications or destructive actions against buckets/keys/etc in the Garage instance; allowing users to enable the builtin functionality of specifying a username and password (htpasswd hashed) to the App definition is critical, in my opinion.

This PR adds additional optional questions that will result in a rendered docker-compose that has the AUTH_USER_PASS environment variable passed to the garage-webui container. I've also added an additional test file that can be used to render a docker-compose that has these options enabled. I'm uncertain the best way to test this on my local TrueNAS install that will allow me to be certain these questions are rendered correctly and behave like I expect them to.

This is my first time submitting anything to TrueNAS Apps, so please let me know anything I've done wrong and I'll do my best to fix it!


stavros-k (approved):
lgtm, thanks

KittyKatt:
Awesome, thanks!

PR #3050, opened 2025-08-29
tube-archivist: allow multiple app urls

### LINKED ISSUE #3049: Tube Archivist - Can't use multiple hostnames
According to the [Tube Archivist Documentation](https://docs.tubearchivist.com/installation/env-vars), you should be able to specify multiple hostnames space separated in the TA_HOST environment variable (for example, if you want it to work with both http://x.x.x.x:x and https://ta.x.com). However, with the current schema entry being defined as ["type: uri"](https://github.com/truenas/apps/blob/3c901d3e1cf860da5329916f0789f79be8fed33d/trains/community/tubearchivist/1.0.3/questions.yaml#L53), it doesn’t seem possible to enter multiple hostnames.

stavros-k:
Closes #3049


ix-dev/community/tubearchivist/templates/docker-compose.yaml
```diff
@@ -18,7 +18,7 @@
 {% do c1.environment.add_env("ELASTIC_PASSWORD", values.tube.elastic_password) %}
 {% do c1.environment.add_env("TA_PORT", values.network.web_port.port_number) %}
 {% do c1.environment.add_env("TA_BACKEND_PORT", values.network.web_port.port_number+1) %}
-{% do c1.environment.add_env("TA_HOST", values.tube.app_url) %}
+{% do c1.environment.add_env("TA_HOST", values.tube.app_urls|join(",")) %}
```
veronoicc: I’m not sure if I’m correct, but wouldn’t this result in TA_HOST being something like "http://example1.com,http://example2.com"? From what I can see in the documentation, the hostnames should be space-separated, not comma-separated.
```suggestion
{% do c1.environment.add_env("TA_HOST", values.tube.app_urls|join(" ")) %}
```

stavros-k: Thanks!

PR #3083, opened 2025-09-04
Add Application: Traefik

N-Hertstein:
Working on adding Traefik in a way that makes sense.
Will not be as configurable as doing it yourself, but maybe we'll get there in the future :)

Feel free to suggest things that anyone is missing, correct me (but please explain why what I may have done is less then ideal).


stavros-k:
Hello, I'd like to suggest to not pick that app. Its complexity is not a good starting point. And besides that there are features that traefik needs that are not suported yet. (for example networks).

It will be really hard for me to review this app when done by someone that has no experience with the catalog at all.
Even simple apps done from someone new, are sometimes basically me re-doing the whole app, spending twice the time.

Thanks

N-Hertstein:
> Hello, I'd like to suggest to not pick that app. Its complexity is not a good starting point. And besides that there are features that traefik needs that are not suported yet. (for example networks).
> 
> It will be really hard for me to review this app when done by someone that has no experience with the catalog at all. Even simple apps done from someone new, are sometimes basically me re-doing the whole app, spending twice the time.
> 
> Thanks



I understand I'm a new face.
The primary reason I had some difficulty at the start was due to a lack of information. Doing some yaml wasn't the hard part.
I had to figure out the terms to use for this specific catalog and went moderately insane spending a few hours reading jinja/yaml & python without the guidence of comments.

The app should be basically complete enough to be able to handle anything basic. This includes HTTP & TLS challenges. I will finish up The TLS challenge configuration, clean up my commentary and do some tests. I'll leave the DNS challenge for now.

 If you wouldn't mind taking a quick look to see if it adheres to the nessecary standards that would be cool (don't take time to rewrite an app that won't yet work due to a lack of networking functionality).
 If it happens to be some solid groundwork I would like to propose labeling it as traefik-unstable and moving it to the dev or test branch to pick up at a later time when networking is added.
 The reason I suggest this is that there seems to be a large demand for Traefik (including from me) and having some groundwork to build upon in an app with this many variables might make it easier for me or another Traefik enjoyer to finish when there are more options for docker networking.
 
 Otherwise I'll slowly build upon it, keeping it in the ol' backpocket for when the time is right.
 And I may dig into getting an option for adding a traefik-proxy network to the docker compose template and connecting specific containers to it (as a database of any form has no business being proxied when only in use by lets say nextcloud)
 
Either way I'll be making some documentation after this for what does what and means what. If you'd be willing to help me with a style-guide and maybe some general additions and/or corrections, we might be able to lighten your "I have to rewrite this new guys app again" workload.

N-Hertstein:
@stavros-k  It's ready.
Works on my machine, was even able to connect a service to it with some manual tweaking.
Launches on truenas.

If you are open to it I would like to place it in the test branch so I can play with all the settings with the truenas interface and fix any issues with HTTP & TLS challenges. I can test these by using labels to proxy the treafik dashboard. Though I may modify it and temporarily add a whoami container in the near future.

After this I will be grabbing my shovel and start digging into the networking limitations and possible resolutions.

My plan is to:
- Get an option to add a network with a specific name to specific containers (this way databases will not be exposed unless specifically conmfigured.
- Build a dummy app to test the HTTP and TLS challenges
- Write documentation on how to add traefik compatibility to anyones application (in a way that will not be harmfull when Traefik is not installed. Preferably in a way that is (mostly) copy & paste.
- Release it as a beta (as the dns challenge will be missing and is in my opinion a major functionality that is the primary reason for people wanting this (correct me if I am wrong))
- Expand documentation for creating custom apps (may do this before making Traefik beta ready as I'd forgotten a bunch of things already after a week and having a proper point of reference will be helpfull)
- Allow for adding custom environment variables
- Start working on DNS challenges in a way that makes it as simple as possible for users to use their dns provider for getting certificates. I'll be adding these one at a time and we'll have to see how often to make it a new release

If you would be so kind as to at least take a look and inform me if there are things that need changing. Whether it is comments or descriptions, the order of certain things or perhaps things I've forgotten or overcomplicated.
Do however do yourself a favor and waste no time rewriting an app that will currently only show you a cool dashboard and do little else. Unless you have a strong desire to help ofcourse :)

stavros-k:
Hello again, I won't be reviewing (not anytime soon) as I've already mentioned as soon as I saw your PR.
Treafik is a different kind of beast and needs careful implementation, cause I will be the one having to maintain this. And migrations is not fun.

> Though I may modify it and temporarily add a whoami container in the near future.

I don't think this is needed.

> * Get an option to add a network with a specific name to specific containers (this way databases will not be exposed unless specifically conmfigured.

Please don't, at least don't try to add something regarding this in the library. This awaits middleware integration first and will drive the library code.

> * Write documentation on how to add traefik compatibility to anyones application (in a way that will not be harmfull when Traefik is not installed. Preferably in a way that is (mostly) copy & paste.

I'm not sure where this documentation will leave. And I dont get why we need such documentation? Doesnt traefik already has it?
Unless you want to start exposing every knob traefik has on the UI, which is not going to happen either.

> Allow for adding custom environment variables

Its already supported on all apps. I dont know what you mean.

> * Expand documentation for creating custom apps (may do this before making Traefik beta ready as I'd forgotten a bunch of things already after a week and having a proper point of reference will be helpfull)

You are free to submit a PR for adding documentation. 
But there isn't much to it. All apps follow a similar pattern. 
Best way to create a new app is to copy an existing and modify. 

There is hardly any IDE integration to help you manually create an app from scratch.
Even my self just do that. Copy and modify. Any other way is guaranteed to miss stuff, been there, done that.

ix-dev/test/traefik/templates/docker-compose.yaml
```diff
@@ -0,0 +1,121 @@
...
+  {% do traefik.add_port(values.network.https_port) %}
+{% endif %}
+
+
+{# # ENVIRONMENT VARIABLES DO NOT YET SEEM ESSENTIAL, NEED TO WORK ON LATER
+# Add Environment variables
+{% do traefik.environment.add_env("ENV_Variable", values.#VARIABLE#) %}
+
+# Add conditional environment variables
+{% if values. %}
```
stavros-k, line 93: this feels like a syntax error

ix-dev/test/traefik/templates/docker-compose.yaml
```diff
@@ -0,0 +1,121 @@
...
+  {% do commands.x.append("--certificatesresolvers.truenasresolver.acme.storage=/letsencrypt/acme.json") %}
+{% endif %}
+{# TLS Challenge #}
+{% if values.proxy.challenge == "tls_challenge" %}
+  {% do commands.x.append("--certificatesresolvers.truenasresolver.acme.tlschallenge=true") %}
+  {% do commands.x.append("--certificatesresolvers.truenasresolver.acme.email=%s" | format(values.proxy.email)) %}
+  {% do commands.x.append("--certificatesresolvers.truenasresolver.acme.storage=/letsencrypt/acme.json") %}
+{% endif %}
+
+{# DNS Challenge BUILD AT SOME POINT (one provider at a time) #}
```
stavros-k, line 58: nope, there wont be such thing. providers is an infinite increasing list.
yes I know that ddns-update does simillar thing, but that is because it replaced a system service for ddns.

ix-dev/test/traefik/templates/docker-compose.yaml
stavros-k, line 1: Overall you are going too much over use cases rather than deployment setup.
Only flags that affect how app is interfaces with the system should be used.
Ie ports, storage etc.

Rest are just use case specific. And for traefik there are a million use cases and frankly a UI is not good for configuring those.

ix-dev/test/traefik/templates/docker-compose.yaml
```diff
@@ -0,0 +1,121 @@
+{# Generate docker compose template #}
+{% set tpl = ix_lib.base.render.Render(values) %}
+
+{# Create traefik container
+   Get name from TrueNAS UI
+   Get container image & tag from ix_values.yaml #}
+{% set traefik = tpl.add_container(values.consts.traefik_container_name, "image") %}
+
+{# Configure Healthcheck #}
+{% do traefik.healthcheck.set_test("http", {"port": values.network.web_port.port_number}) %}
```
stavros-k, line 10: I doubt this will work with secured api

N-Hertstein: It will not, but I haven't created that option yet. Will update with an if statement to switch health check types.

N-Hertstein:
> Hello again, I won't be reviewing (not anytime soon) as I've already mentioned as soon as I saw your PR. Treafik is a different kind of beast and needs careful implementation, cause I will be the one having to maintain this. And migrations is not fun.

I was not planning on abandoning it as soon as it was complete, but I understand that means little to anyone when coming from a stranger.

> > Though I may modify it and temporarily add a whoami container in the near future.
> 
> I don't think this is needed.

It allows for testing with linking a secondary container to Traefik. No need for something fancy, traefik/whoami gives me something to curl and verify whether things are functional.

> > * Get an option to add a network with a specific name to specific containers (this way databases will not be exposed unless specifically conmfigured.
> 
> Please don't, at least don't try to add something regarding this in the library. This awaits middleware integration first and will drive the library code.

I'll want to familiarize with the inner workings of the library anyways. I might be able to "trick" it into adding a second network to certain containers. It justs needs to render the network into the docker-compose file. The rest is default docker functionality after all.
But I will leave it alone if this is not possible.

> > * Write documentation on how to add traefik compatibility to anyones application (in a way that will not be harmfull when Traefik is not installed. Preferably in a way that is (mostly) copy & paste.
> 
> I'm not sure where this documentation will leave. And I dont get why we need such documentation? Doesnt traefik already has it? Unless you want to start exposing every knob traefik has on the UI, which is not going to happen either.

Think a section for apps that says:
Add to Traefik (checkbox)
Domain name (string)
Additional important options
Optional/custom options

Sure this can be done with labels, but having a simple copy-paste solution would be incredibly usefull in the future. 
And for enabling a connection to the proxy network (when such functionality exists) will require a checkmark anyways.

> > Allow for adding custom environment variables
> 
> Its already supported on all apps. I dont know what you mean.

I've not yet written/copied the question and implemented it in the docker-compose file.

> > * Expand documentation for creating custom apps (may do this before making Traefik beta ready as I'd forgotten a bunch of things already after a week and having a proper point of reference will be helpfull)
> 
> You are free to submit a PR for adding documentation. But there isn't much to it. All apps follow a similar pattern. Best way to create a new app is to copy an existing and modify.
> There is hardly any IDE integration to help you manually create an app from scratch. Even my self just do that. Copy and modify. Any other way is guaranteed to miss stuff, been there, done that.

I've had to do significant reading of many apps and the python code to figure things out. With the amount of things out there a simple copy-paste isn't optimal. It's the best way to start, but any serious (or even mild sometimes) deviations or additional options between two similar apps requires an enormous amount of investigation.
It doesn't have to be big, but a list of commands, best practices and that we use jinja2 for the docker-compose template will save many a lot of time.

A list of "taken" ports may also be benificial to reduce app installation conflicts

I understand you're familiar with all this. But an introduction for new people will save you (and me and other new people) headaches.

N-Hertstein:
@stavros-k 
I will be closing this pull request.
I have discovered [Traefik Network Connector](https://github.com/obeone/traefik_network_connector).
A tool that will connect the Traefik docker container to the networks of the containers that request to be proxied.
This **will** allow Traefik to genuinely function with the current apps implementation in TrueNAS.

I will investigate and implement this into the traefik app as a second container.

I **will**  make a new PR when I have reworked the app to a state where it will function with a HTTPS and TLS challenge.

PR #3087, opened 2025-09-04
add zoraxy

### LINKED ISSUE #680: Community Chart Suggestion: Zoraxy
https://github.com/tobychui/zoraxy

stavros-k:
Closes #680

### DIFF
--- a/cspell.config.yaml
+++ b/cspell.config.yaml
@@ -452,4 +452,5 @@ words:
   - zipline
   - zksync
   - zoneedit
+  - zoraxy
   - zwave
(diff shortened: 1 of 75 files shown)


Qubad786 (approved):

JakePeralta7:
Did you manage to run it on TrueNAS? It doesn't show the port
image
And it can't bind to the reverse proxy ports, even though the ports are not in use and the app have the NET_ADMIN capability
image

coinflipcoder:
@JakePeralta7 I have it running without host mode, with an "additional port" added that I mapped to 443 in the container. both this port and the webui port are published.

JakePeralta7:
> @JakePeralta7 I have it running without host mode, with an "additional port" added that I mapped to 443 in the container. both this port and the webui port are published.

Got it, will give it a try

stavros-k:
If you use host networking, you won't see any ports in the card. As it bypasses docker network entirely.
I'm adding an additional cap in #3114 which should fix the permission denied in the host mode.

PR #3096, opened 2025-09-06
Adding Redis and WebUI Secret Key to OpenWebUI

HubGon:
Hey,

with this PR i want to add
1. Redis Caching to OpenWebUI:
https://docs.openwebui.com/getting-started/env-configuration#redis

2. the Env. variable "WEBUI_SECRET_KEY" as an input variable. This variable overrides the randomly generated string used for JSON Web Token. In result the user can restore a login-session after OpenWebUI is restarted.
https://docs.openwebui.com/getting-started/env-configuration#webui_secret_key

I have also seen, that there is an Env. var to disable the update check in the app itself. I think this would make sense in this installation too, as updates only get pushed by this repo and not automatically upstream. I didn't implemented in this PR, as i don't know if this is deliberate.
https://docs.openwebui.com/getting-started/env-configuration#enable_version_update_check

Kind regards

### DIFF
--- a/ix-dev/community/open-webui/app.yaml
+++ b/ix-dev/community/open-webui/app.yaml
@@ -30,10 +30,15 @@ run_as_context:
   group_name: open-webui
   uid: 568
   user_name: open-webui
+- description: Redis runs as any non-root user.
+  gid: 568
+  group_name: redis
+  uid: 568
+  user_name: redis
 screenshots:
 - https://media.sys.truenas.net/apps/open-webui/screenshots/screenshot1.png
 sources:
 - https://github.com/open-webui/open-webui
 title: Open WebUI
 train: community
-version: 1.1.18
+version: 1.1.19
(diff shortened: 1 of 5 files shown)


stavros-k:
> I have also seen, that there is an Env. var to disable the update check in the app itself. I think this would make sense in this installation too, as updates only get pushed by this repo and not automatically upstream. I didn't implemented in this PR, as i don't know if this is deliberate.

If its just notifications we can leave it there. Doesnt hurt.

---

While I dont have a problem adding redis/secret handling, your PR will block everyone from updating. There needs to be migration to prefill the values for existing installations.

Now what happens if someone changes the secret key? Will it just invalidate the sessions or does it do encryption as well?

HubGon:
It does only invalid the sessions. When the variable is not set, the key is always set to a new random string at startup.

stavros-k (approved):

PR #3105, opened 2025-09-07
FileFlows

revenz:
FileFlows App (https://fileflows.com)

Icon

SVG: https://fileflows.com/img/logo.svg
PNG: https://fileflows.com/icon.png

Screenshots

https://fileflows.com/images/webconsole/dashboard.webp
https://fileflows.com/images/webconsole/flow-editor-example.webp
https://fileflows.com/images/reports/processing-summary.webp

### DIFF
--- a/ix-dev/community/fileflows/README.md
+++ b/ix-dev/community/fileflows/README.md
@@ -1,0 +1,3 @@
+# FileFlows
+
+[FileFlows](https://www.fileflows.com/) automates all file types—videos, audio, images, books—with VMAF-optimized AV1 and HEVC transcoding for top-tier quality.
--- a/ix-dev/community/fileflows/app.yaml
+++ b/ix-dev/community/fileflows/app.yaml
@@ -1,0 +1,47 @@
+annotations:
+  min_scale_version: 24.10.2.2
+capabilities:
+- description: FileFlows is able to change file ownership arbitrarily
+  name: CHOWN
+- description: FileFlows is able to bypass permission checks for file operations
+  name: FOWNER
+- description: FileFlows is able to change group ID of processes
+  name: SETGID
+- description: FileFlows is able to change user ID of processes
+  name: SETUID
+categories:
+- media
+changelog_url: https://fileflows.com/docs/versions
+date_added: '2025-09-08'
+description: "FileFlows automates all file types\u2014videos, audio, images, books\u2014\
+  with VMAF-optimized AV1 and HEVC transcoding for top-tier quality."
+home: https://fileflows.com/
+host_mounts: []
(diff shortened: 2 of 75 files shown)


stavros-k:
Hello, it mostly looks great! Nice work!
I'll push couple nits in a few minutes.

In the mean time is it possible to have an env var to set the internal port? 
I see there is an `--api-port` flag, but I'm not sure the entrypoint allows us to pass any additional flags.

Maybe adjust entrypoint like so?

```sh
    if [[ "$mode" == "node" ]]; then
        printf "Launching node as '$user'\n"
        cd /app/Node
        stopLogging
        su -c "/dotnet/dotnet FileFlows.Node.dll --docker true" "$user" &
        dotnet_pid=$!
    else
        printf "Launching server as '$user'\n"
        cd /app/Server
        stopLogging
        su -c "/dotnet/dotnet FileFlows.Server.dll --urls=http://*:5000 --docker --api-port=${API_PORT:-5000}" "$user" &
        dotnet_pid=$!
    fi
```

This is in case a user wants to run it with host network and has something else running in the host.

---

That being said, I think we should have separate apps for node and server (like we have for forgejo/gitea runners).
Thanks!

stavros-k:
If we can have the above with the port would be great and be ready to merge after I adjust the template.
Thanks!

---

I can copy and adjust the app to be able to be used as node after its merged. Unless you want to do that.

revenz:
Not sure where that docker-entrypoint.sh came from, but the one I ship doesn't include a --api-port parameter, and thats not used anywhere.   

Happy if this is split for FileFlows and FileFlows Node.   Thought you'd want a single entry vs double entry.

stavros-k:
> Not sure where that docker-entrypoint.sh came from, but the one I ship doesn't include a --api-port parameter, and thats not used anywhere.

The snippet above is an example of what I'd like the entrypoint to be adjusted to.
Curentlly if you enable **host networking,** it will use port 5000, without a way to change that.
And this can cause port conflicts. Which is why I'm asking if its possible to adjust the entrypoint to allow changing the port.

revenz:
Next version, I've made it use 

```
su -c "/dotnet/dotnet FileFlows.Server.dll --urls=http://*:${FFPORT:-5000} --docker" "$user" &
```

stavros-k:
> Next version, I've made it use
> 
> ```
> su -c "/dotnet/dotnet FileFlows.Server.dll --urls=http://*:${FFPORT:-5000} --docker" "$user" &
> ```

Awesome, once this is available we can just update ix_values.yaml and app.yaml to reflect the new version and we should be good to go!

revenz:
Latest version released today (25.09.2) has the $FFPORT change now.

PR #3116, opened 2025-09-08
Additional Ollama Env. Variables

HubGon:
Hey,

i was testing some env. variables with Ollama and thought that maybe some of them would be interesting for other users too, for an easy installation.
This includes:
Possibility to override default context size,
turn on flash attention and set K/V Cache type,
Override how long models are kept in memory by default and
enable the new Ollama Engine.

https://github.com/ollama/ollama/blob/main/docs/faq.md
In this Link there are some more Variables, but i think those are a bit too deep for an easy installation.

In future maybe one could also look into binding Ollama onto an URL for more security.

Kind regards

### DIFF
--- a/ix-dev/community/ollama/migrations/context_flash_alive_engine_migration
+++ b/ix-dev/community/ollama/migrations/context_flash_alive_engine_migration
@@ -7,13 +7,13 @@ import secrets
 
 
 def migrate(values):
-    values["ollama"]["override_context_length"] = false
+    values["ollama"]["override_context_length"] = False
     values["ollama"]["override_context_length"]["context_length"] = "4096"
-    values["ollama"]["flash_attention"] = false
+    values["ollama"]["flash_attention"] = False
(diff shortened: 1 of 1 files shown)


HubGon:
I don't get a pass for the migration. Maybe you could have look @stavros-k ?

stavros-k:
Hey, so test failing because the file must be executable (exec bit set).
But I'm not sure we should start exposing things like overrides etc.
Users can still use additional env vars to add them.
Also this will break upgrades for users that already use at least one of the variables as additional var.

Just as side note for future "notes" I wouldnt use all those enabled flags.
I would just set the upstream default values as the app defaults as well instead.
(Which also brings another "issue" when exposing all those "niche" options. Whenver upstream defaults change we are being outdated :D)

HubGon:
Ok, got it. Thank you.
I think in this case the PR can be closed. :)

PR #3149, opened 2025-09-11
FileFlows change to latest

revenz:
Changed FileFlows to use latest tag, using a fix tag will eventually break and users will be confused why FileFlows stops working.

### DIFF
--- a/ix-dev/community/fileflows/ix_values.yaml
+++ b/ix-dev/community/fileflows/ix_values.yaml
@@ -1,7 +1,7 @@
 images:
   image:
     repository: revenz/fileflows
-    tag: "25.09"
+    tag: "latest"
 
 consts:
   fileflows_container_name: fileflows


stavros-k:
What you mean eventually break?
Fixed tags are updated daily by a bot.

revenz:
The plugins needed to run dont stay around forever, I keep about 4 months worth of plugins online.   So using a fixed tag like this will cause issues and best to use either :stable or :latest

stavros-k:
Using latest or wildcard tags in general causes another set of problems.
For example FileFlows makes a change that requires docker compose to be updated or even a bug/regression, 
With fixed tag, Bot will open a PR, installation will fail on CI and be caught before reaching end users.

With wildcard tags, users can either get their installations broken without knowning what happened, or new installs are just broken.

Besides that, wildcard tag does not update automatically. 
Once it is pulled on the system it will be there and used until you either remove the image so it re-pulls on next start or you have to manually pull it.

There are already couple of apps that only published wildcard tags and not fixed, and had this issue many times.

PR #3227, opened 2025-09-23
immich: remove deprecated old storage config

stavros-k:


rabotond:
I thought I mention that when trying to upgrade my instance to this version, first of course the immich instance stops, but it never start, so update is somehow broken. As the instance not even trying to start, I dont have any logs from the pods to share.

rabotond:
if I am right we have to migrate our instance and datasets to new ones ? :O Didnt see that coming. is there any official migration guide ? @stavros-k

ngoetti:
@rabotond I guess you have to do the following:

What you essentially have to do is to create a new dataset/directory,
move your data there and then update the configuration by unchecking the 'Use old storage config' option.
And pointing the 'Data Storage' to the new dataset/directory.
For example if you create a new dataset at '/mnt/tank/immich'
In this directory create 2 datasets/directories:
`/mnt/tank/immich/data` and `/mnt/tank/immich/postgres-data`

Now inside the `/mnt/tank/immich/data` directory create the following directories (**NOT** datasets):
`mkdir -p /mnt/tank/immich/data/{upload,thumbs,library,profile,backups,encoded-video}`
Then you have to move your data from the old separate dataset/directories to the new ones.
For example `/mnt/tank/old-immich-data/upload` to `/mnt/tank/immich/data/upload`.

rabotond:
Yeah but how ? Should I ssh into it and copy 800 gbs? It will time out. Or what about db ? It makes backups , how can I restore it ? Or shouldn't I restore just copy ? I see people complain about permission issues that case. It's not so simple for a casual user , would be nice to provide a step by step guide how to do things you know

ngoetti:
> Yeah but how ? Should I ssh into it and copy 800 gbs? It will time out. Or what about db ?

After restoring multiple times from the snapshots and trying a lot of different things I made it work like the following:

1. Stop Immich
2. Rename the video dataset to encoded-video with the command line
3. Open immich app settings and enable the new storage config
4. Select root dataset of all immich child datasets.
5. Save and restart immich.

But this only works if all the immich data folders/datasets are already contained in the same dataset.

rabotond:
> > Yeah but how ? Should I ssh into it and copy 800 gbs? It will time out. Or what about db ?
> 
> After restoring multiple times from the snapshots and trying a lot of different things I made it work like the following:
> 
> 1. Stop Immich
> 2. Rename the video dataset to encoded-video with the command line
> 3. Open immich app settings and enable the new storage config
> 4. Select root dataset of all immich child datasets.
> 5. Save and restart immich.
> 
> But this only works if all the immich data folders/datasets are already contained in the same dataset.

Worked!!, I just had to rename the child datasets to match and gave to a new immich instance the root dataset. Also, as I store thumbs in a ssd too not in hdd root, I added it as a custom mount to /data/thumbs. 

Everything working :)

gozes:
> > Yeah but how ? Should I ssh into it and copy 800 gbs? It will time out. Or what about db ?
> 
> After restoring multiple times from the snapshots and trying a lot of different things I made it work like the following:
> 
>     1. Stop Immich
> 
>     2. Rename the video dataset to encoded-video with the command line
> 
>     3. Open immich app settings and enable the new storage config
> 
>     4. Select root dataset of all immich child datasets.
> 
>     5. Save and restart immich.
> 
> 
> But this only works if all the immich data folders/datasets are already contained in the same dataset.
@ngoetti that didn't work for me. I created the encoded-video folder and got ` ERROR [Microservices:StorageService] Failed to read (/data/encoded-video/.immich): Error: ENOENT: no such file or directory, open '/data/encoded-video/.immich'` when trying to star immich

gozes:
@ngoetti do you happen to have the rename dataset command you use?

gozes:
> > > Yeah but how ? Should I ssh into it and copy 800 gbs? It will time out. Or what about db ?
> > 
> > 
> > After restoring multiple times from the snapshots and trying a lot of different things I made it work like the following:
> > 
> > 1. Stop Immich
> > 2. Rename the video dataset to encoded-video with the command line
> > 3. Open immich app settings and enable the new storage config
> > 4. Select root dataset of all immich child datasets.
> > 5. Save and restart immich.
> > 
> > But this only works if all the immich data folders/datasets are already contained in the same dataset.
> 
> Worked!!, I just had to rename the child datasets to match and gave to a new immich instance the root dataset. Also, as I store thumbs in a ssd too not in hdd root, I added it as a custom mount to /data/thumbs.
> 
> Everything working :)


@rabotond make me wander if one can just keep everything as is, in my case that is a dataset for each one of {upload,thumbs,library,profile,backups,video} under a top `immich` dataset, and just use the new storage config and cutsome mount all of them into `/data` video can just be mounted to `encoded-videos`

JMLucas96:
Is it worth if I follow that steps?
https://docs.immich.app/install/truenas/#updating-the-app
When can I do de 5th step? After or before upgrading immich version to 1.144.1?

image
image

JMLucas96:
> Is it worth if I follow that steps? https://docs.immich.app/install/truenas/#updating-the-app When can I do de 5th step? After or before upgrading immich version to 1.144.1?
> 
> image image

Doing step 5th before upgrading immich version worked. When I confirmed all is working well, then upgrade version.

SavanFlou:
> 2. Rename the video dataset to encoded-video with the command line

Ok but which command line?

rabotond:
> > 2. Rename the video dataset to encoded-video with the command line
> 
> Ok but which command line?
> 
> 

zfs rename pool/old_dataset_name pool/new_dataset_name

SavanFlou:
Thank you for your help, I tried to do it but when I put that in my Linux Shell, I got a error message
"zsh: command not found: zfs"

What did I do wrong?

SavanFlou:
Hello, can someone explain to me exactly where I have to put this command line?
I tried in different tabs but I always get the same answer "zsh: command not found: zfs".

Also "pool/old_dataset/name"= should I write "mnt/myDataset" or just "myDataset", as I can copy it as Path in the Dataset tab in Truenas Sale?

Thanks for your help.

lakeland1990:
> Use old storage config

I deployed Immich v1.124.2 (ix-app 1.7.20) and when I upgrade , the critical **'Use old storage config'** option is not exist at the Immich Web UI Settings.
This option is crucial for users upgrading from some versions to ensure storage configuration compliance (Storage Templates).
**Question:** Has the new storage path logic been automatically enforced or hardcoded within this version, making the UI option unnecessary?
Please confirm the recommended safe upgrade path for users coming from earlier versions , given the missing configuration switch.
I am attempting to upgrade an existing Immich instance deployed via a previous version (like to 1.133) but the  version selection is restricted to **2.1.0**.

Issue #148, opened 2023-04-13
Community request for Traefik


Jarery:

Marimax01:
Hello,

Why I can't see it on TrueNAS ?

![msedge_zOd90A9ufc](https://github.com/user-attachments/assets/9b27d5d0-3b25-48e2-86a3-fe1732e3cd53)

kbhbhk:
Second this!!

tomerh2001:
Any updates?

freundTech:
I've been running traefik as a custom app for a while now. I think the biggest challenge to adding it to the repo would be developing a good default configuration.

I set traefik up to serve selected apps on sub-domains of my domain and with TLS, but this requires having a wildcard domain pointed to your NAS (The default mDNS truenas.local hostname doesn't resolve sub-domains and you can't get valid TLS certificates for it).

I set it up as follows:
Go to `Apps` -> `Discover Apps` -> `Custom App`
#### Image configuration
Repository: `traefik`
Tag: `v3.3` (change this to update in the future)
Pull Policy: `Always pull` (so you get updates even if the tag doesn't change)

####  Container Configuration 
Command: That's where we do the main configuration. Add one entry per command-line argument. Putting multiple into one entry doesn't work.
Add the following arguments:
- `--api.insecure=true` (provide the dashboard without TLS)
- `--providers.docker` (enable the docker provider)
- `--providers.docker.defaultRule=Host(`{{ index .Labels "com.docker.compose.service"}}.example.com`)` (Set the default hostname for discovered services)
- `--certificatesresolvers.letsencrypt.acme.email=your-email@example.com` (contact email for LetsEncrypt certificates)
- `--certificatesresolvers.letsencrypt.acme.storage=data/acme.json` (storage location for certificates)
- `--certificatesresolvers.letsencrypt.acme.dnschallenge=true` (Use DNS challenge. http challenge would require your NAS to be exposed to the public internet)
- `--certificatesresolvers.letsencrypt.acme.dnschallenge.provider=your-dns-provider` (Put your DNS provider here. Also set the ENV variables with your secret. For details see https://doc.traefik.io/traefik/https/acme/#providers) 
- `--providers.docker.exposedByDefault=false` (We only want to expose selected containers)
- `--entryPoints.websecure.address=:443` (Define our entry point port. You might need to change your truenas settings first to free up that port. `System` -> `General Settings` -> `GUI` -> `Web Interface HTTPS Port`).
[...]

Issue #743, opened 2024-10-30
Allow for selecting networks


Renari:
Currently configuration for most of the apps only consist of a port and a checkbox for host networking.

It would be nice if all networks found in `docker network ls` were available as an option instead of just host networking.

For some configurations (macvlan, ipvlan) we would need the ability to specify an IP and a default gateway and a parent interface.

stavros-k:
That is indeed something that I do want to investigate more.

Below are some notes, with some thoughts I had when taking a brief look at that.
Adding them here as notes mainly.

---

Things that won't be able due to docker's limitation:
- `host` networking for apps that have more than 1 container. This breaks docker's dns resolution

Now assuming this:
- application with more than 1 container
- User selects a custom network
Then:
- we should create an additional network, to join the "main" application and its dependencies (eg db).

This would allow to have a stable connection between main container and db container, but still being connected to the custom network user selected.

Challenges on this:
- What happens if the custom network also includes a container that conflicts with the dns of the db?

Possible solutions:
- I think docker provides a way to assign additional dns to a container. We can probably generate a unique address for he db for example.
- The unique address should be idempotent. If it changes each time the compose file is generated, might cause issues on apps that don't react to env variables changing. (eg nextcloud and home assistant need manual change of the db address on a config file).

Renari:
Just wanted to point out that I believe this should support inter-app communications as well, not just networks within the same stack. This is useful for things like reverse proxies, for example we could have a traefik network and join containers on that network so that traefik can act as a reverse proxy for them.
```
networks:
  traefik:
    name: traefik
    external: true
```
I don't recall off the top of my head how the DNS works in this situation, I think you need to specify a static IP when joining networks like this but as stated I'm not 100% it's been a while since I've had to do this.

stavros-k:
Yes the custom network will be between multiple stacks, not single!

tomerh2001:
> That is indeed something that I do want to investigate more.
Any chance there's been an update on this?

I just migrated from plain Ubuntu, and I was hoping to see some configuration show up when I install new apps (for example Jellyfin) after I install a reverse proxy (for example Traefik).

Something like:

1. Installing Traefik
2. Installing Jellyfin:
- During configuration, a "Traefik Labels" (or something similar) will show up
- I could enter the host (`traefik.http.routers.jellyfin.rule`): `jellyfin.domain.tld`
- I could enter the port (`traefik.http.services.my-service.loadbalancer.server.port`): `8086`
- Etc...

[...]

PR #3231, opened 2025-09-23
Add Application: Traefik (but for real this time)

N-Hertstein:
It works. Including the networking.

All thanks to https://github.com/obeone/traefik_network_connector

I've tested http proxying, https proxying & http redirection on both my machine and my TrueNAS server, without touching any docker commands.

Now it is not finished yet, but it is actually functional.

Aims to solve #148

### DIFF
--- a/ix-dev/test/traefik/ix_values.yaml
+++ b/ix-dev/test/traefik/ix_values.yaml
@@ -17,7 +17,6 @@ consts:
     docker:
       enabled: true
       exposed: false
-      proxy_network: ix-traefik_default
   # Static Storage variables
   storage:
     acme: /var/traefik/certs
--- a/ix-dev/test/traefik/questions.yaml
+++ b/ix-dev/test/traefik/questions.yaml
@@ -131,7 +131,7 @@ questions:
           label: Dashboard Port
           schema:
             type: dict
-            show_if: [["webui_enable", "!=", "webui_secure"]]
+            show_if: [["webui_enable", "=", "true"]]
             attrs:
               # Set variable for port binding mode, this wil configure accessibility of the webui to none, inter-container communication and publicly accessible
               # How to refer from Docker-Compose template: values.network.web_port.bind_mode
@@ -191,21 +191,7 @@ questions:
             show_if: [["webui_secure", "=", true]]
             default: "traefik.mydomain.com"
 
-        # Set domain name for secure webui
-        # How to refer from Docker-Compose template: values.network.webui_docker_network
-        - variable: webui_docker_network
-          label: Webui Docker Network
-          description: |
-            Required for proper proxying.</br>
-            Usually consists of ix-$$$_default.</br>
-            Where $$$ is the name you've given the container at the top of this form.</br>
-            Can be confirmed by using 'sudo docker network ls'
-          schema:
-            type: string
-            show_if: [["webui_secure", "=", true]]
-            default: "ix-traefik_default"
-
-        # Set domain name for secure webui
+        # Set certificate challenge for secure webui
         # How to refer from Docker-Compose template: values.network.webui_domain
         - variable: webui_resolver
           label: Webui Certificate challenge
--- a/ix-dev/test/traefik/templates/docker-compose.yaml
+++ b/ix-dev/test/traefik/templates/docker-compose.yaml
@@ -16,7 +16,7 @@
 {% do traefik.environment.add_env("TRAEFIK_LOG_LEVEL", "DEBUG") %}
 {% do traefik.environment.add_env("TRAEFIK_PROVIDERS_DOCKER", values.consts.commands.docker.enabled) %}
 {% do traefik.environment.add_env("TRAEFIK_PROVIDERS_DOCKER_EXPOSEDBYDEFAULT", values.consts.commands.docker.exposed) %}
-{% do traefik.environment.add_env("TRAEFIK_PROVIDERS_DOCKER_NETWORK", values.consts.commands.docker.proxy_network) %}
+{% do traefik.environment.add_env("TRAEFIK_PROVIDERS_DOCKER_NETWORK", "ix-%s_default"|format(values.release_name)) %}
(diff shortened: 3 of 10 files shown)


N-Hertstein:
### Test this thing yourself

To verify that it actually functions really for real this time, use the rendered docker compose:
- Download yaml file: 
  - [traefik-docker-compose.txt](https://github.com/user-attachments/files/22498371/traefik-docker-compose.txt) (I'm not allowed to upload yaml files and the code is long af)
- Copy the yaml code
- Go to your truenas server:
 - Go to apps
 - Go to discover apps
 - Click on the 3 dots in the upper right and click on install via yaml
- Copy the yaml code
- Name the app traefik (incredibly important. things will get confused otherwise. This will not be a problem using the questions)
- Save

Now find a container you wish to proxy and add the following labels:

#### For HTTP Proxying:
```
KEY:    traefik.enable
VALUE:  true
```
```
KEY:    traefik.http.routers.$MYAPPNAME$.rule
VALUE:  Host(`$MY.AMAZING.DOMAIN$`)
```
```
KEY:    traefik.http.routers.$MYAPPNAME$.entrypoints
VALUE:  web
```
```
KEY:    traefik.docker.network
VALUE:  ix-$APPLICATION_NAME$_default
```

#### For HTTPs Proxying:
```
KEY:    traefik.enable
VALUE:  true
```
```
KEY:    traefik.http.routers.$MYAPPNAME$.rule
VALUE:  Host(`$MY.AMAZING.DOMAIN$`)
```
```
KEY:    traefik.http.routers.$MYAPPNAME$.entrypoints
VALUE:  websecure
```
```
KEY:    traefik.http.routers.$MYAPPNAME.tls
VALUE:  true
```
```
KEY:    traefik.docker.network
VALUE:  ix-$APPLICATION_NAME$_default
```

Where:
- `$MYAPPNAME$` needs to be replaced by the name of your app
- `$APPLICATION_NAME$` is replaced by the name you have given the app all the way at the top at the application name variable
- `$MY.AMAZING.DOMAIN$` is replaced by a domain of you choosing


To connect to these sites you can do one of the following things:
- Use Nginx proxy manager to redirect the domain to the correct port (8880 for HTTP, 8443 for HTTPs)
- Make sure there is an external dns record that points to your WAN IP and that port 443 and/or port 80 are redirected by your router to port 8880 and 8443
- Change the exposed ports to port 443 and port 80 (ofcourse make sure your truenas dashboard and other apps are not using these ports) and make sure there is a dns record on your local dns server that refers the domain of choice to traefik.

### And if you have feedback I would love to have some

N-Hertstein:
### Currently TO DO to make sure this thing's gonna work (will update as needed)

- [x] Get http proxying working
- [x] Get https proxying working
- [x] Get Dashboard/api working over http
- [x] Get Dashboard/api working over https
- [x] Test HTTP challenge
- [x] Test TLS challenge
- [x] Test cloudflare dns challenge
- [x] Test custom dns challenge
- [ ] Verify if advanced settings disable all other configuration
- [x] Get variable name for application names as networking and the containername ar based off off this and it would reduce manual configuration

stavros-k:
Perhaps I wasn't clear on the other PR.

Traefik is **blocked** until #743 is completed.

`traefik_network_connector` does not change that.

N-Hertstein:
> Perhaps I wasn't clear on the other PR.
> 
> Traefik is **blocked** until #743 is completed.
> 
> `traefik_network_connector` does not change that.

I understand the networking is a big part of Traefik, however it should "change that".
If I could manually select networks I would still build the app the same way.

There are two ways of proxying: 
- Adding Traefik to the network of the app you wish to proxy (which there is no reason to do manually as it can be done automatically)
- Adding your app to the Traefik Network (potentially a secondary proxy network)
The problem with the latter is that this allows for all containers connected to the Traefik network to access each other, negating the isolation docker is for and potentially causing security concerns. This is ofcourse undesireable.

Adding Traefik to the network of the app(s) you wish to proxy allows only Traefik access to that app and the app to Traefik, creating a smaller security risk. Sure, Traefik is still the single point of failure, but sonarr won't interfere with radarr, where this could happen when using a singular proxy network.

All that would arguably need to happen when networks can be selected is further isolation on a per app basis by adding the web service to a secondary network.

Think nextcloud gets an ix-nextcloud_traefik on top of ix-nextcloud_default network, but postgres, redis & cron would only connect to ix-nextcloud_default

This would minimize Traefiks access to other apps to the bare minimum. This can also be implemented as an extra option without needing to get rid or modify current options

I can also create a current (updateable for selectable networks) questions section (with corresponding docker-compose) template for any app to have something as simple as:
- Proxy trough Traefik:
  - Enable
  - Domain name
  - Protocol:
    - HTTP
    - HTTPs
  - Certresolver
Any advanced configuration can (and should, because there's so  many options) already be done trough labels.


Blocking a functional app that requires no added functionality seems a bit unnecessary.

At least **think** about putting it in the test or dev branch (maybe launch an alpha or beta branch) for those who really want it and promote it to the community branch later, when it is proven to be stable and work and/or when #743 gets solved.

I can't force you to accept an application. But I am more than willing to maintain and migrate this when necessary.
Hell I'll put my name and e-mail in the app.yaml if needed

stavros-k:
There are alternatives until blocker is gone/completed. Custom app.

Yes there are a lot of ways to make it work. That does not mean that "hacks" or ways that cannot be validated for correctness will make it into the catalog.

I'm not sure in what other words I can put it. This is blocked. Users who want to install traefik, can via custom options today.

N-Hertstein:
> There are alternatives until blocker is gone/completed. Custom app.
> 
> Yes there are a lot of ways to make it work. That does not mean that "hacks" or ways that cannot be validated for correctness will make it into the catalog.
> 
> I'm not sure in what other words I can put it. This is blocked. Users who want to install traefik, can via custom options today.

Well it's reasonable that we'd want to verify that labels are properly read and "docker network connect" commands are issued correctly. I'll take some time looking over the two python files in traefik_network_connect that make up the logic and see if they conform to the docker sdk python library's documentation.

PR #3253, opened 2025-09-25
radarr: add WebUI SSL options

jcassette:
Hello

Can you consider integrating this change please?

This allows to configure the Radarr WebUI as HTTPS.

N.B: The PKCS 12 file is not provisioned automatically. The user must provide it separately, e.g. via a host mount.

### DIFF
--- a/ix-dev/community/radarr/questions.yaml
+++ b/ix-dev/community/radarr/questions.yaml
@@ -131,6 +131,26 @@ questions:
                         required: true
                         $ref:
                           - definitions/node_bind_ip
+              - variable: ssl_enable
+                label: Enable SSL
+                description: Enable SSL for the WebUI.
+                schema:
+                  type: boolean
+                  default: false
+              - variable: ssl_cert_path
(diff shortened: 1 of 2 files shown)


stavros-k:
Hello, I will only consider it if it can accept the cert using the certificates from the truenas system (Like we do on other apps).

---

From a quick look it accepts only pfx (ie no PEM).
If you want to make this setup, you can use a custom app and mount certs from your host.

Thanks

jcassette:
How about just setting the `RADARR__SERVER__SSLPORT` environment variable instead of `RADARR__SERVER__PORT` if a checkbox is ticked?

stavros-k:
That wouldnt be a complete setup either. Would be? it will still require user to provide pfx.

jcassette:
Yes that would not be a complete setup.
Maybe just allow to expose another port for SSL?
Like it is already done for others apps, e.g. emby, jellyfin, ...
They do not provide a complete setup either.

PR #3390, opened 2025-10-14
tdarr-node: allows nfs/smb on transcode

### LINKED ISSUE #3372: [Enhancement]: TDarr Node: Add NFS Share to Options for Volume Type for Transcode Storage
### App Name

TDarr Node

1.0.8

### Enhancement Description

I would like to be able to select NFS share as a volume type when configuring the transcode storage on the app.  Mapped nodes need to be able to communicate with the server's filesystem.  I have included a link to the relevant documentation below.

https://docs.tdarr.io/docs/nodes/nodes

### Use Case

 Enabling this option would make it easier to deploy the Tdarr Node on a different system than the server.

### Proposed Solution

Add NFS Share as an option to this list.

Image

### Alternatives Considered

I have gotten around it by mounting the NFS share to a local directory and using a host path mount.  Setting this up is cumbersome as I have to have a script run to mount the share each time the host reboots.

### Additional Context

_No response_

stavros-k:
Closes #3372

### DIFF
--- a/ix-dev/community/tdarr-node/app.yaml
+++ b/ix-dev/community/tdarr-node/app.yaml
@@ -38,4 +38,4 @@ sources:
 - https://docs.tdarr.io/docs
 title: Tdarr Node
 train: community
-version: 1.0.8
+version: 1.0.9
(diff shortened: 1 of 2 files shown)


WilliamsAve:
I appreciate the quick turnaround on this.  I've tested it out and it seems that when selecting the NFS Share option, it does not present the options to map the share.  Please see the below screenshot.

image

Interestingly, when selecting the hostpath, there are options to configure SMB and NFS.

image

stavros-k:
ugh, thanks for letting me know

WilliamsAve:
It works beautifully.  Thank you for the quick response!

PR #3480, opened 2025-10-28
authentik: update and remove redis (adapts to upstream changes)

stavros-k:


ix-dev/community/authentik/ix_values.yaml
stavros-k:
```suggestion
    tag: "2025.10.0"
```

PR #3569, opened 2025-11-08
Update source URL for ViTransfer in app_versions.json

MansiVisuals:
Using the correct https://github.com/MansiVisuals/ViTransfer instead of https://github.com/crypt010/vitransfer. 
crypt010 is only for the docker images on docker.io


stavros-k (changes requested):
https://github.com/truenas/apps/blob/eda150d6ef92dc56e645a7e6bdcdb4b5cf8c7173/CONTRIBUTIONS.md?plain=1#L24-L25

PR #3606, opened 2025-11-13
qbit: add deprecation yaml 

stavros-k:


ix-dev/community/qbittorrent/deprecations.yaml
```diff
@@ -0,0 +1,13 @@
+deprecations:
```
sonicaj: Can we have more keys here ?
If this one is going to be fixed, i think we should not have this key at all ?

stavros-k:
```suggestion
```

PR #3608, opened 2025-11-13
Channels-DVR Add Nvidia Image Option

zchrykng:
Channels offers a docker image that works with NVIDIA GPUs, this adds that image as an option for the app.


zchrykng:
@stavros-k not sure why that check is failing. Something to do with the NVIDIA drivers, etc?

zchrykng:
Okay, actually looks like the nvidia image is missing the `cn` command. Created a thread over on their forum.

zchrykng:
Thanks!

PR #3630, opened 2025-11-16
Add Donetick

daaain:
This is my first app so winging it a bit 😅 sorry if I misunderstood something and the structure isn't right!

Managed to get it running with:

```sh
./.github/scripts/ci.py --app donetick --train community --test-file basic-values.yaml --wait=true
```

Loads correctly:

Screenshot 2025-11-16 at 00 52 31

Could also sign up and create a task that persisted after page reload:

image


daaain:
Realised the previous implementation could be improved by implementing Additional Environment Variables list instead of specific JWT Secret so this way people can customise any aspect of the app: https://docs.donetick.com/getting-started/configration#configuration-via-environment-variables

daaain:
Was also wondering if we should use SQLite for simplicity or Postgres for scaling? How do you usually decide between these options? Or maybe you have a pattern for making these options as conditional templates? Or would that be too much maintenance with the different permutations multiplying?

stavros-k:
Hello, I'm sorry but this is completely out of standard with all other apps.
Making it impossible to add on the catalog and maintain.
I'm not sure why you'd do that especially when its your first app.

You also assumed a lot of users preferences like host paths.
Removed completely any resource, label, and other options.

Healthcheck didnt pass either :)

stavros-k:
Blocked by https://github.com/donetick/donetick/pull/420
Tests won't pass until linked PR is merged and a new release is made.

daaain:
> Hello, I'm sorry but this is completely out of standard with all other apps. Making it impossible to add on the catalog and maintain. I'm not sure why you'd do that especially when its your first app.
> 
> You also assumed a lot of users preferences like host paths. Removed completely any resource, label, and other options.
> 
> Healthcheck didnt pass either :)

Sorry about that, I did check `apps/CONTRIBUTIONS.md` but it doesn't have any info about these conventions, so I relied on the Python code doing the validation and making the container work.

If you could add some more info (and maybe links to app examples) to make it easier to learn these patterns than randomly clicking around into other apps' directories, that would be super useful for new contributors!

Anyway, thanks for picking it up!

stavros-k:
Will be merged when donetick releases the next version (out of beta)

PR #3668, opened 2025-11-20
Bump immich app version to v2.3.0

LeLunZ:
There is currently a bug in immich which makes using the old webapp not possible when a new minor version is released. 
A fix for this will come in a future version, but for now updating is the only "fix".

I am not sure if I can get this to trigger a workflow or somehow speed up the process of releasing a new version, but this is my try of exactly doing that :)

### DIFF
--- a/ix-dev/community/immich/app.yaml
+++ b/ix-dev/community/immich/app.yaml
@@ -43,4 +43,4 @@ sources:
 - https://github.com/immich-app/immich
 title: Immich
 train: community
-version: 1.10.12
+version: 1.10.13


LeLunZ:
It seems the ghcr.io/immich-app/immich-machine-learning rocm tag for 2.3.0 is still missing.

ghazyami:
More context
https://github.com/immich-app/immich/issues/24009#issuecomment-3555408681

And 2.3.1 is already out
https://github.com/immich-app/immich/releases/tag/v2.3.1

stavros-k:
As always updates happen automatically. Please be patient!

LeLunZ:
@stavros-k It's just that the app is currently broken and an update would fix it. 
So there is no way to manually start the truenas app pipeline to update an app?

What exactly are the requirements for the automatic update to happen? 
Is a requirement for the automatic run that all immich containers (server, ml...) have an image with the latest version? (Because the ml-rocm image does seem to have a problem and won't build for the latest v2.3.0/release tag)

---

also it seems like the automatic update already does it job: #3670 
But I am still interested, what exactly is needed so the automatic update is run :) I didn't find any documentation about that anywhere

stavros-k:
It runs on a schedule daily. It picks up all available updated tags. Nothing more or less :)

PR #3772, opened 2025-12-03
Enhanced contribution guide for TrueNAS Apps catalog

attilagyorffy:
## Overview

This PR significantly enhances the contribution guide to provide a comprehensive, hopefully well-structured resource for both new and experienced contributors to the TrueNAS Apps Catalog.

## My motivation

I have tried to contribute to the catalog myself and even well-intentioned fixes and updates seem to have been  incomplete or out of character, leading to duplicated effort and frustration for both maintainers and potential contributor.

A comprehensive contribution guide isn't just documentation, it's an investment in the project's future. When contributors have clear guidance, they spend less time guessing and more time building. When maintainers can point to documented patterns, reviews move faster. The result is a virtuous cycle: better apps, happier users, more contributors, and a thriving ecosystem.

This is my attempt to transform contributing from a daunting task into a well-lit path forward.

### What's Covered

The updated guide now includes:

- **Complete Project Context**: Explains TrueNAS Apps architecture and mission, helping contributors understand the "why" behind the system
- **Detailed Step-by-Step Tutorials**: Full walkthroughs from cloning the repository to submitting a PR, with real code examples for every file type
- **Architecture Deep Dive**: Top-down explanation of how the library system, rendering process, and CI/CD pipeline work together
- **Configuration System Reference**: Comprehensive documentation of all schema types (string, int, boolean, dict, list, enums) with examples
- **Template API Guide**: Complete library function reference for building Docker Compose templates
- **Testing Workflows**: Clear instructions for local testing and troubleshooting common issues
- **Migrations Guide**: When and how to create migrations for breaking changes
- **Best Practices**: Security, performance, and code quality guidelines with examples
- **PR Submission Template**: Structured checklist to ensure quality submissions

### Benefits

**For new contributors:**
- Lower barrier to entry (no prior knowledge assumed)
- Clear examples reduce confusion and trial-and-error
- Reduces time from "interested" to "first PR submitted"

**For more experienced contributors:**
- Quick reference for schema types, template functions, and patterns
- Migration guides prevent data loss during updates
- Best practices ensure consistent, high-quality submissions

**For maintainers:**
- PRs should arrive more complete and well-structured
- Fewer back-and-forth review cycles
- Common questions answered in advance, reducing support burden
- Standardized approach makes reviews faster and more consistent

**For the Catalog:**
- Higher quality applications with better security and UX
- More contributors = faster catalog growth
- Consistent patterns make maintenance easier across all apps
- Better documentation means apps are easier to debug and update

---

[...]

Issue #2224, opened 2025-05-07
App request: Ghost Blog


razvanphp:
I might push a PR if all is good on local test.

Ghost is an independent platform for publishing online by web and email newsletter. It has user signups, gated access and subscription payments built-in (with Stripe) to allow you to build a direct relationship with your audience. It's fast, user-friendly, and runs on Node.js & MySQL8.

https://hub.docker.com/_/ghost
https://ghost.org/

Thank you!

stavros-k:
> & MySQL8.

If it cannot run with MariaDB, it won't be accepted right now. Not until a MySQL dependency is available in the render library

razvanphp:
What do you mean by render library? Can we also contribute there with some dev power and testing?

I think it should run with a local sqlite file for development purposes, maybe we can add this first.

stavros-k:
> What do you mean by render library?

https://github.com/truenas/apps/tree/master/library

> Can we also contribute there with some dev power and testing?

Yes, but currently I'm not going to accept mysql, as I'm not looking into maintaining it and its upgrades in the future.
99.99% of the apps that use mysql are compatible with mariadb as well, which is already added in the render library.
Last time I checked Ghost uses some very specific mysql features that dont work on mariadb.

> I think it should run with a local sqlite file for development purposes, maybe we can add this first.

If its about development purposes, then I think custom app might be more appropriate.

stavros-k:
Given that ghost won't support any other DB anytime soon, I'll close this one.
if at any point is decided to support mysql, I'll revisit this and reopen

attilagyorffy:
I have taken a stab at adding Ghost using an inline MySQL container (first one of its kind in the Apps Catalog afaik)  and I'd love some feedback/testing pls: https://github.com/truenas/apps/pull/3783

PR #3783, opened 2025-12-04
Add Ghost to TrueNAS Apps Catalog

attilagyorffy:
Adds Ghost 6.9.3, a powerful open-source publishing platform for professional content creators.

### Features

- Ghost 6.9.3 (latest stable) with ~~MariaDB 12.1.2 backend~~ an inline MySQL container
- Dual portals: main site (/) and admin panel (/ghost)
- Persistent storage for content (themes, images, files) and database
- Non-root execution (Ghost: UID 1000, MariaDB: UID 999)

### Implementation notes

This implementation was done via a manually assisted AI-driven session, heavily based on the [official Ghost Docker guide](https://docs.ghost.org/install/docker) and the new, [enhanced contribution guide](https://github.com/truenas/apps/pull/3772).

- Follows WordPress app pattern (MariaDB + CMS)
- ~~Uses `tpl.deps.mariadb` helper from library v2.1.66~~ → Uses a separate MySQL container
- Permission container with check mode
- Excludes optional features per requirements: Tinybird analytics, ADMIN_DOMAIN, Caddy, ActivityPub

### Configuration

- Site URL configuration (required)
- Database credentials (auto-generated)
- Port binding (default: 2368)
- Storage: ix_volume or host_path support
- Additional environment variables for advanced features

### Testing

- ✅ CI script validation passed (render + deployment)
- ✅ All containers start successfully and become healthy
- ✅ MariaDB healthcheck working correctly
- ✅ Ghost connects to database and runs migrations automatically
- ✅ Web UI and admin panel accessible
- ✅ Manual Docker Compose testing completed

### DIFF
--- a/ix-dev/community/ghost/README.md
+++ b/ix-dev/community/ghost/README.md
@@ -26,7 +26,7 @@ Examples:
 
 ### Database
 
-Ghost uses MariaDB for data storage. The database is automatically configured with the credentials you provide during setup.
+Ghost uses MySQL for data storage. The database is automatically configured with the credentials you provide during setup.
 
 - Database passwords are required and should be strong
 - The database is stored in a separate volume for persistence
@@ -37,7 +37,7 @@ Ghost uses MariaDB for data storage. The database is automatically configured wi
 Ghost requires persistent storage for:
 
 - **Content Storage**: Themes, images, uploaded files, and application data
-- **Database Storage**: MariaDB data files
+- **Database Storage**: MySQL data files
 
 Both storage volumes are essential for Ghost operation and should be backed up regularly.
 
@@ -67,7 +67,7 @@ Refer to [Ghost Configuration Documentation](https://ghost.org/docs/config/) for
 ## Notes
 
 - Ghost runs as non-root user (UID 1000)
-- MariaDB runs as non-root user (UID 999)
+- MySQL runs as non-root user (UID 999)
 - Initial setup requires web UI access
 - URL changes require app restart
 - Content uploads are stored in the content volume


attilagyorffy:
WIP: Need to fix [the CI in the Github workflows](https://github.com/truenas/apps/actions/runs/19940671440/job/57177372557?pr=3783), marking the PR as a draft in the meanwhile.

attilagyorffy:
TODO: Find a temporary placeholder icon/screenshot so that it won't fail the CI until maintainers can upload these to the CDN.

Note: Official logos can be found here: https://docs.ghost.org/logos

attilagyorffy:
@stavros-k I would love some feedback on the migration from MariaDb done in 1d0dd8bf:

- Replaced MariaDB helper with inline MySQL 8.4.7 container
- First catalog app to use native MySQL (others use MariaDB)
- Uses MySQL 8.4.7 LTS (stable version officially supported by Ghost)
- Uses an inline MySQL config (required approach to avoid adding MySQL to base library as another dependency)
- Uses standard permissions (`999:999` for MySQL, `1000:1000` for Ghost)
- Added a custom healthcheck

stavros-k:
This comment still stands. https://github.com/truenas/apps/issues/2224#issuecomment-2862241004

But I'd like to add that supporting mysql is not something I'm looking to do currently.

attilagyorffy:
> This comment still stands. [#2224 (comment)](https://github.com/truenas/apps/issues/2224#issuecomment-2862241004)
> 
> But I'd like to add that supporting mysql is not something I'm looking to do currently.

Thanks for the feedback, @stavros-k, much appreciated.

I understand the hesitation around MariaDB (you were right), that’s why I migrated the implementation to MySQL 8. We are aligned.

I want to clarify that the current app definition uses a MySQL container encapsulated inside "the app" itself. It functions as an internal service required for Ghost to run, **not as a standalone database offering to the wider catalog**. In that sense it’s no different from other multi-service apps already present, and it aligns with [Ghost’s official Docker setup](https://docs.ghost.org/install/docker).

With that in mind, I’d really like to understand what remaining concerns or uncertainties prevent this approach from being acceptable. If there are specific risks, policy constraints, or technical blockers, I’m more than happy to address them so we can find a path that brings Ghost into the catalog in a way that fits your expectations and without the unnecessary maintainability overhead.

Thanks again for taking the time to review this, looking forward to your thoughts.

stavros-k:
Encapsulating it into the app like "other" services, ie workers etc, doesn't make it more maintainable.
Making migration to library dependency later hard. The moment this is accepted there are a bunch of other apps that support only MySQL and people will rush in to add them or request them. And you can't say yes to one and no to the other.

Having mysql as dependency involves research on future paths, like upgrades on major versions etc. So design have to be solid. For example postgres had a wrong setup on the data paths, which I had to spent a week crafting migrations around that. I definitely don't want to be in the same situation with something like mysql.

Note that mysql is not "just another internal service". It is not controlled by the app author and cannot provide automatic migrations or backwards compatibility (the author). Its **external** service.

stavros-k:
Closing this for now. Once there is a plan to add mysql dep, We can revisit.

Thanks

PR #3825, opened 2025-12-10
Add MAKEMKV_KEY UI question to makemkvp app

williamhatcher:
Added a question to enable adding a makemkv license key if you have one.


williamhatcher:
Do variable names need to be lowercase?

stavros-k:
Yes.

PR #3869, opened 2025-12-15
opencloud: add calendar support using `radicale`

wojcik-online:
Add calendar support to the existing `opencloud` app by following the template provided in https://github.com/opencloud-eu/opencloud-compose?tab=readme-ov-file#with-radicale

[Radicale](https://radicale.org) has been chosen as the official way to provide a calendar for opencloud, see [radicale integration docs](https://docs.opencloud.eu/docs/admin/configuration/radicale-integration).

### DIFF
--- a/ix-dev/community/opencloud/templates/macros/radicale_config.ini
+++ b/ix-dev/community/opencloud/templates/macros/radicale_config.ini
@@ -21,5 +21,5 @@ type = none
 
 [storage]
 filesystem_folder = {{ values.consts.radicale_data_path }}/collections
-predefined_collections = {{ collections | to_json }}
+predefined_collections = {{ collections | tojson }}
 {% endmacro %}


stavros-k (commented):
Hello, The issue with this integration is that it takes full control over the proxy.yaml file.
1. User cannot add its own config.
2. Will overwrite existing configs.

Besides that, why do we need https for radical if its only meant to be accesed by opencloud within the docker network? I mean its exposed. But given the drawback of taking full control of the proxy.yaml file I'm not sure I can accept this integration.

ix-dev/community/opencloud/templates/docker-compose.yaml
```diff
@@ -79,6 +98,9 @@
   {% set cert = values.ix_certificates[values.network.certificate_id] %}
   {% do c1.configs.add("private", cert.privatekey, values.consts.ssl_key_path) %}
   {% do c1.configs.add("public", cert.certificate, values.consts.ssl_cert_path) %}
+
+  {% do radicale.configs.add("radicale_private", cert.privatekey, values.consts.radicale.ssl_key_path) %}
+  {% do radicale.configs.add("radicale_public", cert.certificate, values.consts.radicale.ssl_cert_path) %}
```
stavros-k: this will fail when radicale is not enabled.

wojcik-online:
> User cannot add its own config.

Added field to `questions.yaml` in which a user can provide their own custom proxy config.


> why do we need https for radical if its only meant to be accesed by opencloud within the docker network? 

Simplified the code, to always connect to radicale via http.


> Will overwrite existing [proxy.yaml file].

The file will be overwritten, which will be the same behavior as with `csp.yaml`.
The user will have to provide their custom config via app installation/edit page, same as with csp.yaml.

Ideally, an existing proxy.yaml file would not be overwritten, and only provided if the user has not created their own, but I did not find such an option.

Unfortunately, modifying proxy.yaml is necessary.
- > Due to the complexity when defining routes, these can only be defined in the yaml file but not via environment variables.
  > -- https://docs.opencloud.eu/docs/dev/server/services/proxy/information

stavros-k:
> Added field to `questions.yaml` in which a user can provide their own custom proxy config.

Unfortunately having a string which is single line wont work.
And even if it does using `\n` etc, its just going to cause more issues.

I think we either take full control of it, or user has to manually configure external radicale.
That being said, what is the difference of radicale image of opencloud vs the official one?

wojcik-online:
> user has to manually configure external radicale

In this case, the user would have to provide their own `proxy.yaml`, which IMO goes against the idea of easy to install apps.

Especially considering that radicale is officially supported by OpenCloud, and it complains without it:
image


> we either take full control of it, or 

Implemented a double opt-in.
First the user enables calendar support, which adds and configures the `radicale` container.
Then separately, the user opts in to take full control of `proxy.yaml`.

Most users who just want it to work, should select both checkboxes, but if anyone want to have their own custom proxy configuration, they can leave the second checkbox false.
(The assumption is such a user is savvy enough to manually configure radicale)


> what is the difference of radicale image of opencloud vs the official one?

The dockerfile of opencloud is here: [opencloud-eu/container-radicale](https://github.com/opencloud-eu/container-radicale)  
The only difference compared to the official one, that I can see, is fixed dependency versions.
My guess is that OpenCloud will provide vetted / for sure compatible version updates.

itn3rd77:
@stavros-k Would really love to see this working. My feeling is this is just the start and the calendar will get a tighter integration into OpenCloud in the next releases. As you can read at https://docs.opencloud.eu/docs/admin/configuration/radicale-integration/ there are already some benefits from this solution like automatically create default calendar and address book for every authenticated user.

stavros-k:
I'm still skeptical of this.
There are things that are only configurable via proxy.yaml (ie https://github.com/opencloud-eu/opencloud/blob/main/services/proxy/README.md#automatic-quota-assignments) or Role Mapping.

I guess we will figure it out when the need arises.

stavros-k (approved):

PR #3928, opened 2025-12-22
Authentik storage

stavros-k:
Blocked until 2025.12.0 is released.
---

"Media" has been moved under a unified "Data" storage by upstream.

## New Installs:

The "media" storage is "disabled" and will result in an exception if a user tries to use it. This avoids doing a fresh install with a soon-to-be-removed setup

## Existing installs:

During upgrade the "Use Deprecated Media" flag will be set, which will do a "double" mount.
This will continue to "work" as it should.
"Data" under `/data`
"Media" under `/data/media`.

A deprecated notice will be shown in the Notes card and on the app wizard.

### Migration steps:

1. Stop the app.
2. Create a directory for Data storage whereever you want ton the Host. (ie `/mnt/tank/apps/authentik/data`)
3. Go to the path where Media storage is mounted on.  (ie `/mnt/tank/apps/authentik/media`)
4. Move any contents from `/mnt/tank/apps/authentik/media` to `/mnt/tank/apps/authentik/data/media` (ie if there is a file like `/mnt/tank/apps/authentik/media/my-file.png` it should end up being `/mnt/tank/apps/authentik/data/media/my-file.png`)
5. Edit the Authentik App.
6. Change the type of the Data storage to host path and point to `/mnt/tank/apps/authentik/data`
7. Uncheck the `Use Deprecated Media Storage` checkbox. (Media storage should no longer show)
8. Save
9. Done.


zaid-marji:
- What happens if user deselects "Use Deprecated Media Storage" without moving the media directory? Data loss?
- Will this be removed in the future? And if yes, will media be moved automatically?

stavros-k:
> What happens if user deselects "Use Deprecated Media Storage" without moving the media directory? Data loss?

It will just not work correctly. Data are still in the host path you used or in the hidden dataset if its ix volume. 


> Will this be removed in the future? 

Yes

> And if yes, will media be moved automatically?

No.

PR #3958, opened 2025-12-28
portainer: run rootless

stavros-k:
Not sure if it was always the case, but it can definitely run rootless, as I use it as rootless for a while now.
Migration will set the run user to 0 for existing installs. A deprecation is added and rootful setup will be removed later.


meijieru:
Portainer installed before cannot start now.

PR #4070, opened 2026-01-15
Stirling PDF - Rename environment variable for security settings

Chaphasilor:
- The previous value - although shown in the error message when trying to connect from the desktop app - seems to be outdated, and has no effect
- Docs now mention the new value, and that actually enables the login for the app: https://docs.stirlingpdf.com/Installation/Docker%20Install

I'm not sure if this is the only change needed, but I thought doing this as a PR would help you more than a simple issue. Feel free to close if it's not helpful!


stavros-k:
I dont think its a replacement, but rather an addition.
The upstream docs have both
https://github.com/Stirling-Tools/Stirling-PDF/blob/main/docker/unified/README.md#security-considerations

Chaphasilor:
So I guess I should add a new question/checkbox for this then, keeping the existing one for "Enable Security" and a new one for "Enable Login"?  
Or should both env vars be controlled by a single checkbox?

stavros-k:
What does exactly each do? Is it actually two different things or just outdated docs?

stavros-k:
Any updates?

Issue #3937, opened 2025-12-23
[Enhancement]: Tailscale Serve Setup Options


Shrinks99:
### App Name

Tailscale

### Current Version (TrueNAS Catalog App Version)

1.3.26

### Enhancement Description

[Tailscale Serve](https://tailscale.com/kb/1242/tailscale-serve) is an easy way of exposing _specific_ services to your Tailscale network.  When used in conjunction with [Tailscale Services](https://tailscale.com/kb/1552/tailscale-services), one can easily expose a service running on any port to a subdomain on their Tailscale network `subdomain.example-tailnet.ts.net` Neat!

Unfortunately right now within the Talescale app in TrueNAS, there is no way to run `tailscale serve` commands on container launch.  We can add `tailscale set` commands with the "extra arguments" field[^1], set [Tailscale Daemon flags](https://tailscale.com/kb/1278/tailscaled?q=daemon#flags-to-tailscaled), and set environment variables.  All helpful, but none of them allow me to run multiple `tailscale serve` commands in the background.

[^1]: Should probably also rename this to "Tailscale set flags" or similar as all arguments run under `tailscale set`??

### Use Case

I'm trying to provision certs for the various services on my NAS, most notably Nextcloud which requires it for its sync client despite the fact that all my NAS services are behind Tailscale (an encrypted VPN).  Despite this, `tailscale serve` should make it very easy to provision certs, as this is automatically done for the user for each service they have configured.

### Proposed Solution

It would be really handy to expose both `tailscale serve` in the app setup UI?  Being able to assign ports to `tailscale serve` along with a service name (if configured) with `--service=svc:SERVICENAME` would be a super easy way to provision certs and configure subdomains for individual services!

### Alternatives Considered

There are good guides out there for [using Tailscale with Nginx Proxy Manager](https://www.youtube.com/watch?v=ezr-kNLfd7A) or using the `TS_SERVE_CONFIG` environment variable to point to a dataset containing a json file instead of using CLI flags.  Generally I like to keep my apps as self-contained and stock-configured as possible.  Adding these flags to the stock Tailscale config would give me the piece of mind to know that they'll likely continue to work across versions, or that a migration path will be documented somewhere.

### Additional Context

[`tailscale cert`](https://tailscale.com/kb/1153/enabling-https) could also be nice, though the command it isn't really relevant as part of what I'm trying to do.  

[...]

PR #4104, opened 2026-01-19
Adds Tailscale Serve Support

Shrinks99:
Closes https://github.com/truenas/apps/issues/3937

Full disclosure, much of this was generated with Claude Code.  I prompted it to review the contributing guide (which is well documented!) and draft a plan to add the Tailscale Serve config using a config file based on the Tailscale docs.  I then edited the plan to guide it to add the serve config as a list with controls that map to the [Tailscale services config file](https://tailscale.com/kb/1589/tailscale-services-configuration-file?q=docker).

I tested this with the recommended `./.github/scripts/ci.py --app myapp --train community --test-file basic-values.yaml --render-only=true` command, using `render-only` because it didn't work without it (due to unmapped volumes which I figure will work as expected on TrueNAS?  I use the default ixvolume for Tailscale, but whatever the user sets _should_ work.)

I hope this doesn't seem like a poorly thought out LLM one-shot, but also I have limited ways of testing this as I can't deploy it to my TrueNAS instance easily?  I really wish there was a better way to load these forms into the server so I could test it more thoroughly before sending a patch!  In any case, I have read through the code, and I hope that you don't think I'm wasting your time here.

## Changes
- Added Tailscale serve config to the setup form
- Changed the label for "Extra Arguments" to "Tailscale Set Arguments" to better describe what arguments are relevant


stavros-k:
I wonder why we should do that if user can just mount that json file themselfs?

ix-dev/community/tailscale/templates/docker-compose.yaml
```diff
@@ -46,6 +46,62 @@
...
+  {% for svc in values.tailscale.serve.services %}
+    {% set svc_name = svc.service_name %}
+    {% if svc_name not in services_by_name %}
+      {% do services_by_name.update({svc_name: []}) %}
+    {% endif %}
+    {% do services_by_name[svc_name].append(svc) %}
+  {% endfor %}
+
+  {# Build config for each service group #}
+  {% set serve_json = {"Services": {}} %}
```
stavros-k: based on linked docs this is lowercased in the docs?
And version key is missing?

ix-dev/community/tailscale/templates/docker-compose.yaml
```diff
@@ -46,6 +46,62 @@
...
+    {% set tcp_config = {} %}
+    {% set web_config = {} %}
+    {% set funnel_config = {} %}
+
+    {% for svc in svcs %}
+      {% if svc.protocol == "https" %}
+        {# HTTPS Proxy mode - TLS termination with HTTP proxy #}
+        {% do tcp_config.update({svc.port|string: {"HTTPS": true}}) %}
+
+        {% set domain_key = "${TS_CERT_DOMAIN}:%d"|format(svc.port) %}
```
stavros-k, line 77: what is this env var?

ix-dev/community/tailscale/templates/docker-compose.yaml
```diff
@@ -46,6 +46,62 @@
...
+
+  {% for svc_name, svcs in services_by_name.items() %}
+    {% set tcp_config = {} %}
+    {% set web_config = {} %}
+    {% set funnel_config = {} %}
+
+    {% for svc in svcs %}
+      {% if svc.protocol == "https" %}
+        {# HTTPS Proxy mode - TLS termination with HTTP proxy #}
+        {% do tcp_config.update({svc.port|string: {"HTTPS": true}}) %}
```
stavros-k, line 75: I cant find HTTPS key in the docs?

ix-dev/community/tailscale/templates/docker-compose.yaml
```diff
@@ -46,6 +46,62 @@
...
+
+        {% set domain_key = "${TS_CERT_DOMAIN}:%d"|format(svc.port) %}
+        {% if domain_key not in web_config %}
+          {% do web_config.update({domain_key: {"Handlers": {}}}) %}
+        {% endif %}
+        {% do web_config[domain_key]["Handlers"].update({svc.path_prefix: {"Proxy": svc.proxy_target}}) %}
```
stavros-k, line 81: proxy and handlers are not in the config file structure either.
I only found a reference in the k8s docs.

stavros-k:
You can only test on a local machine, not on truenas. CI script is wrote for non-truenas systems.

ix-dev/community/tailscale/templates/docker-compose.yaml
Shrinks99: This is used for the [services config file](https://tailscale.com/kb/1589/tailscale-services-configuration-file) as the `<protocol>`

ix-dev/community/tailscale/templates/docker-compose.yaml
Shrinks99: I've succumbed to the slop :(

I'll do a more thourough pass at this today, serves me right for submitting it at 3 AM.  LLMs seem pretty biased to still thinking of Truenas as a k8s system.

ix-dev/community/tailscale/templates/docker-compose.yaml
stavros-k: Can you point me where is the HTTPS key ?

Shrinks99: > The `<local-target>` defines where to forward traffic on the Service host and uses a URI-style syntax in the format `<protocol>://<host:port>`, where `<protocol>` is one of the following:
>
>    - http
>    - https
>    - tcp

Unless I'm missing something, this is what I was trying to support

Maybe also relevant to your question about users just providing the JSON file themselves, I wanted to do it this way because I figured the UI would be more user-friendly, though I'd be open to having them author the JSON in the setup form?  Personally I find this preferable because I don't have to set up a dataset but both options have their benefits.  Using the JSON alone would likely help with maintnance as you wouldn't really have to update anything if Tailscale changes stuff.

stavros-k: Thats all lowercase, and are on a map of string: string.
I dont see HTTPS: boolean anywhere.

ix-dev/community/tailscale/templates/docker-compose.yaml
stavros-k: Sadly, LLMs are no good for whats happening on this repo.
There is nothing in their training set to work with.

Shrinks99: Yeah sorry if this is wasting your time, though in fairness to the robot I don't think my first pass would have had fewer errors, only different ones?  I do appreciate having a first-pass to work off of in this case... But it seems like it's time to fix these for real.  In any case, both the machine and your comments have helped me understand this codebase better and I appreciate you taking the time.

Also should have marked as a draft.

ix-dev/community/tailscale/templates/docker-compose.yaml
Shrinks99: The version is missing because the robot is sloppy and missed it despite being required, I didn't review the JSON config well enough to catch it.

Casing is fixed, version has been hard coded as `0.0.1`.

stavros-k:
> I would also assume that a different command would need to be used to start the services (which will no longer start the normal tailscale node when this command is set).

I'd be more focused on checking this first.
if the command to start the services overrides the other "vpn" features, then I'm not sure it should be done in this app.

ix-dev/community/tailscale/templates/docker-compose.yaml
Shrinks99: Tailscale uses it to substitute the env var with the name of the user's tailnet. For details, see [their blog post on using Tailscale with Docker](https://tailscale.com/blog/docker-tailscale-guide#:~:text=Simply%20replace%20any%20hardcoded%20tailnet%20related%20values%20with%20%24%7BTS%5FCERT%5FDOMAIN%7D%20and%20on%20startup%2C%20the%20Tailscale%20sidecar%20proxy%20container%20will%20do%20all%20the%20heavy%20lifting%20of%20substituting%20those%20values%20as%20required%2E).

Shrinks99:
This part I actually _do_ know, I'm using the the [`TS_SERVE_CONFIG` env var](https://tailscale.com/kb/1282/docker?q=docker#ts_serve_config) to set this up with the main tailscale command and so we shouldn't need another entrypoint.

RE testing, yeah, I still need to do this throughly, as mentioned I only built the config.  When I try to run on my Mac without the render-only flag I get an error from Docker saying that the path is not known / shared from the host.  I don't think any other part of this config actually reads a file from anywhere so it makes sense that this wouldn't have been an issue up to now...  Should I try to pass this through to Docker locally?

Shrinks99:
Closing this.  Going to re-write manually, the old fashioned way... And hopefully feel like less of an ass.

stavros-k:
> RE testing, yeah, I still need to do this throughly, as mentioned I only built the config. When I try to run on my Mac without the render-only flag I get an error from Docker saying that the path is not known / shared from the host. I don't think any other part of this config actually reads a file from anywhere so it makes sense that this wouldn't have been an issue up to now... Should I try to pass this through to Docker locally?

It should work if you have these in the resources
image

stavros-k:
Ah I think its due to `/dev/tun`.
Set userspace to true in the test values, so it doesnt mount the tun device.

Shrinks99:
Thanks, I'll try that on the next go-around.  I think I'm also going to forego the fancy form for generating the JSON file and just give users a multi line input box if possible?  The contributing guide doesn't list this as an option (though it does say there are more form input available), and poking through some of the other questions files I can't find one.

Is there a questions schema somewhere I could reference?

stavros-k:
> Thanks, I'll try that on the next go-around.  I think I'm also going to forego the fancy form for generating the JSON file and just give users a multi line input box if possible?  The contributing guide doesn't list this as an option (though it does say there are more form input available), and poking through some of the other questions files I can't find one.
> 
> 
> 
> Is there a questions schema somewhere I could reference?

There was a multiple line support added sometime ago but I don't think it's worth adding. Better to just let the user mount a file via additional storage. (Already possible). And the user can use any ide to edit the file which will use syntax highlighting etc...

Shrinks99:
I find it preferable to keep these types of settings in the config because it lets me roll them back.  If I'm not mistaken, they get backed up with the server configuration, and also I don't have to create a dataset (or set up permissions) to store a single json file.  If you really think the host path version is preferable I'll implement it, but it seems overkill to me?  I feel like most users will copy the config on Tailscale's site and then change a few values.

EDIT: After some searching, if I'm not mistaken the type is `text` and it's only used in the ix-app questions file?  It also seems kinda broken on my NAS with a height of 0 and no `<textarea>` seems to be rendered on the page.

PR #4153, opened 2026-01-24
Add Tailscale Serve support to Tailscale app config

### LINKED ISSUE #3937: [Enhancement]: Tailscale Serve Setup Options

Shrinks99:
Closes #3937 

## AI

- [ ] Part or All of this PR was generated by an LLM.
  - [x] Not this time 🙃 

## Description

Adds a multi-line text field for entering a [Tailscale Services config file](https://tailscale.com/kb/1589/tailscale-services-configuration-file), copies this file and uses the [`TS_SERVE_CONFIG`](https://tailscale.com/kb/1282/docker?q=docker#ts_serve_config) env var.

## Testing

Tested locally with:

- [ ] basic-values.yaml

## Checklist

- [ ] App runs successfully locally
- [ ] Only modified files under /ix-dev/ or /library/
- [ ] Multiple test scenarios tested
- [ ] All automated CI checks pass


ix-dev/community/tailscale/templates/docker-compose.yaml
```diff
@@ -26,6 +26,11 @@
   {% do args.x.append(arg) %}
 {% endfor %}
 
+{% if values.tailscale.serve_config %}
```
stavros-k, line 29: If that's all its needed for serve, it can be done already without any mods.

you can mount a json file via additional storage + set the env var using additional envs

stavros-k:
Given that this PR does not really provide something that isn't easily available currently I'll close it for now.
I strongly believe that being able to use an actual editor to edit the json file is much better than copying back and forth in the UI.

PR #4200, opened 2026-01-30
home-assistant: Add NET_ADMIN capability

fenichelar:
Update: looks like TrueNAS does not include BlueZ so this won't work.

Needed for bluetooth adapters: https://www.home-assistant.io/integrations/bluetooth/#additional-details-for-container


stavros-k:
> Update: looks like TrueNAS does not include BlueZ so this won't work.

In that case, this PR should be closed, right?

fenichelar:
@stavros-k yes, I was not able to get bluetooth working on TrueNAS so I don't think this would be useful add to the container.

PR #4221, opened 2026-01-31
Fix: Tracearr - Add max_locks_per_transactionTimescaleDB

connorgallopo:
Fixes history backfill failing with 'out of shared memory' errors on large libraries. This parameter increases the lock table size which is needed when TimescaleDB creates many chunks during backfill operations.

Fixes connorgallopo/Tracearr#268

# Pull Request


stavros-k:
Thanks

PR #4235, opened 2026-02-02
LENS: Increment lens app version from 1.0.0 to 1.1.0

amrit3701:


stavros-k:
Hello, the renovate config must be updated instead so it picks updates automatically.
No manual updates are acceptable.

PR #4236, opened 2026-02-02
Arti: vanguards mode, renamings and update image tag

Keeqler:


ix-dev/community/arti/questions.yaml
```diff
@@ -85,8 +99,8 @@ questions:
                         min: 1
                         max: 65535
                         required: true
-                    - variable: hidden_service_port
-                      label: Hidden Service Port
+                    - variable: onion_service_port
+                      label: Onion Service Port
```
stavros-k: thats breaking change. anyone with a defined onion service will fail to upgrade

Keeqler: I can undo the variable renaming. Will renaming the label fail to upgrade as well?

stavros-k: Label can change freely yea

Keeqler:
Thanks

PR #4262, opened 2026-02-05
lib: initial support for networks

stavros-k:
Add support to join containers into networks.

- [x] Allows creation of "internal" networks. Those are used to interconnect the containers of the stack when needed (ie a db). Those internal networks have a label `ix.internal: true` attached, so we can later filter them out on our UI.
- [x] Allows to "register" an external network (it is needed to be present in the yaml and marked as `external: true`, in order be able to be attached to a container. Lifecycle of these networks are managed by the user, either via CLI, other tools, or later by our UI/middleware
- [x] Allows to set most of the common settings in the service/container level network options, such as `interface_name`, `ipv4_address`, `ipv6_address`, `mac_address`, `gw_priority`, `priority`. (Settings like driver, ipam, etc are on the top level network, which is handled externally).
- [x] Validates that external network name exists on the docker daemon by using the python SDK for docker client. Later this validation can either be removed (as it will be a `$ref` in the questions, or replaced with a middleware call)


amateescu:
This will make https://github.com/tjhorner/dragonify obsolete, right?

PR #4263, opened 2026-02-05
vaultwarden: allow joining networks

stavros-k:


ix-dev/community/vaultwarden/templates/test_values/basic-values.yaml
```diff
@@ -18,24 +18,31 @@ network:
     port_number: 8080
   certificate_id:
   domain: http://localhost:8080
+  networks:
+    - name: test_network
+      config:
+        interface_name: eth0
+      containers:
+        - vaultwarden
+        - postgres
```
stavros-k:
```suggestion
```

ix-dev/community/vaultwarden/questions.yaml
```diff
@@ -185,7 +185,80 @@ questions:
           schema:
             type: uri
             default: ""
-
+        - variable: networks
```
stavros-k, line 188: FIXME: this needs to be inverted...
Each container will have a list of networks + their config. 
That is because each container should have its own ip, if name etc.

PR #4273, opened 2026-02-06
dockhand: explicitly set the host path var

### LINKED ISSUE #4267: [Bug Report]: Dockhand wrong mount path
### App Name

Dockhand

1.0.6

### App Issue

Using the Host Path option for the default storage mount.

Previously, my stacks get stored in `/app/data/stacks/...`, but version 1.0.6 loses all my stacks. If I go poking around manually, I can find all of them in `/path/to/dockhand/config/on/truenas/stacks/...`, which dockhand doesn't know about by default. It's as if instead of mounting `/path/to/dockhand/config/on/truenas` to `/app/data`, it mounts `/path/to/dockhand/config/on/truenas` to `/path/to/dockhand/config/on/truenas`.

### App Logs

```shell

```

stavros-k:
Closes #4267

### DIFF
--- a/ix-dev/community/dockhand/app.yaml
+++ b/ix-dev/community/dockhand/app.yaml
@@ -32,4 +32,4 @@ sources:
 - https://hub.docker.com/r/fnsys/dockhand
 title: Dockhand
 train: community
-version: 1.0.6
+version: 1.0.7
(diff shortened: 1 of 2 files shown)


ajpanton:
@stavros-k I tried to message you in the bug report discussion, but got no reply. This does not fix the issue.

PR #4313, opened 2026-02-10
newt: add tmpfs for config directory

amateescu:
Newt writes its config to `/.config/newt-client/config.json` on startup. Since the container runs as a non-root user, it can't create directories at the filesystem root, resulting in:

```
INFO: 2026/02/10 22:27:13 Config file does not exist at /.config/newt-client/config.json, will create it
2026/02/10 22:27:13 Failed to create config directory: mkdir /.config: permission denied
INFO: 2026/02/10 22:27:14 Server version: 1.15.2
2026/02/10 22:27:14 Failed to create config directory: mkdir /.config: permission denied
INFO: 2026/02/10 22:27:14 Saving config to: /.config/newt-client/config.json
ERROR: 2026/02/10 22:27:14 Failed to save config: open /.config/newt-client/config.json: no such file or directory
```

Add a tmpfs mount at /.config so newt can write its config without needing root permissions. The config is ephemeral and doesn't need to persist across restarts.

### DIFF
--- a/ix-dev/community/newt/ix_values.yaml
+++ b/ix-dev/community/newt/ix_values.yaml
@@ -6,4 +6,4 @@ images:
 consts:
   newt_container_name: newt
   health_file_path: /tmp/newt_healthcheck
-  config_file_path: /tmp/newt_config
+  config_file_path: /tmp/newt_config.json


stavros-k:
I used an env var to move the config file to /tmp. should be good

ix-dev/community/newt/ix_values.yaml
```diff
@@ -6,3 +6,4 @@ images:
 consts:
   newt_container_name: newt
   health_file_path: /tmp/newt_healthcheck
+  config_file_path: /tmp/newt_config
```
stavros-k:
```suggestion
  config_file_path: /tmp/newt_config.json
```

PR #4322, opened 2026-02-11
Add XMRig app

Keeqler:
# App Addition

- [ ] I have opened an [issue](https://github.com/truenas/apps/issues) to discuss this app addition before submitting this pull request.

# AI

- [ ] Part or All of this PR was generated by an LLM.

## Description

Adds XMRig to the community train.

An open source CPU/GPU miner.

## App Information

- **Upstream**: https://github.com/xmrig/xmrig
- **Documentation**: https://xmrig.com/docs/miner
- **App Version**: 6.25.0

## Testing

Tested locally with:

- [x] basic-values.yaml

## Icons and Screenshots

- Icon: 
XMRig

## Special Notes

- XMRig requires connecting to a mining pool. We recommend using the P2Pool TrueNAS app.

## Checklist

- [x] README.md included


ix-dev/community/xmrig/templates/macros/entrypoint.sh
```diff
@@ -0,0 +1,125 @@
...
+# Creates configuration file if it doesn't exist.
+# The XMRig image already ships with a configuration file at this path, but the bind mount wipes it.
+
+set -e
+
+CONFIG_FILE_PATH="$HOME/.config/xmrig.json"
+
+if [ ! -e "$CONFIG_FILE_PATH" ]; then
+  mkdir -p "$(dirname "$CONFIG_FILE_PATH")"
+  cat > "$CONFIG_FILE_PATH" <<'JSON'
```
stavros-k: Do we really need this? Does it fail if the config is missing?

Can't we just have the user use the flags instead of overriding the config? Or are there settings only available to the config?

If the config is required in some cases, can the container have a default somewhere where we can copy it?

ix-dev/community/xmrig/templates/docker-compose.yaml
```diff
@@ -0,0 +1,85 @@
+{% from "macros/entrypoint.sh" import entrypoint %}
+{% set tpl = ix_lib.base.render.Render(values) %}
+
+{% set perm_container = tpl.deps.perms(values.consts.perms_container_name) %}
+{% set perm_config = {"uid": values.consts.run_as_user, "gid": values.consts.run_as_group, "mode": "check"} %}
+
+{% set xmrig = tpl.add_container(values.consts.xmrig_container_name, "image") %}
+{% do xmrig.set_user(values.consts.run_as_user, values.consts.run_as_group) %}
+{% do xmrig.set_privileged(true) %}
```
stavros-k: Why is that needed?

ix-dev/community/xmrig/templates/macros/entrypoint.sh
Keeqler: There are features only available with the config file and it also allows more advanced configuration, so some users might prefer using the config file.

A config file is shipped with the docker image at /root/.config/xmrig.json, which is where xmrig looks for it. But while testing, I found out that the storage bind mount wipes it because the volume's host dir is empty, so I made `entrypoint.sh` write another config file like that.

If having a huge json in `entrypoint.sh` like I did is an issue, I can place the config file somewhere else in the image and have it copied to the appropriate path from the entrypoint.

stavros-k: Yes having it somewhere where we can copy it is much better.
Also we can skip the flags altogether, and use `jq` to modify the config file.
(And we will always use the config file)

ix-dev/community/xmrig/templates/docker-compose.yaml
Keeqler: You mean `set_privileged`?

[MSR](https://xmrig.com/docs/miner/randomx-optimization-guide/msr), which allows increased hashrate, only works with a privileged container.

The container is actually supposed to be run as privileged only when the MSR option is enabled, I forgot to add this condition.

Keeqler: The "use MSR" question defaults to disabled and warns about privileged access being granted if enabled.

ix-dev/community/xmrig/templates/macros/entrypoint.sh
Keeqler: So is using the flags for the questions an issue in this case? I think it's much more simple and clean, the config file is not as simple.

Fwiw currently if `use_config_file` is enabled, all other questions get hidden/ignored and the config file edited by the user only is used, and no flags are used at all.

stavros-k: we shouldnt have the "additional flags" option, users can edit the json file.
We will just update it with jq with the limited options we have on ui

Keeqler: I'm actually trying to understand why we need to edit the config with jq instead of just using the flags

Keeqler: "Additional flags" could be hidden as well if "Use config file" is enabled.

stavros-k: So we dont have 2 ways of doing things.
We do only things via config file.

Anything not exposed in the UI can be done by editing the config file (by the user).

Having mutliple ways of doing the same thing doesnt really fit the apps

PR #4367, opened 2026-02-16
renovate: update emby's versioning

stavros-k:
https://forums.truenas.com/t/emby-stable-train-is-back-in-beta/55522/44

Emby's current released version will remain at beta until the next stable release.
The reason is: I don't know if we can downgrade exiting beta to previous (stable) version.

### DIFF
--- a/.github/renovate-config.js
+++ b/.github/renovate-config.js
@@ -92,7 +92,7 @@ module.exports = {
       ],
     ),
     customVersioning(
-      // YYYY-MM-DD-rN 
+      // YYYY-MM-DD-rN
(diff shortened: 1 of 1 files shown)


Qubad786 (approved):

psxlover:
@stavros-k emby released a stable version with the expected naming scheme (4.9.5.**0**) last week. Shouldn't the app catalog version have an update already?

stavros-k:
> @stavros-k emby released a stable version with the expected naming scheme (4.9.5.**0**) last week. Shouldn't the app catalog version have an update already?

Container is 17h ago
image

psxlover:
Oh, sorry. I had seen the release announcement last week, I saw that there was a docker version today when someone complained again, but I didn't think to check the date on the docker 😅

PR #4396, opened 2026-02-18
apps: add support for joining networks

stavros-k:


stavros-k:
Anything-llm fails to pull on GHA for some reason, it also fails to pull on my mac, but pulls fine on 2 truenas systems.

PR #4469, opened 2026-02-24
plex: fix name conflict

### LINKED ISSUE #4442: [Bug Report]: Adding Docker network to plex results in error
### App Name

Plex

1.3.2

### Enhancement Description

Adding an already existing Docker network to plex results in the following error:

Image

The newtork setup looks like this:

Image

### Use Case

Connect Plex to a Docker network.

### Proposed Solution

_No response_

### Alternatives Considered

_No response_

### Additional Context

_No response_

stavros-k:
Closes #4442

There is an issue in UI currently where the variable name is conflicting and causing unexpected results.
Rename the variable to fix the issue for existing TrueNAS version. UI will handle it for future releases.

### DIFF
--- a/ix-dev/stable/plex/questions.yaml
+++ b/ix-dev/stable/plex/questions.yaml
@@ -670,7 +670,7 @@ questions:
             type: list
             default: []
             items:
-              - variable: storageEntry
+              - variable: storage_entry
                 label: Storage Entry
                 schema:
                   type: dict


ix-dev/stable/plex/questions.yaml
```diff
@@ -71,7 +71,7 @@ questions:
               - 172.16.0.0/12
               - 192.168.0.0/16
             items:
-              - variable: network
+              - variable: network_entry
```
Qubad786, line 74: How would this work for existing users who already have the correct version installed?

stavros-k: I've tested it and it works fine! 
Basically this variable name does not actually gets used. 
The validation runs against the resulting array.

ix-dev/stable/plex/questions.yaml
```diff
@@ -670,7 +670,7 @@ questions:
             type: list
             default: []
             items:
-              - variable: storageEntry
+              - variable: storage_entry
```
stavros-k, line 673: will also update the rest of the apps in another PR to have a consistent casing across apps

PR #4490, opened 2026-02-27
donetick: set default rate limit env vars

BP602:
## Problem

On a fresh TrueNAS install, Donetick starts without a config file. When this happens, `RateLimit` and `RatePeriod` in `ServerConfig` are never populated by viper and default to zero. `ulule/limiter` treats a zero rate as immediately exhausted, so every request returns 429, including signup on a fresh install.

`selfhosted.yaml` sets `rate_limit: 300` and `rate_period: 60s`, but those values only apply if the file is found. No fallback defaults exist in the struct itself.

## Fix

Set `DT_SERVER_RATE_LIMIT` and `DT_SERVER_RATE_PERIOD` explicitly in the template, using the same values from Donetick's own [`config/selfhosted.yaml`](https://github.com/donetick/donetick/blob/main/config/selfhosted.yaml): 300 requests per 60 seconds.

## Test plan

- [x] Fresh install: signup completes without 429 errors
- [x] Rendered compose includes `DT_SERVER_RATE_LIMIT: "300"` and `DT_SERVER_RATE_PERIOD: 60s`


stavros-k:
Hello,

1. That sounds like a bug that has to be reported upstream
2. You are hardcoding these values and users wont be able to ever change them.

BP602:
1. Will report upstream. That said, the app is currently unusable on a fresh install: signup returns 429 on the first attempt, before the user can do anything.

2. Users can override these via Additional Environment Variables in the app config. `add_user_envs` runs after, so their values take precedence. These are just safe defaults to match what Donetick's own `selfhosted.yaml` ships with.

stavros-k:
> 2\. Users can still override these via Additional Environment Variables in the app config — `add_user_envs` runs after, so their values take precedence. These are just safe defaults to match what Donetick's own `selfhosted.yaml` ships with.

No, order doesnt matter here, if the env is set it blocks re-setting.

PR #4499, opened 2026-02-28
Listmonk : Make admin username and password optional in docker-compose

romainmp:
# Pull Request

Add conditional environment variable setup for admin credentials.

Since Listmonk version 4.0 admin username and password have been removed from configuration file.
Adding those varialbles throws warining in the application.

https://listmonk.app/docs/upgrade/#upgrading-to-v4xx


stavros-k:
If its removed why do we keep them?

romainmp:
> If its removed why do we keep them?

I guess to specify admin username and password for new installations and for migrating from older versions?

PR #4504, opened 2026-03-01
HyOS bump, memory env var

Reiss-Cashmore:
# Pull Request

This bumps HyOS and includes a new var, this var passes the container memory value that comes from the built in truenas question and passes it to the HyOS app so that it can dynamically scale the JVM memory values to fit within


ix-dev/community/hyos/templates/docker-compose.yaml
```diff
@@ -54,14 +55,14 @@
...
-{% do mgr.environment.add_env("HYTALE_SERVER_PORT", values.network.api_port.port_number) %}
+{% do mgr.environment.add_env("HYTALE_SERVER_PORT", values.consts.internal_api_port) %}
 {% do mgr.environment.add_env("API_CLIENT_SECRET", values.hytale.api.client_secret) %}
 {% do mgr.environment.add_env("API_CLIENT_ID", values.hytale.api.client_id) %}
 {% do mgr.environment.add_env("COOKIE_SECURE", values.hytale.cookie_secure) %}
 {% do mgr.environment.add_env("PORT", values.network.manager_port.port_number) %}
 
 {% do server.add_port(values.network.game_port, {"protocol": "udp"}) %}
-{% do server.add_port(values.network.api_port) %}
+{% do server.add_port(values.network.api_port, {"container_port": values.consts.internal_api_port}) %}
```
stavros-k: what is this change?

Reiss-Cashmore: Port <truenas-ip>:30381 should expose 8080 from the internal server docker container network. Right now 30381 is mapped to 30381 internal. Just following same implementation as ix-dev/community/castopod/ix_values.yaml and ix-dev/community/rackula/ix_values.yaml

stavros-k: You can set the API_PORT I think?
Pattern is to match outside:inside ports when possible.

ix-dev/community/hyos/templates/docker-compose.yaml
```diff
@@ -27,6 +27,7 @@
 {% do server.environment.add_env("HYTALE_AUTH_MODE", values.hytale.auth_mode) %}
 {% do server.environment.add_env("SKIP_BROKEN_MODS", values.hytale.skip_broken_mods) %}
 {% do server.environment.add_env("MAX_VIEW_RADIUS", values.hytale.max_view_radius) %}
+{% do server.environment.add_env("CONTAINER_MEMORY_MB", values.resources.limits.memory) %}
```
stavros-k: I think you can find this value from `/sys/fs/cgroup/memory.max` It should work on rootless as well

Reiss-Cashmore: Perfect, will look into it thank you

PR #4510, opened 2026-03-02
Re add Nextcloud 32 for while

stavros-k:


ix-dev/stable/nextcloud/to_keep_versions.yaml
```diff
@@ -1,2 +1,4 @@
 # Last version of v31.x.x | Remove 07 Jan 2026
```
stavros-k, line 1: Lets keep this for couple more months.

Qubad786 (approved):

PR #4552, opened 2026-03-09
[Enhancement]: SeaweedFS — Expose S3 IAM read-only flag in UI   

tina-junold:
**App Name**                                                  
  SeaweedFS

**Current Version (TrueNAS Catalog App Version)**
  1.2.2

**Enhancement Description**
  Add a configurable S3 IAM Read Only toggle to the SeaweedFS app. The SeaweedFS s3 command supports a -readOnly flag that controls whether the S3 gateway enforces read-only access via IAM policy. Currently this flag is not
  exposed in the TrueNAS app UI and always defaults to the SeaweedFS binary default.

**Use Case**
  SeaweedFS ships with IAM read-only mode active by default. Users who need to manage IAM users and policies programmatically — e.g. via Terraform — must be able to disable this restriction so that write operations against the
   IAM API are permitted. Without this toggle, there is no way to provision S3 credentials through the standard toolchain. 

**Proposed Solution**
  Add a boolean option s3_iam_readonly to questions.yaml (under the seaweedfs configuration group). In docker-compose.yaml, always pass -readOnly=true or -readOnly=false explicitly to the S3 container entrypoint based on the
  selected value.

**Alternatives Considered**
  Restricting access via IAM user policies alone — possible but requires manual setup outside the app and does not provide a single, visible toggle for operators.

**Additional Context**
  Relevant SeaweedFS source: https://github.com/seaweedfs/seaweedfs/blob/master/weed/command/s3.go — the -readOnly flag is documented there.

### DIFF
--- a/ix-dev/stable/seaweedfs/app.yaml
+++ b/ix-dev/stable/seaweedfs/app.yaml
@@ -59,4 +59,4 @@ sources:
 - https://hub.docker.com/r/chrislusf/seaweedfs
 title: SeaweedFS
 train: stable
-version: 1.2.4
+version: 1.2.5


ix-dev/stable/seaweedfs/templates/test_values/basic-values.yaml
```diff
@@ -10,6 +10,10 @@ seaweedfs:
   admin_user: admin
   admin_password: admin123
   master_peers: []
+  additional_flags:
+    - container: s3
+      flag: "readOnly"
```
stavros-k:
```suggestion
      flag: "iam.readOnly"
```

Issue #4559, opened 2026-03-10
SplitPro App Missing `uploads` Storage Configuration


YTKme:
I was trying to install SplitPro but encountered the below error.

I checked the [docker documentation](https://github.com/oss-apps/split-pro/blob/main/docker/prod/compose.yml#L43), it seem to be missing a mapping to the `/app/uploads` mount. I tried adding an Additional Storage, but that did not work either.

```text
[EFAULT] Failed to render compose templates: Traceback (most recent call last): File "/usr/bin/apps_render_app", line 33, in sys.exit(load_entry_point('apps-validation==0.1', 'console_scripts', 'apps_render_app')()) ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "/usr/lib/python3/dist-packages/catalog_templating/scripts/render_compose.py", line 48, in main render_templates_from_path(args.path, args.values) File "/usr/lib/python3/dist-packages/catalog_templating/scripts/render_compose.py", line 19, in render_templates_from_path rendered_data = render_templates( ^^^^^^^^^^^^^^^^^ File "/usr/lib/python3/dist-packages/catalog_templating/render.py", line 31, in render_templates rendered_templates[i.name] = env.get_template(i.name).render( ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "/usr/lib/python3/dist-packages/jinja2/environment.py", line 1301, in render self.environment.handle_exception() File "/usr/lib/python3/dist-packages/jinja2/environment.py", line 936, in handle_exception raise rewrite_traceback_stack(source=source) File "/mnt/.ix-apps/app_configs/split-pro-tertiary/versions/1.0.1/templates/docker-compose.yaml", line 41, in top-level template code {% do c1.add_storage("/app/uploads", values.storage.uploads) %} ^^^^^^^^^^^^^^^^^^^^^^^^^ File "/mnt/.ix-apps/app_configs/split-pro-tertiary/versions/1.0.1/templates/library/base_v2_2_8/container.py", line 323, in add_storage if config.get("type", "") == "tmpfs": ^^^^^^^^^^ jinja2.exceptions.UndefinedError: 'dict object' has no attribute 'uploads' 
```

PR #4561, opened 2026-03-10
[fix] Add uploads to split-pro questions

drryu:
Fixes - https://github.com/truenas/apps/issues/4559

uploads is missing from the questions.yaml. The docker-compose render fails due to missing uploads.


YTKme:
Thank you for the update!

PR #4595, opened 2026-03-15
Add HTTPS support to immich app

lumarel:
This PR makes it possible to run immich with HTTPS for the web_port.

I've been using collabora office for a while and noticed how it is setup there, so I thought I would add the same nginx layer for immich too.

This also changes the internal immich web_port to always be the default 2283 (mentioned in the immich docs).

If there are any further improvements I can add to this PR I will gladly change it.

Cheers, lumarel

### DIFF
--- a/ix-dev/community/immich/app.yaml
+++ b/ix-dev/community/immich/app.yaml
@@ -1,7 +1,17 @@
 annotations:
   min_scale_version: 24.10.2.2
 app_version: v2.5.6
-capabilities: []
+capabilities:
+- description: Nginx is able to change file ownership arbitrarily
+  name: CHOWN
+- description: Nginx is able to bypass file permission checks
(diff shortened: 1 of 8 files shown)


stavros-k:
> This also changes the internal immich web_port to always be the default 2283 (mentioned in the immich docs).

There was no reason at all to make that change.


That being said, I see no benefit of adding this. Collabora's nginx is there because of backwards compat coming from the k8s backend. If you need https, I'd suggest to use one of the many reverse proxy options out there. Some of them are already in the catalog.

I'm not keen maintaining nginx config files unless its absolutely necessary for the app to work.

Sorry, but I'll have to close this. Please open issues before extending things to avoid doing work that might not be accepted. 

Thanks

lumarel:
No worries, thanks anyways 🙂

PR #4596, opened 2026-03-16
Update I2P version and URLs to use i2p.net

WaxySteelWorm:
Bump version to I2P-2.11.0-4 and updating urls from geti2p.net to i2p.net.


stavros-k:
> Bump version to I2P-2.11.0-4

There is no such tag.
image

PR #4612, opened 2026-03-17
fix condition evaluation on registration and federation in continuwuity

atridadl:
This PR fixes an issue with Continuwuity 1.0.0 where selecting Federation or Registration does not bring up the text fields to enter the token or trusted servers. I suspect this meant trying to deploy the app sent empty strings.

Screenshot of issue:
image

I'm fairly sure this was responsible:

[["allow_registration", "=", true]]

The boolean was a string literal previously. None of the other apps appear to do this.

After running the tests, the app appears to run locally:

image


stavros-k:
https://github.com/truenas/apps/blob/11cea4231d844ef15128941d2142a20d20902c95/CONTRIBUTIONS.md?plain=1#L139

atridadl:
Oh god missed that. LOL I'll fix that my bad.

atridadl:
Sorry about that @stavros-k ... long day. Totally should have seen that. Makes a LOT more sense now though as that had me confused.

catalog.json
stavros-k, line 1: revert this....

atridadl: Whoops. Reverted to the wrong commit. Fixed.

PR #4666, opened 2026-03-23
homepage: Add NET_RAW capability for ICMP ping health checks

louga31:
# AI

The PR description was generated with AI assistance.

## Description

Adds the `NET_RAW` Linux capability and removes `no-new-privileges` security option from the Homepage container to allow ICMP ping status checks to function correctly.

Homepage uses ICMP ping to verify the status of configured services via the `ping` field in `services.yaml`. The current app definition drops all capabilities (`cap_drop: ALL`) and sets `no-new-privileges=true`, which prevents the setuid `ping` binary from working even when `NET_RAW` is added back. This causes all ping-based status checks to fail with:

```
ping: socket: Operation not permitted
ping: => missing cap_net_raw+p capability or setuid?
```

This results in all services showing as offline in the Homepage dashboard when using the `ping` field, even though the services are reachable.

### Root cause

The `ping` binary in the Homepage container image (`/bin/ping`) relies on the setuid bit (`-rwsr-xr-x`). The `no-new-privileges` security option blocks setuid from elevating privileges, and since the container runs as non-root (uid 1000), the effective capabilities remain empty (`CapEff: 0000000000000000`) despite `NET_RAW` being in the bounding set.

### Fix

- Add `NET_RAW` capability via `add_caps`
- Remove `no-new-privileges` via `remove_security_opt`

## Changes

- **`ix-dev/community/homepage/app.yaml`**: Added `NET_RAW` to `capabilities`, bumped version to `1.3.4`
- **`ix-dev/community/homepage/templates/docker-compose.yaml`**: Added `{% do c1.add_caps(["NET_RAW"]) %}` and `{% do c1.remove_security_opt("no-new-privileges") %}`

## Testing

Tested locally with:

- [x] basic-values.yaml

Verified that:

- Homepage container starts correctly with the added capability
- `CapEff` now includes `NET_RAW` for the container process
- ICMP ping checks succeed for configured services
- No regression on existing functionality (web UI, widgets, Docker socket integration)

## Checklist

- [x] Only modified files under /ix-dev/


ix-dev/community/homepage/templates/docker-compose.yaml
```diff
@@ -5,6 +5,8 @@
 {% set perms_config = {"uid": values.consts.run_user, "gid": values.consts.run_group, "mode": "check"} %}
 
 {% do c1.set_user(values.consts.run_user, values.consts.run_group) %}
+{% do c1.add_caps(["NET_RAW"]) %}
+{% do c1.remove_security_opt("no-new-privileges") %}
```
stavros-k, line 9: Why is this needed?
(The removal of `no-new-privileges`)

PR #4664, opened 2026-03-22
feat: add stdapi.ai app

JGoutin:
# App Addition

- [x] I have opened an [issue](https://github.com/truenas/apps/issues) to discuss this app addition before submitting this pull request. See : https://github.com/truenas/apps/issues/4663

# AI

- [x] Part or All of this PR was generated by an LLM.

## Description

Adds **stdapi.ai** to the community train.

[stdapi.ai](https://stdapi.ai) is an OpenAI and Anthropic compatible API gateway for AWS Bedrock. Can be used to connect various AI tools to AWS Bedrock models using OpenAI and Anthropic APIs, this includes applications already available on the TrueNas app catalog, like **Open WebUI** or **AnythingLLM**.

## App Information

- **Upstream**: https://github.com/stdapi-ai/stdapi.ai
- **Documentation**: https://stdapi.ai/
- **App Version**: 1.7.0

## Testing

Tested locally with:

- [x] basic-values.yaml

All tests passed successfully.

## Icons and Screenshots

Please upload the following to the CDN:

- Icon: https://stdapi.ai/styles/logo.svg

## Special Notes

- First-time setup instructions: AWS Credentials with minimal permission required. [See documentation](https://stdapi.ai/operations_configuration/#iam-permissions). 
- Placeholder for credential is also set into ix-dev/community/stdapi-ai/templates/test_values/basic-values.yaml
- Application tested on Truenas successfully (As a custom app).
- Application tested locally using generated docker-compose successfully (With AWS credentials).

## Checklist

- [x] App runs successfully locally
- [x] Only modified files under /ix-dev/ or /library/
- [x] README.md included
- [x] Multiple test scenarios tested
- [x] questions.yaml has clear descriptions and follows structure of existing apps
- [x] All automated CI checks pass

PR #4689, opened 2026-03-25
fix: stdapi-ai app health check to use wget instead of curl

JGoutin:
I noticed the app gets stuck in the "Deploying" state after installation, even though it works perfectly fine otherwise.

It looks like a health check issue. The current template uses curl to check the `/health` endpoint:
`{% do c1.healthcheck.set_test("curl", {"port": values.network.api_port.port_number, "path": "/health"}) %}`

Since the image is Alpine-based, curl isn't installed, causing the check to fail. We can easily fix this by switching to BusyBox's wget, which is already available on the image:
`{% do c1.healthcheck.set_test("wget", {"port": values.network.api_port.port_number, "path": "/health"}) %}`

Follow #4664

### DIFF
--- a/ix-dev/community/stdapi-ai/templates/docker-compose.yaml
+++ b/ix-dev/community/stdapi-ai/templates/docker-compose.yaml
@@ -3,7 +3,7 @@
 {% set c1 = tpl.add_container(values.consts.stdapi_ai_container_name, "image") %}
 
 {% do c1.set_user(65532, 65532) %}
-{% do c1.healthcheck.set_test("wget", {"port": values.network.api_port.port_number, "path": "/health"}) %}
+{% do c1.healthcheck.set_test("wget", {"port": values.network.api_port.port_number, "path": "/health", "busybox": true}) %}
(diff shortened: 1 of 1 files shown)


ix-dev/community/stdapi-ai/templates/docker-compose.yaml
```diff
@@ -3,7 +3,7 @@
 {% set c1 = tpl.add_container(values.consts.stdapi_ai_container_name, "image") %}
 
 {% do c1.set_user(65532, 65532) %}
-{% do c1.healthcheck.set_test("curl", {"port": values.network.api_port.port_number, "path": "/health"}) %}
+{% do c1.healthcheck.set_test("wget", {"port": values.network.api_port.port_number, "path": "/health"}) %}
```
stavros-k:
```suggestion
{% do c1.healthcheck.set_test("wget", {"port": values.network.api_port.port_number, "path": "/health", "busybox": true}) %}
```

Issue #4699, opened 2026-03-26
[BUG] Rust Desk no longer supports Host Network


N-Hertstein:
Due to Issue # & PR # the Host network option has been removed.

The Rustdesk documentation specifically states Host network as a requirement for the pro image. It most definitely should be an option.
https://rustdesk.com/docs/en/self-host/rustdesk-server-pro/installscript/docker/

And saying that automatic relay setup will be non-functional like said in issue #3627 is not a valid argument as in my testing with both the pro and OSS image this does in fact work. I have gone over the current docker-compose template and you do not "We use internal docker network to connect server/relay" There is no template level configuration required or available to configure this. Rustdesk connects based on a shared configuration file used by both images (oversimplification).

Breaking the intended setup by the developer is **not** an appropriate solution. Especially since the MAC-Address workaround is unstable at best.

Solutions:
- Split off Pro image as seperate app.
- Re-add host network

stavros-k:
> There is no template level configuration required or available to configure this.

There is, they connect via a docker network, which is not possible with host network.

https://github.com/truenas/apps/blob/4666d9467760494bb2cdb27903d8e3ffbe33522b/ix-dev/community/rust-desk/templates/docker-compose.yaml#L3-L9

> Especially since the MAC-Address workaround is unstable at best.

How so? 

> Rustdesk connects based on a shared configuration file used by both images

Server needs to know the address/port of the relay (used in `command:` generated).
When using host network, it bypasses the docker networks, so we can't reliably generate this address.

N-Hertstein:
> > There is no template level configuration required or available to configure this.
> 
> There is, they connect via a docker network, which is not possible with host network.
> 
> [apps/ix-dev/community/rust-desk/templates/docker-compose.yaml](https://github.com/truenas/apps/blob/4666d9467760494bb2cdb27903d8e3ffbe33522b/ix-dev/community/rust-desk/templates/docker-compose.yaml#L3-L9)
> 
> Lines 3 to 9 in [4666d94](/truenas/apps/commit/4666d9467760494bb2cdb27903d8e3ffbe33522b)
>  {% set rust_desk_net = tpl.networks.create_internal("rust-desk-net") %} 
>   
>  {% set server = tpl.add_container(values.consts.rust_desk_container_name, values.rust_desk.image_selector) %} 
>  {% do server.add_network(rust_desk_net) %} 
[...]

Issue #4560, opened 2026-03-10
Rust desk app fails to start with Host Network enabled


akhudushin:
Rust desk app fails to start with `Host Network` parameter set to enabled.

Probably related to https://github.com/truenas/apps/commit/f097c597b5854df72e05645d381769c03a7ece86
https://github.com/truenas/apps/blob/dfb2770640c2fd9526ae01457f401773a8bd583e/ix-dev/community/rust-desk/templates/docker-compose.yaml#L3

Error:
```
[...]

PR #4700, opened 2026-03-26
Add Host Network option to Rust Desk

N-Hertstein:
We do not know better how to deploy Rust Desk than the Rust Desk devs. Being unable to deploy using the developer intended solution is **arrogant** at best. This is why the Duckstation guy hates people like us.

Solves Issue #4699 and #4560

### DIFF
--- a/ix-dev/community/rust-desk/app.yaml
+++ b/ix-dev/community/rust-desk/app.yaml
@@ -3,38 +3,38 @@ annotations:
 app_version: 1.1.15
 capabilities: []
 categories:
-  - productivity
+- productivity
 changelog_url: https://github.com/rustdesk/rustdesk-server/releases
-date_added: "2024-09-13"
+date_added: '2024-09-13'
 description: Rust Desk is an open-source remote desktop, and alternative to TeamViewer.
 home: https://rustdesk.com
 host_mounts: []
 icon: https://media.sys.truenas.net/apps/rust-desk/icons/icon.png
 keywords:
-  - remote
-  - desktop
+- remote
+- desktop
 lib_version: 2.2.2
 lib_version_hash: 68a92a0c1f8193b040cc1193f9648e9ec8bd6cef0b45d9d5c0d1a6e641f8146b
 maintainers:
-  - email: dev@truenas.com
-    name: truenas
-    url: https://www.truenas.com/
+- email: dev@truenas.com
+  name: truenas
+  url: https://www.truenas.com/
 name: rust-desk
 run_as_context:
-  - description: Container [rust-desk] can run as any non-root user and group.
-    gid: 568
-    group_name: Host group is [apps]
-    uid: 568
-    user_name: Host user is [apps]
-  - description: Container [rust-desk-relay] can run as any non-root user and group.
-    gid: 568
-    group_name: Host group is [apps]
-    uid: 568
-    user_name: Host user is [apps]
+- description: Container [rust-desk] can run as any non-root user and group.
+  gid: 568
+  group_name: Host group is [apps]
+  uid: 568
+  user_name: Host user is [apps]
+- description: Container [rust-desk-relay] can run as any non-root user and group.
+  gid: 568
+  group_name: Host group is [apps]
+  uid: 568
+  user_name: Host user is [apps]
 screenshots: []
 sources:
-  - https://hub.docker.com/r/rustdesk/rustdesk-server
-  - https://github.com/rustdesk/rustdesk-server
+- https://hub.docker.com/r/rustdesk/rustdesk-server
+- https://github.com/rustdesk/rustdesk-server
 title: Rust Desk
 train: community
 version: 1.6.1


stavros-k:
> Being unable to deploy using the developer intended solution is **arrogant** at best.

I'd argue that is weird for a **containerized** app to depend on the hostnet _just_ for licensing purposes. 

> This is why the Duckstation guy hates people like us.

I have no idea what this means.

---

To be clear, this won't be merged, if we conclude that host network is required for good reason, the correct solution is to split relay and server.

So users will install each, and provide the actual connection string of the relay.

Can you please explain to me why host network is required?

ix-dev/community/rust-desk/templates/docker-compose.yaml
stavros-k, line 33: How does this work when host network is enabled?

N-Hertstein: Relay servers that are correctly created in the same compose file as the OSS or Pro image do not require additional configuration. Will test with a secondary relay.

N-Hertstein:
> > Being unable to deploy using the developer intended solution is **arrogant** at best.
> 
> I'd argue that is weird for a **containerized** app to depend on the hostnet _just_ for licensing purposes.

I'd argue that not supporting an app the intended way is worse than not supporting it, because it creates answers from the developer to questions that resemble: "Install it the intended way and try again"
> 
> > This is why the Duckstation guy hates people like us.
> 
> I have no idea what this means.


Duckstation is not implemented according to how the developer has designed the sofware on the AUR and this has generated an influx of support requests.

> To be clear, this won't be merged, if we conclude that host network is required for good reason, the correct solution is to split relay and server.
>
> So users will install each, and provide the actual connection string of the relay.

If the correct solution was splitting of the relay server this would have been the setup Rust Desk recommends.

> Can you please explain to me why host network is required?

Because my server doesn't run without it. I tried the mac-address option and would you like to know the response I get from my fancy paid license customer support line?
"If it doesn't work without host network, use host network."

It is required because they made it so. If you do not wish to **properly** support it, deprecate at least the Pro image. I was the one who wanted the pro version in the first place so I can make it into a separate app or manage my own compose file.
If you want 100% understanding of this proprietary version of the app, learn rust. But do not pretend to support the Pro image if you're breaking Rust-Desk customer support.

stavros-k:
> I was the one who wanted the pro version in the first place so I can make it into a separate app or manage my own compose file.
> If you want 100% understanding of this proprietary app, learn rust. But do not pretend to support the Pro image if you're breaking Rust-Desk customer support.

As I already said above, if it IS required, the way forward is to split relay/server.
But I need to know what you mean "unstable" using mac address option.

N-Hertstein:
> > I was the one who wanted the pro version in the first place so I can make it into a separate app or manage my own compose file.
> > If you want 100% understanding of this proprietary app, learn rust. But do not pretend to support the Pro image if you're breaking Rust-Desk customer support.
> 
> As I already said above, if it IS required, the way forward is to split relay/server. 

If that was true that would've been the recommended installation method. However A separate relay would be useful for people with a second truenas server that wish to use multiple relays.

>But I need to know what you mean "unstable" using mac address option.

I mean that every other update I had to reset my license and reregister my server. So I switched to Host Network.

N-Hertstein:
> > I was the one who wanted the pro version in the first place so I can make it into a separate app or manage my own compose file.
> > If you want 100% understanding of this proprietary app, learn rust. But do not pretend to support the Pro image if you're breaking Rust-Desk customer support.
> 
> As I already said above, if it IS required, the way forward is to split relay/server. But I need to know what you mean "unstable" using mac address option.

Oh and going from one non-recommended installation option to another would still break my Rust Desk customer support function into only outputting "use the intended installation method"

stavros-k:
> Oh and going from one non-recommended installation option to another would still break my Rust Desk customer support function into only outputting "use the intended installation method"

How is splitting the components non-recommended? The only thing that is shared when using the same compose file is the network, which is not the case anymore once you enable host network.

N-Hertstein:
> > Oh and going from one non-recommended installation option to another would still break my Rust Desk customer support function into only outputting "use the intended installation method"
> 
> How is splitting the components non-recommended? The only thing that is shared when using the same compose file is the network, which is not the case anymore once you enable host network.

Well, not recommended is a bit much I apologize. But let's say I have my Pro image and my separate relay and they don't connect.
As a customer support guy hearing "I installed the 'components your docs say should be installed together for maximum reliability', separately" My first troubleshooting step to give would be "Could you try installing them together?"

stavros-k:
> > > Oh and going from one non-recommended installation option to another would still break my Rust Desk customer support function into only outputting "use the intended installation method"
> > 
> > 
> > How is splitting the components non-recommended? The only thing that is shared when using the same compose file is the network, which is not the case anymore once you enable host network.
> 
> Well, not recommended is a bit much I apologize. But let's say I have my Pro image and my separate relay and they don't connect. As a customer support guy hearing "I installed the 'components your docs say should be installed together for maximum reliability', separately" My first troubleshooting step to give would be "Could you try installing them together?"

In this case, even if we match their compose file 100%, their next response will be: "Please dont use an app catalog but deploy the compose file manually" or "please dont use a NAS, but a vps".

We all know how support follows a strict script and answers.

> Fun fact, I almost paid a lic for 2k machines last year. But the lack of proper communication moved me to another solution which was equally priced and managed.

---

I still believe that splitting is the best option, one could just want to deploy only a relay in their TrueNAS.

N-Hertstein:
> > Fun fact, I almost paid a lic for 2k machines last year. But the lack of proper communication moved me to another solution which was equally priced and managed.

What's the other solution? If the support is going to be bad either way I might switch next year.

> I still believe that splitting is the best option, one could just want to deploy only a relay in their TrueNAS.

I do believe having a relay-only setup should be added, however the combined setup should remain to support current users. And this does not solve the license/host network issue for the Pro image.

N-Hertstein:
Also I did make a new issue that might be a better place to have this discussion than in a PR draft.

> We do not know better how to deploy Rust Desk than the Rust Desk devs. Being unable to deploy using the developer intended solution is **arrogant** at best. This is why the Duckstation guy hates people like us.
> 
> Solves Issue #4699 and #4560

It's #4699

N-Hertstein:
I'm over it. I'm running my own compose file the way it's intended. Do as you please with this non-functional pro setup.

Issue #3465, opened 2025-10-25
[Enhancement]: Add `accept-routes` option to Tailscale


MarcMush:
### App Name

Tailscale

### Current Version (TrueNAS Catalog App Version)

1.3.16

### Enhancement Description

add a checkbox to pass the `--accept-routes` argument

### Use Case

another machine on my tailscale network advertises subnets that I want to accept from my truenas machine

PR #4735, opened 2026-03-31
tailscale: explicitly pass =false for boolean flags to reset persiste…

flennic:
I encountered an issue that my TrueNAS was not reachable via its local LAN address on layer 3, while layer 2 was working, when enabling the tailscale addon. When I dug into it I noticed that the flags (in my case --accept-routes) we actually set, while the UI shows them as disabled. I then figured out that if these flags are false, they are omitted from TS_EXTRA_ARGS. This leads to the Tailscale container to read from the persitent state file. So this means that if the flags were enabled once, they are basically always enabled. Explicitely setting them to false should fix this.

**Note:** The flag reads advertised Tailscale networks and if one is the local network, there is some sort of loop where it sends this out to Tailscale instead of directly on the LAN NIC, hence the TrueNAS instance remains reachable only via it's tailscale address but not the local address.

I found a related test but it just tests the parsing, no runtime behavior. So I think this here should work, but I have to admit I have not tested this final fix (as I only manually modified the container settings via the CLI).

**Note:** Whatever template should be visible for PRs, I unfortunately cannot see it. Hence this free-styled version.

This is related to an earlier PR [#3465](https://github.com/truenas/apps/pull/3474) which enabled this flags for #3465.


stavros-k:
--reset seems to also be a bool flag, so passing the flag there always seems more appropriate.
What do you think?

flennic:
Thanks for the quick reply and thanks for bumping the version, I missed that one.

Regarding the `--reset` flag, I am not sure. Looking at https://tailscale.com/docs/reference/tailscale-cli it looks like a one-time flag which does not really take in a true or false, it's more like a flag which exists or not. I have not found any example online where it is `true` or `false` hence I added the exception for it, because I don't know its behavior with a boolean value.

I am not sure on this, but at least you have some input on why I did this.

stavros-k:
https://github.com/tailscale/tailscale/blob/1e2fdfd745fec5339a003ba68cce0ce9f785859a/cmd/tailscale/cli/up.go#L143
This seems to defaulting to false already so it should be safe

stavros-k:
I dont see any weird messaging regarding reset in the logs either!

PR #4738, opened 2026-04-01
InvoiceNinja: Allow specifying db name + user

thirstyice:
There was no template available that matched the purpose of this PR.
Let me know if you need any further info.

# InvoiceNinja

## App Enhancement(?)

Allow user to specify db name and user on app installation
(Makes migration of existing databases much easier).

### DIFF
--- a/ix-dev/community/invoice-ninja/ix_values.yaml
+++ b/ix-dev/community/invoice-ninja/ix_values.yaml
@@ -23,5 +23,3 @@ consts:
   mariadb_container_name: mariadb
   redis_container_name: redis
   admin_internal_port: 2019
-  db_user: ninja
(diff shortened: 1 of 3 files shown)


stavros-k:
At this moment there are no plans to accept this change. 
(Note that if it is ever done, it has to be done across ALL apps)

The db setup does not always match your previous version or settings.
Allowing this will just result in a flux of issues.

The best way forward is to dump your previous db and restore it on the new app.

Thanks

thirstyice:
Can you provide instructions for how to do that? I wasn't able to find anything like that in the web ui. I would very much like to migrate my existing invoiceninja custom app to the more official one.

PR #4825, opened 2026-04-14
Add basic netbird server

### LINKED ISSUE #1674: Community App Suggestion: NetBird Server
https://github.com/netbirdio/netbird

stavros-k:
Closes #1674

---

Currently only management server + dashboard.

Proxy container will be added on subsequent PR once verified that base thing works and that is actual need for the proxy container. As it seems that in order to setup proxy there needs to be a "dance". ie first install it, create a token and then go back to enable the proxy container


TechHutTV:
https://github.com/truenas/apps/issues/1674#issuecomment-4246134620

Qubad786 (approved):

alfonso133:
@stavros-k when I try to install the app via the install button in discover section that should take me to the edit screen to set up the app I instead get the error "Invalid operation: ==".
Checking the questions.yaml file it seems to be caused by row 160: `show_if: [["bind_mode", "==", "published"]]` that shoud be `show_if: [["bind_mode", "=", "published"]]`. I reported it here directly since I don't know if I should open a PR considering I'm not creating a new app and I'm not the maintainer

PR #4858, opened 2026-04-20
forgejo-runner: new registration process

### LINKED ISSUE #4854: [BUG]: Forgejo Runner Configuration deprecated - Runner failed to register
### App Name

forgejo runner

App Version: 12.8.2 Version: 1.1.10

### Enhancement Description

Out of the Box install doesn't work as runner registration has been changed

### Use Case

Installed forgejo and forgejo runnner. While the  config.yaml is being created the App Catalouge Web UI still requires the token to be set 

Mirating to custom app you see whats being called:
```sh
   if [ ! -f /data/.runner ]; then
        echo "Registering the runner"
        forgejo-runner register \
          --config /data/config.yaml \
          --no-interactive \
          --name truenas-runner \
          --instance <URL>
          --token <TOKEN>
```

The container will crash straight away with the following error:

```sh
Runner failed to register
Registering the runner
level=warning msg="`register` has been deprecated; declare connections in the runner configuration instead"
level=info msg="Registering runner, arch=amd64, os=linux, version=v12.8.2."
level=warning msg="Runner in user-mode."
level=debug msg="Successfully pinged the Forgejo instance server"
level=error msg="poller: cannot register new runner" error="invalid_argument: runner registration token not found"
Error: Failed to register runner: invalid_argument: runner registration token not found
Runner failed to register
```

### Proposed Solution

either pre-populate config.yaml with settings from TrueNas UI or remove depecated UI config items and point the user to the config.yaml instead

### Alternatives Considered

_No response_

### Additional Context

Runner failed to register: level=warning msg="`register` has been deprecated; declare connections in the runner configuration instead"

Deprecate and Remove Runner File (.runner) in Favour of Runner Configuration
https://code.forgejo.org/forgejo/forgejo-actions-feature-requests/issues/88

stavros-k:
Forgejo Runner registration works differently now.
https://forgejo.org/2026-04-release-v15-0/#simplified-runner-registration

There is not really a migration path as I can see it, as it needs different credentials (token and uuid).


MRi-LE:
Thx for PR. CoPilot made the same suggestions as Claude

> Looking at the CI logs, the failure is clear from the container logs:
> level=error msg="fail to invoke Declare" error="unavailable: net/http: invalid header field value for \"X-Runner-Token\""
> The problem: The runner token (some-secret) is being written to /tmp/runner-token and passed via --token-url file:///tmp/runner-token, but the forgejo-runner binary is reading that token and sending it as an HTTP header value. The string some-secret contains characters (or possibly a trailing newline from echo) that are invalid in HTTP headers.
> The most likely culprit is that echo 'some-secret' appends a newline character (\n) to the file, and when the runner reads it and stuffs it into the X-Runner-Token header, HTTP rejects it because newlines are illegal in header values.
> The fix in the entrypoint script should use printf instead of echo to avoid the trailing newline:
> bashprintf '%s' 'some-secret' > /tmp/runner-token
> or with echo -n:
> bashecho -n 'some-secret' > /tmp/runner-token
> This is a common gotcha when writing secrets to files for token-URL consumption — the reader passes the raw file contents verbatim into the HTTP header, newline included, which fails RFC 7230 header validation.
> The test value some-secret itself is fine as a string; it's purely the newline that echo adds that breaks it.

Qubad786 (approved):

stavros-k:
> Thx for PR. CoPilot made the same suggestions as Claude
> 
> > Looking at the CI logs, the failure is clear from the container logs:
> > level=error msg="fail to invoke Declare" error="unavailable: net/http: invalid header field value for "X-Runner-Token""
> > The problem: The runner token (some-secret) is being written to /tmp/runner-token and passed via --token-url file:///tmp/runner-token, but the forgejo-runner binary is reading that token and sending it as an HTTP header value. The string some-secret contains characters (or possibly a trailing newline from echo) that are invalid in HTTP headers.
> > The most likely culprit is that echo 'some-secret' appends a newline character (\n) to the file, and when the runner reads it and stuffs it into the X-Runner-Token header, HTTP rejects it because newlines are illegal in header values.
> > The fix in the entrypoint script should use printf instead of echo to avoid the trailing newline:
> > bashprintf '%s' 'some-secret' > /tmp/runner-token
> > or with echo -n:
> > bashecho -n 'some-secret' > /tmp/runner-token
> > This is a common gotcha when writing secrets to files for token-URL consumption — the reader passes the raw file contents verbatim into the HTTP header, newline included, which fails RFC 7230 header validation.
> > The test value some-secret itself is fine as a string; it's purely the newline that echo adds that breaks it.

Irrelevant. It fails because there is no forgejo instance to test against.

PR #4915, opened 2026-04-30
Fix Dashy config save - use fixed UID 1000 matching upstream image

### LINKED ISSUE #4911: [Bug] Dashy cant save configuration
The new versions of Dashy on Truenas (4.0.2 in my case) fail when saving the configuration. This is because the Dashy container now runs with UID 1000, making it impossible to edit the files for UID app 568.

More info:

https://github.com/Lissy93/dashy/issues/2096

Regards

blka:
## Summary

Fixes #4911

Dashy v4.x changed its container to run as `node` (UID 1000, GID 1000), but the TrueNAS app was forcing UID 568 via `set_user()`. This caused permission mismatches — Dashy couldn't write config files or create backups on the mounted volume.

**Changes:**
- Use fixed UID 1000/GID 1000 in `ix_values.yaml` consts (matching upstream image's `node` user)
- Update `docker-compose.yaml` to use `values.consts.user_id/group_id` instead of `values.run_as.user/group`
- Update `app.yaml` `run_as_context` to document UID 1000/GID 1000
- Remove `run_as` section from `questions.yaml` (UID is no longer user-configurable)
- Add `remove_run_as_migration` for existing installations
- Bump version to 2.0.3

**Why UID 1000:** Dashy's upstream image (`lissy93/dashy:4.0.3`) is based on `node:22-alpine` with `USER node` (UID 1000/GID 1000). The image has no support for `USER_UID`/`USER_GID` env vars, so the container must run as UID 1000 to match internal file ownership. This follows the same pattern used by drawio, node-red, briefkasten, and other fixed-UID apps in the catalog.

## Test plan

- [ ] Install Dashy from this branch
- [ ] Verify the container runs as UID 1000/GID 1000
- [ ] Verify Dashy can save configuration through the UI
- [ ] Verify config backups are created successfully
- [ ] Verify existing installations migrate correctly (run_as removed from stored config)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/ix-dev/community/dashy/app.yaml
+++ b/ix-dev/community/dashy/app.yaml
@@ -20,11 +20,11 @@ maintainers:
   url: https://www.truenas.com/
 name: dashy
 run_as_context:
-- description: Container [dashy] can run as any non-root user and group.
-  gid: 568
-  group_name: Host group is [apps]
-  uid: 568
-  user_name: Host user is [apps]
+- description: Container [dashy] runs as non-root user and group.
+  gid: 1000
+  group_name: Host group is [unknown (1000)]
+  uid: 1000
+  user_name: Host user is [unknown (1000)]
 screenshots:
 - https://media.sys.truenas.net/apps/dashy/screenshots/screenshot1.gif
 - https://media.sys.truenas.net/apps/dashy/screenshots/screenshot2.gif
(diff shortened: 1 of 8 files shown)


stavros-k:
The user directive is working. You just need to ensure the perms are correct on the given directory or if you dont want to use 568, you are free to select any user in the form

image

PR #4919, opened 2026-05-01
Monerod: add question to enable MRL banlist

Keeqler:


ix-dev/community/monerod/templates/macros/banlist-setup.sh
stavros-k, line 1: I'm not sure I like this. Installing packages on every app restart and cloning a repo as well...
The repo does not seem to be updated frequently, maybe we can bake the txt file in the image and just add the flag here? (So it can be pushed as update)
Alternative, just have couple lines in the description so users can mount the txt file where needed, and they can have a cron on their host to update that file?

What do you think?

stavros-k (commented):
.

ix-dev/community/monerod/templates/macros/banlist-setup.sh
Keeqler: It turns out that the image we're using already includes the banlist. Very dumb of me. lol

stavros-k (approved):

Issue #4408, opened 2026-02-20
[App Request]: Caddy


systemdbrew:
### App Name

caddy

### App Description

Caddy 2 is a powerful, enterprise-ready, open source web server with automatic HTTPS written in Go.

### App Website / Github Repo

https://github.com/caddyserver/caddy

### Docker Installation Documentation

https://hub.docker.com/_/caddy

### App Icon URL

https://hub.docker.com/api/media/repos_logo/v1/library%2Fcaddy?type=logo

### Docker Image

https://hub.docker.com/_/caddy

### App Category

Web Server

### Requirements Checklist

- [x] The app is actively maintained
- [x] The app has official Docker images available
- [x] The app has versioned image tags available
- [x] The app is open source or has a free tier
- [x] I have tested the Docker installation locally

### Additional Information

Great web server for hosting files without credentials. Great for repos and CDN's. Also can be used as a reverse proxy, it is very small and lightweight.

joelhickok:
Bump.  I second the request.  New Truenas user here, so not sure of the process.

PR #4935, opened 2026-05-05
Add Caddy

amateescu:
# App Addition

- [x] There's an open [issue](https://github.com/truenas/apps/issues/4408) to discuss this app addition before submitting this pull request.

# AI

## Description

Adds Caddy to the community train.

[Caddy](https://caddyserver.com/) is a fast, extensible, open-source web server written in Go, with automatic HTTPS via Let's Encrypt. It can serve as a static file server, reverse proxy, or HTTP/3-capable web server, and is configured via the Caddyfile syntax.

## App Information

- **Upstream**: https://github.com/caddyserver/caddy
- **Documentation**: https://caddyserver.com/docs/
- **Docker image**: https://hub.docker.com/_/caddy
- **App Version**: 2.11.2

## Testing

Tested locally with:

- [x] basic-values.yaml
- [x] test-values.yaml (see below)

Additionally tested on a real TrueNAS SCALE instance with `Install via YAML`:

- Rootless `568:568` startup with `NET_BIND_SERVICE` for ports 80 / 443.
- Permissions container chowning the bind-mounted datasets to the apps user.
- Default Caddyfile seeded on first run; user edits preserved across restarts.
- Healthcheck (wget against the admin API at `/config/`) reporting healthy.
- End-to-end reverse-proxy scenario: replaced the seeded Caddyfile with `https://caddy.local { tls internal; reverse_proxy 192.168.x.x:30116 }`, restarted the app; Caddy issued a leaf cert via its local CA and proxied HTTPS traffic to a LAN service successfully.

## Icons and Screenshots

- Icon: https://caddyserver.com/resources/images/logo-dark.svg

## Checklist

- [x] README.md included

### DIFF
--- a/ix-dev/community/caddy/README.md
+++ b/ix-dev/community/caddy/README.md
@@ -1,0 +1,3 @@
+# Caddy
+
+[Caddy](https://caddyserver.com/) is a powerful, extensible platform to serve your sites, services, and apps with automatic HTTPS.
--- a/ix-dev/community/caddy/app.yaml
+++ b/ix-dev/community/caddy/app.yaml
@@ -1,0 +1,38 @@
+app_version: 2.11.2
+capabilities:
+- description: Caddy is able to bind to privileged ports (< 1024)
+  name: NET_BIND_SERVICE
+categories:
+- networking
+changelog_url: https://github.com/caddyserver/caddy/releases
+date_added: '2026-05-05'
+description: Fast and extensible web server with automatic HTTPS
+home: https://caddyserver.com/
+host_mounts: []
+icon: https://media.sys.truenas.net/apps/caddy/icons/icon.svg
+keywords:
+- web
+- server
+- reverse-proxy
+- https
+lib_version: 2.3.5
+lib_version_hash: c0d042e3de6350fae9aee546d3c79bc0accf1f2011da89cc1ebff29421aa7fc9
+maintainers:
+- email: dev@truenas.com
+  name: truenas
+  url: https://www.truenas.com/
+name: caddy
+run_as_context:
+- description: Container [caddy] can run as any non-root user and group.
+  gid: 568
+  group_name: Host group is [apps]
+  uid: 568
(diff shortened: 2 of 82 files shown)


stavros-k:
Sorry but this is not viable.
The moment someone uses a custom config the healthcheck will fail. Ports is now a missmatch. And the whole template makes a lot of assumptions.

Caddy is a highly configurable thing but only via config file.
So we can't really force a few things needed for the app to function correctly WITHOUT making it impossible for the user to configure it. (unlike traefik).

Also caddy uses a lot of plugins for most of the things, this is impossible to do here.

I'll close this for now.

Thanks

amateescu:
@stavros-k thanks for reviewing! All three are good points, and I think two of them can be easily fixed:

For the healthcheck, we can switch to `pgrep` on the caddy process.

For the assumptions made in the template, we can remove `http_port` and `https_port`, and just let users match their Caddyfile configuration via additional_ports. We can prefill it with one entry mapping the host port to `container_port: 80`, matching the seeded default Caddyfile, so the first-run experience still works.

The only real problem is plugin support, and for that the only realistic path is to instruct users to convert the catalog app to a custom app, where they can specify their own pre-built image (e.g., a `caddy:builder`-based variant with `caddy-dns/cloudflare`).

I personally hit this because I need DNS-01 with Cloudflare for my own setup, so even the PR I submitted wouldn't have served my actual use case, but decided to post it anyway so the work wouldn't be entirely wasted.

Let me know if the third point is acceptable so I can update the PR. If not, [#4408](https://github.com/truenas/apps/issues/4408) should be closed as well :)

stavros-k:
Caddy to be used even for the "most popular" cases, needs plugin. I dont see a point of having an app that needs to be 98% of the time to custom to do anything.
I think it should be benched for now. I do want to add it, but I need to figure out a nice way to make it work all around. Or at least for the most common things. (Thats why I left the request open).
It's not a "wont do". Its a "needs planning".

PR #4939, opened 2026-05-06
paperless: allow per container envs

### LINKED ISSUE #4938: [Enhancement]: Env variables for gotenberg in paperless-ngx app
### App Name

paperless-ngx

1.6.21

### Enhancement Description

Any additional env variables entered are currently only available in the main paperless container as seen in the compose file:
https://github.com/truenas/apps/blob/fa48c6d48c50a953553b8c6734ddb542c759a18e/ix-dev/community/paperless-ngx/templates/docker-compose.yaml#L57

There should be a way to set variables for the other containers as well.

### Use Case

I'd like to be able to send the API_TIMEOUT variable to Gotenberg as I need to increase it from the default 30s.

### Proposed Solution

As I see it there are 3 ways for this:

1.  Apply all env variables to both containers
2. Add a new question for any amount of env variables and apply them to gotenberg
3. Add a new question specifically for the timeout and set that variable for gotenberg specifically.

Is there a preference here to keep consistancy with other apps?

### Alternatives Considered

_No response_

### Additional Context

_No response_

stavros-k:
Closes #4938

### DIFF
--- a/ix-dev/community/paperless-ngx/app.yaml
+++ b/ix-dev/community/paperless-ngx/app.yaml
@@ -71,4 +71,4 @@ sources:
 - https://github.com/paperless-ngx/paperless-ngx
 title: Paperless-ngx
 train: community
-version: 1.6.21
+version: 1.6.22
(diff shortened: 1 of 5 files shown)


Qubad786 (approved):

PhilippAuch:
Thank you very much @stavros-k  for the incredibly fast help :)

PR #4944, opened 2026-05-07
Update docker-compose.yaml

Detrail:
Changed address schema from uri to string in order to stop it from stripping the port when entering the netbird address.
If left as a uri it strips the :443 port and would result in a "netbird missing port in address" error. 
It appears netbird server needs it to be explicitly listed after the url even though :443 is a default https port.

# Pull Request

⚠️⚠️⚠️ **BEFORE CREATING THE PR** ⚠️⚠️⚠️

Please go to the **Preview** tab and select the appropriate template: 

* [🚀 App Addition](?expand=1&template=app_addition.md)


ix-dev/community/netbird-server/questions.yaml
```diff
@@ -36,7 +36,7 @@ questions:
             This should be in the format of [protocol]://[domain]:[port], e.g. https://netbird-mgmt.example.com:443</br>
             The protocol must https. A reverse proxy must be configured to handle TLS termination and forward traffic to NetBird.</br>
           schema:
-            type: uri
+            type: string
```
stavros-k: the type does not strip it afik.But the template definitely does. But I can't remember why I had to strip it. Can you take a look?

Detrail: Ah I see, I was under the assumption the normalization done by uri was stripping the :443 port.
But it does look like line 15 of the template explicitly strips it "{% set netbird_address = netbird_address.removesuffix(":443") %}" 

I'll test changing this later and get back to you.

Detrail: I wasn't able to find any issues when restoring :443 to the netbird_address by removing the stripping function. So I guess its good probably?

stavros-k: we should revert to uri here tho

stavros-k:
please update the PR title as well and make sure CI is green.

stavros-k (approved):

Issue #1908, opened 2025-03-21
Immich ROCM Machine Learning support


Anulo2:
Hey there, Immich as recently added the ability to use ROCM as an acceleration api (https://github.com/immich-app/immich/blob/main/machine-learning/README.md?plain=1#L10), woould be awesome to be able to set in in the app options, thanks for the great work!

leovanalphen:
How can I use this with truenas? I've got Immich running with the rocm ml image, but the container keeps crashing:
HIP failure 100: no ROCm-capable device is detected

In the app settings I've got passthrough GPU's checked. And the GPU is isolated in truenas settings (7800xt). I thought you also need the ROCM and AMD drivers on the host, but AFAIK those are not present in truenas?

stavros-k:
You shouldnt isolate the gpu to use it for apps.
isolation is only for VM passthrough.

That being said, the `kdf` device is not currently getting passed on the container, but it will start doing that once 25.04 is released. Middleware code is already added on that release.

leovanalphen:
Hi Stavros,

Thanks for your reply! I managed to get it working last night, switched between 24.10 and 25.4 a couple of times before I figured out the gpu isolation was the culprit 😅

Currently running 25.4. rc1 The immich ML container with rocm now works and I got ComfyUI working with rocm as well (in an incus system container).

Happy with the outcome 👍

Thanks again!

Sent from Proton Mail Android

-------- Original Message --------

Issue #942, opened 2024-11-16
Immich Cuda ML doesn't work after upgrade to 24.10


Karmalakas:
I'm experiencing issue with Machine Learning on Immich with nVidia GPU. It worked perfectly on 24.04, but not any more. Issue seems pretty much the same as @Fireflaker [reported on charts repo](https://github.com/truenas/charts/issues/2891)

Immich server log is flooded with:
```
2024-11-16 10:44:02.863611+00:00[31m[Nest] 7  - [39m11/16/2024, 12:44:02 PM [31m  ERROR[39m [33m[Microservices:JobService][39m [31mUnable to run job handler (smartSearch/smart-search): Error: Machine learning request to "http://machine-learning:32002" failed with SocketError: other side closed[39m
2024-11-16 10:44:02.863694+00:00[31m[Nest] 7  - [39m11/16/2024, 12:44:02 PM [31m  ERROR[39m [33m[Microservices:JobService][39m [31mError: Machine learning request to "http://machine-learning:32002" failed with SocketError: other side closed
2024-11-16 10:44:02.863704+00:00at /usr/src/app/dist/repositories/machine-learning.repository.js:18:19
2024-11-16 10:44:02.863710+00:00at async MachineLearningRepository.predict (/usr/src/app/dist/repositories/machine-learning.repository.js:17:21)
2024-11-16 10:44:02.863717+00:00at async MachineLearningRepository.encodeImage (/usr/src/app/dist/repositories/machine-learning.repository.js:41:26)
2024-11-16 10:44:02.863722+00:00at async SmartInfoService.handleEncodeClip (/usr/src/app/dist/services/smart-info.service.js:104:27)
2024-11-16 10:44:02.863815+00:00at async JobService.onJobStart (/usr/src/app/dist/services/job.service.js:151:28)
2024-11-16 10:44:02.863825+00:00at async EventRepository.onEvent (/usr/src/app/dist/repositories/event.repository.js:122:13)
2024-11-16 10:44:02.863831+00:00at async Worker.processJob (/usr/src/app/node_modules/bullmq/dist/cjs/classes/worker.js:394:28)
2024-11-16 10:44:02.863836+00:00at async Worker.retryIfFailed (/usr/src/app/node_modules/bullmq/dist/cjs/classes/worker.js:581:24)[39m
[...]

PR #4970, opened 2026-05-12
immich: clarify ML image type choice with GPU-vendor enum labels

eugenelim:
The Machine Learning Image Type dropdown doesn't tell you which option matches which GPU vendor — users enable passthrough, leave the image on CPU, then wonder why the GPU isn't being used.

Rewrites the description and enum labels to be vendor-explicit, and hides the question when ML is disabled.

Enum values are unchanged, so existing selections still resolve.

Related #1908 #942


stavros-k:
Can you please dont let AI write a book long description for such change?
Description not only adds zero value but makes any reviewer spent 10x the time needed for review.

eugenelim:
Noted, will have to the point PR descriptions going forward. Hopefully this is a simple change.

stavros-k (approved):

stavros-k:
> Related #1908 #942

How are those related? 
This feels like an AI is forced to find something to "fix" for the sake of submitting _something_

eugenelim:
I'm taking a user first approach in investigating Immich friction points in particular, to reduce the amount of technical hands on and troubleshooting that needs to happen. 

Can we take that lens? The change is small, but doesn't it materially improve from a user understanding perspective?

stavros-k:
Sorry not going to keep interacting with an LLM.
At this point is very clear that there is no user answering here, but an LLM.

eugenelim:
I think we got off on the wrong foot. I'm actively looking to help. is there a way we can talk directly on where I'm coming from and the overall ideas before you just blindly reject PRs? there is a lot of user friction today compared to another vendor's solution, and I want to help fix it while also being respectful of maintainer's limited time.

PR #4993, opened 2026-05-16
Add Seafile app with optional SeaDoc

### LINKED ISSUE #260: Community Chart Suggestion: Seafile
File sync and share platform like NextCloud, however Seafile is more stable and reliable compared to NextCloud. It is a high performance cloud storage system that has low hardware utilization compared to NextCloud.

Currently there is no alternative to NextCloud for a more simplistic cloud storage system. Having at least one alternative in this case would be ideal.

Here is the seafile docker: [https://github.com/haiwen/seafile-docker](https://github.com/haiwen/seafile-docker)

Is there a way I could directly contribute to accelerate the process of adding this app into TrueNas? Feel free to let me know I would like to contribute!

mreid-tt:
# App Addition

- [x] I have opened an [issue](https://github.com/truenas/apps/issues/260) to discuss this app addition before submitting this pull request.

# AI

## Description

Adds Seafile to the community train.

Seafile is an open source cloud storage system with privacy preservation and teamwork features. This package includes:

- Core Seafile server (`seafileltd/seafile-mc:13.0.24`) with MariaDB (`mariadb:12.3.2`) and Valkey (`valkey/valkey:9.1.0`)
- Nginx reverse proxy (`nginxinc/nginx-unprivileged:1.31.1`) with optional TrueNAS certificate support for HTTPS
- Optional SeaDoc server (`seafileltd/sdoc-server:2.0.9`) for collaborative document editing

## App Information

- **Upstream**: https://www.seafile.com/
- **Documentation**: https://manual.seafile.com/
- **Docker Image**: https://hub.docker.com/r/seafileltd/seafile-mc
- **App Version**: 13.0.24

## Testing

Tested locally with:

- [x] basic-values.yaml
- [x] seadoc-enabled.yaml
- [x] all-features-enabled.yaml
- [x] custom-storage.yaml
- [x] https-values.yaml

All tests passed successfully (render and deployment).

## Icons and Screenshots

- Icon: https://cdn.jsdelivr.net/gh/selfhst/icons@main/svg/seafile.svg

## Special Notes

- The seafile-mc image uses baseimage-docker's `my_init` as PID 1, which requires root. The container runs as root; the init system drops privileges for seafile processes internally.
- Requires `SETUID`/`SETGID` capabilities. Applied via `cap_add` since the library defaults to `cap_drop: [ALL]`.
- SeaDoc server is conditional — when `enable_seadoc` is true, the sdoc-server container is deployed alongside seafile.
- Nginx (`nginxinc/nginx-unprivileged`) serves as the reverse proxy with dynamic port mapping (ports >= 1024), eliminating the need for `NET_BIND_SERVICE` capability. Nginx config is injected via Docker configs.
- HTTPS via TrueNAS certificate manager — selecting a certificate enables TLS on a separate HTTPS port (`30441`) with HTTP to HTTPS redirect on `30440`.
- All image tags are pinned to concrete versions.
- Tested on Ubuntu 22.04 VM with Docker Compose. All five test configurations deployed successfully with web UI accessible.

## Checklist

- [x] README.md included
- [x] Multiple test scenarios tested (5 test files)


mreid-tt:
Hey @stavros-k, hope you’re doing well. If there are any recommended changes or suggested next steps, please let me know. I’m very eager to help get this into the catalog.

ix-dev/community/seafile/templates/docker-compose.yaml
```diff
@@ -0,0 +1,154 @@
...
+  {% do caddyfile_lines.append("    }") %}
+{% endif %}
+{% do caddyfile_lines.append("    reverse_proxy " ~ values.consts.seafile_container_name ~ ":80") %}
+{% do caddyfile_lines.append("}") %}
+{% set caddyfile_content = caddyfile_lines | join("\n") %}
+
+{% set caddy = tpl.add_container(values.consts.caddy_container_name, "caddy_image") %}
+{% do caddy.add_network(seafile_net) %}
+{% do caddy.set_user(values.run_as.user, values.run_as.group) %}
+{% do caddy.add_caps(["NET_BIND_SERVICE"]) %}
```
stavros-k: The caddy file that seafile docs suggest is just an example. Users can and probably will want to customize it by a lot.
Also if they have already a reverse proxy in their network, they probably dont want another one. 

If seafile is added, it wont be with caddy.

Still this is a very quick review, I have not checked anything else, but this is a blocker.

mreid-tt: Thanks for the quick review, appreciate you taking the time.

You raise a good point about Caddy. My understanding is that the upstream uses it for path-based routing to SeaDoc (`/sdoc-server/`), and without it I'm not sure how SeaDoc would be reachable within the isolated network of a TrueNAS app. But I may be missing something.

FWIW, I noticed Windmill in the community train also bundles Caddy, so there's some precedent, though I don't know if that's considered a different scenario.

If you think there's a cleaner approach that avoids bundling Caddy, I'm happy to follow your guidance. Otherwise, if Caddy is a hard no, we can drop SeaDoc from the app and keep it simpler. Let me know what you'd prefer.

Krysztal: @mreid-tt as I see SeaDoc does not have to be /sdoc-server/
you can configure it as SEADOC_SERVER_URL without /sdoc-server/

Krysztal: @mreid-tt if you want I can share my test custom app yaml for truenas with seafile without Caddy

mreid-tt: Hey @stavros-k, just following up to see if you've had a chance to complete your review of this PR. I was also wondering if you've had an opportunity to consider my response and the question I raised. Please let me know if there's any additional information I can provide.

mreid-tt:
Hey @stavros-k, following up on the feedback about Caddy. I've replaced it with nginx and addressed the other review points. Here's a summary of the changes:

### What changed

- **Caddy removed, replaced with nginx** (`nginxinc/nginx-unprivileged:1.31.1`) — same pattern used by wger, zammad, and other community apps
- **No elevated capabilities** — nginx listens on dynamic ports (>= 1024), no `NET_BIND_SERVICE` needed
- **All image tags pinned** to concrete versions (was using rolling tags like `13.0-latest`, `7-alpine`):
  - seafile: `13.0.24`
  - sdoc-server: `2.0.9`
  - mariadb: `10.11.18`
  - redis: `7.4.9-alpine`
- **Ports updated** to avoid CI conflicts: web_port `30440`, https_port `30441`
- **Bug fixes** found during testing:
  - `ENABLE_SEADOC` now explicitly set to `"false"` when disabled (image defaulted to enabled)
  - `extra_hosts` on seafile container so its backend can resolve `SEADOC_SERVER_URL` internally
  - Host header uses `$http_host` (includes port) — fixed Django CSRF failure
  - Healthcheck uses `/health` endpoint — works in both HTTP and HTTPS modes

### Testing

All five test value files pass render and deploy successfully. HTTP and HTTP+SeaDoc tested manually — login, document editing, and revision workflows verified.

Would appreciate another review when you get a chance.

ix-dev/community/seafile/templates/docker-compose.yaml
mreid-tt: > @mreid-tt as I see SeaDoc does not have to be /sdoc-server/ you can configure it as SEADOC_SERVER_URL without /sdoc-server/

Thanks for the suggestion @Krysztal. I explored removing the `/sdoc-server/` path prefix, but decided to keep it for a few reasons:

1. The upstream documentation and default deployment use `/sdoc-server/` as the standard path
2. With the nginx reverse proxy, path-based routing to the seadoc container is clean and well-tested
3. The nginx config follows the reference config from the upstream "Use other reverse proxy" docs, which also uses `/sdoc-server/`
4. Keeping the upstream default makes it easier for users to follow upstream troubleshooting guides

That said, I confirmed during testing that HTTP+SeaDoc works correctly — document editing and revision workflows both pass with the `/sdoc-server/` path.

ix-dev/community/seafile/templates/docker-compose.yaml
```diff
@@ -0,0 +1,131 @@
...
+  "volume": values.storage.redis_data,
+} %}
+{% set redis = tpl.deps.redis(values.consts.redis_container_name, "redis_image", redis_config, perm_container) %}
+{% do redis.container.add_network(seafile_net) %}
+
+{% set seafile = tpl.add_container(values.consts.seafile_container_name, "image") %}
+{% do seafile.add_network(seafile_net) %}
+{% do seafile.add_caps(["CHOWN", "DAC_OVERRIDE", "FOWNER", "SETGID", "SETUID"]) %}
+{% do seafile.healthcheck.set_test("curl", {"port": 80, "path": "/"}) %}
+{% do seafile.healthcheck.set_timeout(10) %}
```
stavros-k: Why do we change the timings in healthchecks here?

ix-dev/community/seafile/templates/docker-compose.yaml
```diff
@@ -0,0 +1,131 @@
...
+{% do seafile.healthcheck.set_retries(3) %}
+{% do seafile.healthcheck.set_start_period(10) %}
+{% do seafile.environment.add_env("SEAFILE_MYSQL_DB_HOST", values.consts.mariadb_container_name) %}
+{% do seafile.environment.add_env("SEAFILE_MYSQL_DB_PORT", values.consts.mariadb_port) %}
+{% do seafile.environment.add_env("SEAFILE_MYSQL_DB_USER", values.consts.db_user) %}
+{% do seafile.environment.add_env("SEAFILE_MYSQL_DB_PASSWORD", values.database.db_password) %}
```
stavros-k: why do we have the database section when all other apps use different pattern?

ix-dev/community/seafile/templates/docker-compose.yaml
```diff
@@ -0,0 +1,131 @@
...
+{% do seafile.environment.add_env("INIT_SEAFILE_MYSQL_ROOT_PASSWORD", values.database.db_root_password) %}
+{% do seafile.environment.add_env("SEAFILE_MYSQL_DB_CCNET_DB_NAME", "ccnet_db") %}
+{% do seafile.environment.add_env("SEAFILE_MYSQL_DB_SEAFILE_DB_NAME", "seafile_db") %}
+{% do seafile.environment.add_env("SEAFILE_MYSQL_DB_SEAHUB_DB_NAME", "seahub_db") %}
+{% set seafile_proto = "https" if values.network.certificate_id else "http" %}
+{% set seafile_port = values.network.https_port.port_number if values.network.certificate_id else values.network.web_port.port_number %}
```
stavros-k: why we want 2 ports if the app only listens at one a time?

ix-dev/community/seafile/templates/docker-compose.yaml
```diff
@@ -0,0 +1,131 @@
...
+{% set default_port = 443 if values.network.certificate_id else 80 %}
+{% set seafile_hostname = values.seafile.server_hostname ~ (":" ~ seafile_port if seafile_port != default_port else "") %}
```
stavros-k: why not always append the :443 or :80? 
We can also just strip these 2 suffixes without having to do all these checks.

ix-dev/community/seafile/templates/docker-compose.yaml
```diff
@@ -0,0 +1,131 @@
...
+{% do seafile.environment.add_env("SEAFILE_SERVER_HOSTNAME", seafile_hostname) %}
+{% do seafile.environment.add_env("SEAFILE_SERVER_PROTOCOL", seafile_proto) %}
+{% do seafile.environment.add_env("INIT_SEAFILE_ADMIN_EMAIL", values.seafile.admin_email) %}
+{% do seafile.environment.add_env("INIT_SEAFILE_ADMIN_PASSWORD", values.seafile.admin_password) %}
+{% do seafile.environment.add_env("TIME_ZONE", values.TZ) %}
+{% do seafile.environment.add_env("CACHE_PROVIDER", "redis") %}
+{% do seafile.environment.add_env("REDIS_HOST", values.consts.redis_container_name) %}
+{% do seafile.environment.add_env("REDIS_PORT", values.consts.redis_port) %}
+{% do seafile.environment.add_env("REDIS_PASSWORD", values.database.redis_password) %}
+{% do seafile.environment.add_env("ENABLE_GO_FILESERVER", "true") %}
```
stavros-k: bool values are also working no need to make them strings

ix-dev/community/seafile/templates/docker-compose.yaml
```diff
@@ -0,0 +1,131 @@
...
+  {% set seadoc = tpl.add_container(values.consts.seadoc_container_name, "seadoc_image") %}
+  {% do seadoc.add_network(seafile_net) %}
+  {% do seadoc.set_user(0, 0) %}
+  {% do seadoc.add_caps(["CHOWN", "DAC_OVERRIDE", "FOWNER", "SETGID", "SETUID"]) %}
+  {% do seadoc.add_storage("/shared", values.storage.seadoc_data) %}
+  {% do seadoc.environment.add_env("DB_HOST", values.consts.mariadb_container_name) %}
+  {% do seadoc.environment.add_env("DB_PORT", values.consts.mariadb_port) %}
+  {% do seadoc.environment.add_env("DB_USER", values.consts.db_user) %}
+  {% do seadoc.environment.add_env("DB_PASSWORD", values.database.db_password) %}
+  {% do seadoc.environment.add_env("DB_NAME", "seahub_db") %}
```
stavros-k, line 55: why does this connect to a different database that the one that we create?

ix-dev/community/seafile/templates/docker-compose.yaml
```diff
@@ -0,0 +1,131 @@
...
+
+{% do seafile.environment.add_env("ENABLE_SEADOC", "true" if values.seafile.enable_seadoc else "false") %}
+{% if values.seafile.enable_seadoc %}
+  {% do seafile.environment.add_env("SEADOC_SERVER_URL", seafile_proto + "://" + seafile_hostname + "/sdoc-server") %}
+{% endif %}
+
+{% do seafile.environment.add_user_envs(values.seafile.additional_envs) %}
+{% do seafile.add_extra_host(values.seafile.server_hostname, "host-gateway") %}
+
+{% do seafile.add_storage("/shared", values.storage.data) %}
```
stavros-k, line 112: what does it share with seadoc?

ix-dev/community/seafile/ix_values.yaml
```diff
@@ -0,0 +1,37 @@
+images:
+  image:
+    repository: seafileltd/seafile-mc
+    tag: 13.0.24
+  mariadb_image:
+    repository: mariadb
+    tag: 10.11.18
+  redis_image:
+    repository: redis
```
stavros-k: why do we use redis while ALL other apps use valkey?

ix-dev/community/seafile/ix_values.yaml
```diff
@@ -0,0 +1,37 @@
...
+    tag: 1.31.1
+  seadoc_image:
+    repository: seafileltd/sdoc-server
+    tag: 2.0.9
+  container_utils_image:
+    repository: ixsystems/container-utils
+    tag: 1.0.2
+
+consts:
+  seafile_container_name: seafile
```
stavros-k, line 22: why do all containers get a seafile suffix? this will generate a weirdly repeated container name

ix-dev/community/seafile/ix_values.yaml
```diff
@@ -0,0 +1,37 @@
...
+  mariadb_container_name: seafile-db
+  redis_container_name: redis
+  nginx_container_name: seafile-nginx
+  seadoc_container_name: seafile-seadoc
+  perms_container_name: permissions
+  mariadb_port: 3306
```
stavros-k: why are these consts here?

ix-dev/community/seafile/templates/docker-compose.yaml
```diff
@@ -0,0 +1,131 @@
...
+  {% do seadoc.environment.add_env("DB_PASSWORD", values.database.db_password) %}
+  {% do seadoc.environment.add_env("DB_NAME", "seahub_db") %}
+  {% do seadoc.environment.add_env("TIME_ZONE", values.TZ) %}
+  {% do seadoc.environment.add_env("JWT_PRIVATE_KEY", values.seafile.jwt_private_key) %}
+  {% do seadoc.environment.add_env("SEAHUB_SERVICE_URL", "http://" + values.consts.seafile_container_name) %}
+  {% do seadoc.add_extra_host(values.seafile.server_hostname, "host-gateway") %}
+  {% do seadoc.environment.add_env("NON_ROOT", "false") %}
+  {% do seadoc.healthcheck.set_test("tcp", {"port": 80}) %}
+  {% do seadoc.depends.add_dependency(values.consts.mariadb_container_name, "service_healthy") %}
+  {% do perm_container.add_or_skip_action("seadoc_data", values.storage.seadoc_data, {"uid": 0, "gid": 0, "mode": "check"}) %}
```
stavros-k: why do we need permissions action here if it runs as root?

ix-dev/community/seafile/templates/docker-compose.yaml
```diff
@@ -0,0 +1,131 @@
+{% from "macros/nginx.conf" import nginx_config %}
+{% set tpl = ix_lib.base.render.Render(values) %}
+
+{% if not values.seafile.server_hostname %}
```
stavros-k: isnt this enforced by the questions schema?

stavros-k:
Why do we ship seadoc with seafile? Isn't that a compatible WOPI server? 
ie like collabora? Shouldnt we have it separately?

ix-dev/community/seafile/ix_values.yaml
```diff
@@ -0,0 +1,37 @@
+images:
+  image:
+    repository: seafileltd/seafile-mc
+    tag: 13.0.24
+  mariadb_image:
+    repository: mariadb
+    tag: 10.11.18
```
stavros-k: why is mariadb 2 major versions behind?

ix-dev/community/seafile/questions.yaml
```diff
@@ -0,0 +1,867 @@
...
+    label: ""
+    group: Seafile Configuration
+    schema:
+      type: dict
+      attrs:
+        - variable: server_hostname
+          label: Server Hostname
+          description: |
+            The hostname or IP address where Seafile will be accessible.</br>
+            Do not include the port — it will be added automatically based on the network configuration.</br>
```
stavros-k: That's going to cause issues if you construct it from network config.
One can have it listen on port 1234 on LAN, but 443 on external. 
Your method will pin the hostname to xxx:1234 and will be invalid when accessed from outside.

ix-dev/community/seafile/questions.yaml
```diff
@@ -0,0 +1,867 @@
...
+                schema:
+                  type: int
+                  default: 2
+                  required: true
+              - variable: memory
+                label: Memory (in MB)
+                description: Memory limit for Seafile.
+                schema:
+                  type: int
+                  default: 2048
```
stavros-k: Why is this default different from ALL other apps?

mreid-tt:
> Why do we ship seadoc with seafile? Isn't that a compatible WOPI server?
> ie like collabora? Shouldnt we have it separately?

SeaDoc is included as part of Seafile because it's Seafile-specific — it shares Seafile's MariaDB database, requires Seafile's JWT secret, and fetches document content through Seafile's HTTP API. It cannot function without a Seafile server. The upstream docs treat it the same way, listing `seadoc.yml` alongside `seafile-server.yml` in the default `COMPOSE_FILE`.

That said, if you feel the added complexity isn't worth it, happy to drop SeaDoc and keep the app simpler. We'll defer to your judgment.

ix-dev/community/seafile/templates/docker-compose.yaml
mreid-tt: We mirrored the values from the upstream `seafile-server.yml` which sets `timeout: 10s, retries: 3, start_period: 10s`. Removed those overrides to use the library defaults. Kept only `set_start_period(120)` because on fresh install, Seafile creates MySQL databases and runs Django migrations which takes 60-90s — without an extended start period the container is flagged unhealthy before setup completes.

ix-dev/community/seafile/templates/docker-compose.yaml
mreid-tt: Agreed. Moved `db_root_password`, `db_password`, and `redis_password` into the main `seafile:` dict, matching how other apps handle these. Removed the separate `database:` section and group from questions.yaml.

ix-dev/community/seafile/templates/docker-compose.yaml
mreid-tt: This follows the standard pattern used by all apps that support HTTPS certificates in the catalog (~24 apps). Only one is active at a time: with a certificate selected, `web_port` redirects HTTP→HTTPS and `https_port` serves TLS; without one, only `web_port` is published.

ix-dev/community/seafile/templates/docker-compose.yaml
mreid-tt: Simplified to always append the port from network config. The user provides a bare hostname and the template adds the port (e.g., `hostname:30440`). No conditional logic.

ix-dev/community/seafile/templates/docker-compose.yaml
mreid-tt: Changed to native bools. Also made `ENABLE_SEADOC` explicitly set on every startup — previously it was only set when `true`, and the seafile-mc image defaults to enabled, which caused the UI to show SeaDoc features for users who didn't enable it.

ix-dev/community/seafile/templates/docker-compose.yaml
mreid-tt: SeaDoc connects to `seahub_db` — the upstream `seadoc.yml` for version 13.0 uses `SEAFILE_MYSQL_DB_SEAHUB_DB_NAME` (defaults to `seahub_db`) by design. SeaDoc is an extension of Seahub (the web interface), not the core file server, so it uses the seahub database.

ix-dev/community/seafile/templates/docker-compose.yaml
mreid-tt: The seafile and seadoc containers both mount storage at `/shared`, but they map to different host paths (`values.storage.data` and `values.storage.seadoc_data` respectively). They don't share data — the paths just happen to have the same internal container name.

ix-dev/community/seafile/ix_values.yaml
mreid-tt: Switched to `valkey/valkey:9.1.0`. Had to add `--stop-writes-on-bgsave-error no` because Seafile's redis client triggers BGSAVE on first session write, and valkey's default blocks writes when RDB persistence encounters issues — this caused a 500 error on the login page on first access. Verified working with this change.

ix-dev/community/seafile/ix_values.yaml
mreid-tt: Good point. Renamed to bare names: `seafile-db` → `db` (matching the upstream `seafile-server.yml` service name), `seafile-nginx` → `nginx`, `seafile-seadoc` → `seadoc`. This matches the convention used by wger, n8n, and others.

ix-dev/community/seafile/ix_values.yaml
mreid-tt: Container name consts are defined in every community app and are used across the template (env vars, depends_on, networks). Removed `redis_port` and `mariadb_port` consts — those are just standard ports and are now written directly where they're used, matching how other apps handle them.

ix-dev/community/seafile/templates/docker-compose.yaml
mreid-tt: Removed the permissions action for seadoc_data. Since the seadoc container runs as root, it can create files with any ownership it needs.

ix-dev/community/seafile/templates/docker-compose.yaml
mreid-tt: You're right. Removed the Jinja validation — the `required: true` in questions.yaml already prevents empty values.

ix-dev/community/seafile/ix_values.yaml
mreid-tt: The upstream `seafile-server.yml` uses `${SEAFILE_DB_IMAGE:-mariadb:10.11}` as its default, so we matched that initially. Updated to `mariadb:12.3.2` to match what other apps use. Verified working in local tests.

ix-dev/community/seafile/questions.yaml
mreid-tt: Changed to: the user provides a bare hostname, and the template appends the port from the network config. The user controls the hostname, so they can enter their external domain regardless of the internal port.

ix-dev/community/seafile/questions.yaml
mreid-tt: The upstream recommends 2GB minimum for Seafile CE, but the community catalog standard is 4096 as the default across virtually all apps. Changed to 4096 for consistency.

mreid-tt:
Thank you @stavros-k for the thorough review. I've worked through each of your comments — all CI checks are now passing. Let me know if there's anything else you'd like me to address.

EDIT: Regarding running as non-root — I attempted setting `NON_ROOT=true` on both the seafile and seadoc containers. The image creates a `seafile` user (uid 8000) internally when this is set. However, the seafile-mc image's init process creates directories and files as root during its bootstrap, before the permission check runs. This causes the permission validation to fail on every startup, as the newly created files are still root-owned.

The upstream's own documentation for non-root mode requires a manual `chmod -R a+rwx` on the host after the first deploy, then a container restart — which isn't practical in an automated deployment. The upstream defaults to root for both containers, so I've left it that way for now.

mreid-tt:
I've pushed a commit to run seafile and seadoc as non-root. The image supports `NON_ROOT=true` which creates a `seafile` user (uid 8000) internally and drops privileges. I also added an nginx dependency on seadoc (when enabled) to fix a startup ordering issue.

Two things needed custom workarounds:

1. **A startup script injected via Docker config** — The image's init process creates directories as root during boot, then checks that they're owned by uid 8000. The permissions container can't fix files that don't exist yet, so I inject a script at `/etc/my_init.d/99_fix_permissions.sh` that runs `chown` after the directories are created but before the ownership check. This uses the image's built-in extension mechanism rather than modifying its entrypoint.

2. **Nginx dependency on seadoc** — Nginx resolves upstream hostnames at startup. Without this dependency, `seadoc` may not be resolvable yet when nginx parses its config, causing nginx to fail.

Happy to revert to root mode if either of these workarounds is undesirable — the upstream defaults to root for both containers.

ix-dev/community/seafile/templates/docker-compose.yaml
```diff
@@ -0,0 +1,130 @@
...
+} %}
+{% set mariadb_container = tpl.deps.mariadb(values.consts.mariadb_container_name, "mariadb_image", mariadb_config, perm_container) %}
+{% do mariadb_container.container.add_network(seafile_net) %}
+
+{% set cache_config = {
+  "password": values.seafile.redis_password,
+  "volume": values.storage.redis_data,
+} %}
+{% set cache = tpl.deps.redis(values.consts.redis_container_name, "redis_image", cache_config, perm_container) %}
+{% do cache.container.set_command(["--port", "6379", "--requirepass", values.seafile.redis_password, "--stop-writes-on-bgsave-error", "no"]) %}
```
stavros-k: What is this?

ix-dev/community/seafile/templates/docker-compose.yaml
```diff
@@ -0,0 +1,130 @@
...
+{% do seafile.environment.add_env("INIT_SEAFILE_ADMIN_PASSWORD", values.seafile.admin_password) %}
+{% do seafile.environment.add_env("TIME_ZONE", values.TZ) %}
+{% do seafile.environment.add_env("CACHE_PROVIDER", "redis") %}
+{% do seafile.environment.add_env("REDIS_HOST", values.consts.redis_container_name) %}
+{% do seafile.environment.add_env("REDIS_PORT", 6379) %}
+{% do seafile.environment.add_env("REDIS_PASSWORD", values.seafile.redis_password) %}
+{% do seafile.environment.add_env("ENABLE_GO_FILESERVER", true) %}
+{% do seafile.environment.add_env("JWT_PRIVATE_KEY", values.seafile.jwt_private_key) %}
+{% do seafile.environment.add_env("NON_ROOT", true) %}
+{% do seafile.configs.add("fix-perms-seafile", "#!/bin/sh\nchown -R 8000:8000 /shared\n", "/etc/my_init.d/99_fix_permissions.sh", "0500") %}
```
stavros-k: Absolutely no.
The permissions container purposefully does not alter data permissions.
Bails out on non empty dirs. And a bunch of other checks.

This changes every time permissions of a directory that could contain data.

stavros-k: Still does not make sense, as it seems like it DOES chown
https://github.com/haiwen/seafile-docker/blob/f4906bc142b6ad422745bde407fdacb216045696/scripts/scripts_9.0/enterpoint.sh#L26-L37

stavros-k: And to address the elephant in the room.
Its is not "NON ROOT". PID 1 is still running as root.

ix-dev/community/seafile/templates/macros/nginx.conf
```diff
@@ -0,0 +1,63 @@
...
+}
+
+server {
+    listen {{ values.network.https_port.port_number }} ssl;
+    ssl_certificate {{ values.consts.ssl_cert_path }};
+    ssl_certificate_key {{ values.consts.ssl_key_path }};
+{% else %}
+server {
+    listen {{ values.network.web_port.port_number }};
+{% endif %}
```
stavros-k: I still dont get why we need both?

ix-dev/community/seafile/templates/macros/nginx.conf
```diff
@@ -0,0 +1,63 @@
+{% macro nginx_config(values, seadoc_enabled) -%}
+{% if values.network.certificate_id %}
+server {
+    listen {{ values.network.web_port.port_number }};
+    location /health {
+        return 200;
+    }
+    return 301 https://$host{{ ":" ~ values.network.https_port.port_number if values.network.https_port.port_number != 443 else "" }}$request_uri;
```
stavros-k: ...?

ix-dev/community/seafile/ix_values.yaml
stavros-k: I dont think we have any apps that names the container `db`

ix-dev/community/seafile/templates/docker-compose.yaml
mreid-tt: The `--stop-writes-on-bgsave-error no` flag is needed because Seafile's redis client triggers a BGSAVE on the first session write. Valkey's default behavior (inherited from Redis) is to block writes when RDB persistence encounters issues. This caused a 500 error on the login page on first access during testing. Setting `stop-writes-on-bgsave-error no` prevents this, which is safe because Valkey already logs persistence errors and this container is ephemeral/cache-only for Seafile.

ix-dev/community/seafile/templates/docker-compose.yaml
mreid-tt: **Agreed on all points.** The non-root experiment was a mistake — `my_init` as PID 1 requires root, and a custom init script that chowns on every start is unsafe. Reverted in full (commits `dbdd63c`, `ab9a7a1`). The seafile and seadoc containers now run as root again (matching the upstream image design), and no permissions manipulation is done beyond the standard permissions init container.

ix-dev/community/seafile/templates/macros/nginx.conf
mreid-tt: When a TrueNAS certificate is selected, nginx starts two server blocks:
1. **HTTP block** (port 30042) — serves the `/health` endpoint and redirects everything else to HTTPS via `301`
2. **HTTPS block** (port 30442) — serves all actual traffic with TLS

When no certificate is selected, only a single HTTP server block is used. This pattern matches how other community apps handle optional HTTPS (e.g., Homarr, Firefly III) and gives users flexibility to terminate TLS at a load balancer while keeping HTTP available for health checks.

ix-dev/community/seafile/ix_values.yaml
mreid-tt: Already fixed in commit `ab9a7a1` — renamed `db` → `mariadb` to match the convention used by all other community apps. The const is now `mariadb_container_name: mariadb`.

ix-dev/community/seafile/templates/macros/nginx.conf
mreid-tt: The `location /health { return 200; }` before the server-level `return 301` allows Docker health checks to still reach HTTP even when HTTPS is enabled. In nginx, the location block takes precedence over the server-level return for matching requests, so `/health` responds 200 while everything else gets redirected to HTTPS. This is important because Docker's health check system doesn't handle TLS redirects gracefully.

mreid-tt:
hey @stavros-k thanks for today's review. Here's what's changed in response:

- **Non-root experiment:** reverted in full (`dbdd63c`, `ab9a7a1`). You were right on all three points — `my_init` as PID 1 requires root, the fix-perms init script was unsafe, and the upstream image already handles `/shared` ownership internally.
- **Container name:** `db` → `mariadb` (`ab9a7a1`).

On the remaining items where no code change was made:
- **Redis `--stop-writes-on-bgsave-error`:** explanation in thread.
- **Nginx dual server blocks:** explanation in thread — follows the pattern used by Homarr, Firefly III, etc. for optional HTTPS.
- **HTTP redirect block with `/health`:** explanation in thread — needed so Docker health checks can reach HTTP even when TLS is enabled.

ix-dev/community/seafile/templates/macros/nginx.conf
stavros-k: If its just for healthcheck, we do not need to map the port. healthcheck runs internal to the container.

ix-dev/community/seafile/templates/docker-compose.yaml
stavros-k: Why does it encounter issues? 
What is the actual problem?

I checked upstream and does not seem to use this.

ix-dev/community/seafile/templates/macros/nginx.conf
stavros-k: also we dont use the `~` formatter, but the explicit format() func

mreid-tt: Good point. Health checks run within the container network and don't require a published port. Apologies for the oversight.

mreid-tt: Fixed in cb7e767 — replaced `~` with `|format()` to match the catalog convention.

ix-dev/community/seafile/templates/docker-compose.yaml
mreid-tt: Hmm, I'm not sure why I was seeing the 500 errors previously, but I haven't been able to reproduce them with this flag removed. Since Seafile uses Redis solely as a Django cache, `BGSAVE` persistence isn't relevant in this case, so the flag appears to be unnecessary.

mreid-tt:
Hey @stavros-k, I believe I've addressed all of the current review comments. Please let me know if there's anything else you'd like me to update or if there are any remaining concerns.

ix-dev/community/seafile/templates/macros/nginx.conf
stavros-k: We still have 2 ports for the user to configure. But we only need 1.

mreid-tt: Sorry but I don't understand. If we have options for HTTP and HTTPS, wouldn't we have to configure ports for both?

stavros-k: If we only expose one at a time, why would we need both? One web port can be configured either for https or http based on the certificate selection.

mreid-tt: Okay, I think I understand now. I noticed a similar pattern in the Nextcloud app and have pushed a commit following that approach. Let me know if I've implemented it correctly.

ix-dev/community/seafile/templates/docker-compose.yaml
```diff
@@ -0,0 +1,131 @@
...
+  {% do seadoc.environment.add_env("DB_HOST", values.consts.mariadb_container_name) %}
+  {% do seadoc.environment.add_env("DB_PORT", mariadb.get_port()) %}
+  {% do seadoc.environment.add_env("DB_USER", values.consts.db_user) %}
+  {% do seadoc.environment.add_env("DB_PASSWORD", values.seafile.db_password) %}
+  {% do seadoc.environment.add_env("DB_NAME", "seahub_db") %}
+  {% do seadoc.environment.add_env("JWT_PRIVATE_KEY", values.seafile.jwt_private_key) %}
+  {% do seadoc.environment.add_env("SEAHUB_SERVICE_URL", "http://%s"|format(values.consts.seafile_container_name)) %}
+
+  {# ??? #}
+  {% do seadoc.add_extra_host(values.seafile.server_hostname, "host-gateway") %}
```
stavros-k: why is this added?

ix-dev/community/seafile/templates/docker-compose.yaml
```diff
@@ -0,0 +1,131 @@
...
+{% do seafile.environment.add_env("SEAFILE_SERVER_HOSTNAME", seafile_hostname) %}
+{% do seafile.environment.add_env("SEAFILE_SERVER_PROTOCOL", seafile_proto) %}
+{% do seafile.environment.add_env("JWT_PRIVATE_KEY", values.seafile.jwt_private_key) %}
+{% do seafile.environment.add_env("ENABLE_SEADOC", values.seafile.enable_seadoc) %}
+{% if values.seafile.enable_seadoc %}
+  {% do seafile.environment.add_env("SEADOC_SERVER_URL", "%s://%s/sdoc-server"|format(seafile_proto, seafile_hostname)) %}
+{% endif %}
+{% do seafile.environment.add_user_envs(values.seafile.additional_envs) %}
+{# ??? #}
+{% do seafile.add_extra_host(values.seafile.server_hostname, "host-gateway") %}
```
stavros-k: why is this added?

ix-dev/community/seafile/templates/macros/nginx.conf
```diff
@@ -0,0 +1,52 @@
...
+    ssl_certificate {{ values.consts.ssl_cert_path }};
+    ssl_certificate_key {{ values.consts.ssl_key_path }};
+    error_page 497 =301 https://$host:{{ values.network.web_port.port_number }}$request_uri;
+    {% endif %}
+
+    location /health {
+        return 200;
+    }
+
+    location / {
```
stavros-k, line 14: all blocks have very slight differencies with what the upstream docs provide. 
care to explain?

ix-dev/community/seafile/questions.yaml
```diff
@@ -0,0 +1,723 @@
...
+            min_length: 8
+            default: ""
+            private: true
+            required: true
+        - variable: jwt_private_key
+          label: JWT Private Key
+          description: |
+            A random string with a length of no less than 32 characters.</br>
+            Required for Seafile authentication.</br>
+            You can generate one with `pwgen -s 40 1``
```
stavros-k: Command does not exist on neither my mac, nor on my truenas

ix-dev/community/seafile/questions.yaml
```diff
@@ -0,0 +1,723 @@
...
+            private: true
+        - variable: redis_password
+          label: Redis Password
+          description: The password for Seafile Redis.
+          schema:
+            type: string
+            default: ""
+            required: true
+            private: true
+        - variable: server_hostname
```
stavros-k: How does someone expose this behind a reverse proxy that listens on 443, while this app listen 30447 or whatever other free port is on the system?

ix-dev/community/seafile/templates/macros/nginx.conf
mreid-tt: The differences are mostly safe additions common across TrueNAS app nginx configs. `$http_host` is used instead of upstream's `$host` to preserve the port number (required for non-standard ports like 30447). `X-Forwarded-Proto` and `proxy_redirect off` are standard proxy hardening headers used by other catalog apps. The `/health` endpoint is a custom Docker healthcheck following the same pattern as wger and collabora. Two minor omissions from upstream — `X-NginX-Proxy true` and `proxy_redirect off` on `/socket.io` — can be added if needed.

ix-dev/community/seafile/templates/docker-compose.yaml
mreid-tt: It was added as a precaution but is unnecessary. `SEAFILE_SERVER_HOSTNAME` is only used for client-facing URLs — the container's internal nginx only proxies to `127.0.0.1` (gunicorn and fileserver), never to the external hostname. Can be removed.

ix-dev/community/seafile/templates/docker-compose.yaml
mreid-tt: Also unnecessary. SeaDoc connects to seafile via Docker DNS (`SEAHUB_SERVICE_URL=http://seafile`), and `SEADOC_SERVER_URL` is consumed by the browser, not the container itself. Can be removed.

ix-dev/community/seafile/questions.yaml
mreid-tt: You're right, my apologies. `pwgen` isn't available everywhere. The description should suggest a more portable alternative, e.g. `openssl rand -base64 32`.

ix-dev/community/seafile/questions.yaml
mreid-tt: That's a fair point, thank you. Currently the template always appends `web_port.port_number` to the hostname, which assumes the app's port is also the external port. For a reverse proxy scenario, the user needs to configure the external hostname and port separately. I'll look into adding a separate `public_hostname` or `external_url` field, or making the port optional when behind a proxy.

ix-dev/community/seafile/templates/docker-compose.yaml
```diff
@@ -0,0 +1,127 @@
...
+{% do seafile.environment.add_env("SEAFILE_MYSQL_DB_USER", values.consts.db_user) %}
+{% do seafile.environment.add_env("SEAFILE_MYSQL_DB_PASSWORD", values.seafile.db_password) %}
+{% do seafile.environment.add_env("INIT_SEAFILE_MYSQL_ROOT_PASSWORD", values.seafile.db_root_password) %}
+{% do seafile.environment.add_env("SEAFILE_MYSQL_DB_CCNET_DB_NAME", "ccnet_db") %}
+{% do seafile.environment.add_env("SEAFILE_MYSQL_DB_SEAFILE_DB_NAME", "seafile_db") %}
+{% do seafile.environment.add_env("SEAFILE_MYSQL_DB_SEAHUB_DB_NAME", "seahub_db") %}
+{% do seafile.environment.add_env("INIT_SEAFILE_ADMIN_EMAIL", values.seafile.admin_email) %}
+{% do seafile.environment.add_env("INIT_SEAFILE_ADMIN_PASSWORD", values.seafile.admin_password) %}
+{% do seafile.environment.add_env("SEAFILE_SERVER_HOSTNAME", seafile_hostname) %}
+{% do seafile.environment.add_env("SEAFILE_SERVER_PROTOCOL", seafile_proto) %}
```
stavros-k, line 102: Is protocol just an internal thing?
ie what should the protocol be when there is no certificate configured but I have an external reverse proxy?

mreid-tt: No, it's not just internal. Seafile uses `SEAFILE_SERVER_PROTOCOL` to construct all client-facing URLs (download links, avatars, file sharing, SeaDoc URL, etc.), so when an external proxy serves HTTPS, it needs to be `https` — regardless of whether a TrueNAS certificate is configured in the app.

The reason we currently tie it to `certificate_id` is that with the internal nginx handling SSL, the protocol and the cert go together. But for the external proxy case, they don't. The fix would be to decouple `SEAFILE_SERVER_PROTOCOL` from `certificate_id`, though we'd also need to handle the server-side callback issue where SeaDoc tries to reconnect to the server via HTTPS with a potentially untrusted certificate.

mreid-tt: I think the cleanest approach is to add an optional `external_url` field. When set, it overrides the auto-constructed hostname and protocol entirely — the user provides the full public URL (e.g. `https://seafile.example.com`), and that's what gets used for `SEAFILE_SERVER_HOSTNAME`, `SEAFILE_SERVER_PROTOCOL`, `SEADOC_SERVER_URL`, and the portal link. When left empty, it behaves as today: hostname + port derived from internal config.

This keeps the direct-access path simple (still works with no extra config) while giving proxy users a single place to set the correct public URL. It also avoids the CORS issue since the browser and the SeaDoc API URL will always be the same origin. Would this address your concern?

ix-dev/community/seafile/templates/macros/nginx.conf
```diff
@@ -0,0 +1,52 @@
...
+        proxy_set_header Host $http_host;
+        proxy_set_header X-Real-IP $remote_addr;
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+        proxy_set_header X-Forwarded-Proto $scheme;
+        proxy_redirect off;
+        proxy_http_version 1.1;
+        client_max_body_size 0;
+    }
+}
+    {% if values.seafile.enable_seadoc %}
```
mreid-tt: I believe this closing "}" for the server block should be further down

ix-dev/community/seafile/questions.yaml
```diff
@@ -0,0 +1,725 @@
...
+                schema:
+                  type: list
+                  items:
+                    - variable: container
+                      label: Container
+                      schema:
+                        type: string
+                        required: true
+                        enum:
+                          enum:
```
mreid-tt: This seems like a duplication

mreid-tt:
Apologies, I was mistaken in my earlier replies (re: `extra_host`) — I forgot why these were originally added. After further testing on the current commit, both `extra_host` entries are required for SeaDoc to work.

The seafile container's seahub API calls back to `http://seafile.example.com:8080/sdoc-server/...` and the seadoc server calls back via the same hostname for content loading. Without the `extra_host` entries, neither container can resolve the public hostname internally, producing `content_load_invalid` and `NameResolutionError` failures.

I'll push a commit to re-introduce them.

mreid-tt:
Following up on your earlier query about the nginx.conf variance from upstream — I tested the differences and confirmed they don't cause any functional issues (SeaDoc works identically either way). However, I'll still push a commit to align with upstream docs regardless, keeping only the intentional changes (`$http_host` for non-standard ports, container names for Docker networking, and the custom `/health` endpoint).

stavros-k (approved):

PR #5036, opened 2026-05-26
Update Bitcoin Knots versioning in renovate-config.js

Retropex:
The bot didn't catch the latest update.

### DIFF
--- a/.github/renovate-config.js
+++ b/.github/renovate-config.js
@@ -352,6 +352,11 @@ module.exports = {
       "^(?<major>\\d+)\\.(?<minor>\\d+)\\.(?<patch>\\d+)(\\.(?<build>\\d+))?$",
       ["jvmilazz0/kavita"],
     ),
+    customVersioning(
+      // 1.0.0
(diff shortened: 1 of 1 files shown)


ix-dev/community/bitcoind-knots/ix_values.yaml
```diff
@@ -1,7 +1,7 @@
 images:
   image:
     repository: ghcr.io/retropex/bitcoin-truenas
-    tag: v29.3
+    tag: v29.3.3
```
stavros-k: Is the versioning going to be vX.Y AND vX.Y.Z? If that the case, a permanent fix should be applied on renovate-config.js and let bot pick the update (to verify that it work)

.github/renovate-config.js
```diff
@@ -352,6 +352,11 @@ module.exports = {
       "^(?<major>\\d+)\\.(?<minor>\\d+)\\.(?<patch>\\d+)(\\.(?<build>\\d+))?$",
       ["jvmilazz0/kavita"],
     ),
+    customVersioning(
```
stavros-k, line 355: This will ONLY pickup `x.y.z` (no `v` prefix) 
And will ignore `x.y`.
Is that expected?

PR #5074, opened 2026-06-02
zwave-js-ui: fix healthcheck

### LINKED ISSUE #5067: [Bug]: zwave-js-ui health check leaves zombie ssl_client processes
# App

Name: zwave-js-ui
Version: 1.1.17

# Issue

The compose health check leaves zombie ssl_client processes that never get cleaned up.  This seems to be an upstream issue with busybox's `wget` implementation, which doesn't wait on its child `ssl_client`.

Image

# Possible Fixes

~~1. Modify the `wget` health check type to run `sh -c wget` instead of running `wget` directly.~~

~~Pros: simple~~
~~Cons: wide-reaching; could impact other apps~~

EDIT: Tested, this option doesn't actually fix the issue

2. Use a `pgrep` health check

Pros: simple
Cons: doesn't actually test that the `node` server process is serving requests, so it could miss a silent hang that doesn't actually cause the process to die.

3. Add a new `https` health check that works similarly to the `http` one, but uses `ssl_client`

Pros: Replicates the current health check
Cons: More complex

stavros-k:
Closes #5067


stavros-k:
Failure is due to not having a zwave device.

Qubad786 (approved):

PR #5164, opened 2026-06-13
frigate: add Privileged Mode toggle

scyto:
## Description

Adds an optional **Privileged Mode** toggle to the existing Frigate community app. When enabled the container runs privileged; when off (the default) nothing changes, so existing installs are unaffected.

Some Frigate deployments need privileged access for hardware that the existing **Devices** / **Mount USB Bus** options can't cover - for example PCIE MemryX MX3 device. 

The base library already emits a "Privileged mode is enabled" security note for privileged containers, so the toggle surfaces that automatically.

> This modifies an existing app, so the "App Addition" template doesn't apply.

## AI

## Changes

- `questions.yaml`: new `privileged` boolean (default `false`) with a warning
- `templates/docker-compose.yaml`: `set_privileged(true)` when enabled
- `templates/test_values/{basic,https}-values.yaml`: cover off / on
- `app.yaml`: version `1.3.8` → `1.3.9`

## Testing

Verified on real hardware (TrueNAS 25.10.4): the toggle appears in Frigate's
config form, enabling it sets the container privileged
(`docker inspect … Privileged=true`), and the app's Security/Notes card shows
"Privileged mode is enabled". Defaults to off.

## Checklist

- [x] Only files under `/ix-dev/` are modified (no auto-generated files)
- [x] `version` bumped in `app.yaml`
- [x] `questions.yaml` has a clear label and description

### DIFF
--- a/ix-dev/community/frigate/app.yaml
+++ b/ix-dev/community/frigate/app.yaml
@@ -49,4 +49,4 @@ sources:
 - https://github.com/blakeblackshear/frigate
 title: Frigate
 train: community
-version: 1.3.8
+version: 1.3.9
--- a/ix-dev/community/frigate/questions.yaml
+++ b/ix-dev/community/frigate/questions.yaml
@@ -56,6 +56,16 @@ questions:
           schema:
             type: boolean
             default: false
+        - variable: privileged
+          label: Privileged Mode
+          description: |
+            Run the Frigate container in privileged mode.</br>
+            This grants the container full access to the host, equivalent to root
+            on the host system. Only enable this if you need hardware/device access
+            that the "Devices" and "Mount USB Bus" options above cannot provide.
+          schema:
+            type: boolean
+            default: false
         - variable: devices
           label: Devices
           description: |
--- a/ix-dev/community/frigate/templates/docker-compose.yaml
+++ b/ix-dev/community/frigate/templates/docker-compose.yaml
@@ -4,6 +4,7 @@
 
 {% do c1.set_user(0, 0) %}
 {% do c1.add_caps(["CHOWN", "FOWNER", "DAC_OVERRIDE", "SETGID", "SETUID", "PERFMON", "KILL"]) %}
+{% if values.frigate.privileged %}{% do c1.set_privileged(true) %}{% endif %}
 {% do c1.set_shm_size_mb(tpl.funcs.or_default(values.frigate.shm_size_mb, 30)) %}
 {% if values.frigate.mount_usb_bus %}
   {% do c1.devices.add_usb_bus() %}
--- a/ix-dev/community/frigate/templates/test_values/basic-values.yaml
+++ b/ix-dev/community/frigate/templates/test_values/basic-values.yaml
@@ -7,6 +7,7 @@ frigate:
   image_selector: image
   shm_size_mb: 64
   mount_usb_bus: false
+  privileged: false
   additional_envs: []
 
 network:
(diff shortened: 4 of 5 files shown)


scyto:
To add: This is to allow me to use MemryX on TrueNAS without needing to use custom application mode.
See here. https://github.com/truenas-community-sysexts/memryx-mx3-support

This change was tested on local live TrueNAS with the script below as I couldn't fathom easier way to do it on real hardware.

Yes the PR was done with the assistance of AI, but given the simplicity of the change i thought i would propose as a PR.  If you would prefer to implement yourself that is also fine and we can convert to a feature request.

```
# ---- RUN THIS ON THE TRUENAS BOX (as root) ----
python3 - <<'PY'
import os, shutil

roots = [
    "/mnt/.ix-apps/app_configs/frigate/versions/1.3.8",                 # installed app (the important one)
    "/mnt/.ix-apps/truenas_catalog/trains/community/frigate/1.3.8",     # Discover/fresh-install
    "/mnt/.ix-apps/truenas_catalog/ix-dev/community/frigate",           # catalog source
]

Q_BLOCK = '''        - variable: privileged
          label: Privileged Mode
          description: |
            Run the Frigate container in privileged mode.</br>
            This grants the container full access to the host, equivalent to root
            on the host system. Only enable this if you need hardware/device access
            that the "Devices" and "Mount USB Bus" options above cannot provide.
          schema:
            type: boolean
            default: false
'''
Q_MARKER = "        - variable: devices\n"
COMPOSE_LINE = "{% if values.frigate.privileged %}{% do c1.set_privileged(true) %}{% endif %}\n"

def backup(p):
    if not os.path.exists(p + ".bak"):
        shutil.copy2(p, p + ".bak")

def patch_questions(p):
    s = open(p).read()
    if "variable: privileged" in s:
        print("  = questions.yaml already patched"); return
    if Q_MARKER not in s:
        print("  ! questions.yaml marker not found, SKIPPED"); return
    backup(p); open(p, "w").write(s.replace(Q_MARKER, Q_BLOCK + Q_MARKER, 1))
    print("  + patched questions.yaml")

def patch_compose(p):
    s = open(p).read()
    if "set_privileged" in s:
        print("  = compose already patched"); return
    out, done = [], False
    for ln in s.splitlines(keepends=True):
        out.append(ln)
        if not done and "add_caps(" in ln and "CHOWN" in ln:
            out.append(COMPOSE_LINE); done = True
    if not done:
        print("  ! compose add_caps line not found, SKIPPED"); return
    backup(p); open(p, "w").write("".join(out))
    print("  + patched compose")

for r in roots:
    print(r)
    q, c = os.path.join(r, "questions.yaml"), os.path.join(r, "templates/docker-compose.yaml")
    patch_questions(q) if os.path.exists(q) else print("  (no questions.yaml)")
    patch_compose(c)   if os.path.exists(c) else print("  (no compose)")
print("Done.")
PY
```

stavros-k:
Adding this flag is to _enable_ a functionality that requires modifying the base TrueNAS system in an unsupported way.
If you really want to run this with that functionality is up to you, but this flag won't be shipped in the catalog app.

Thanks

scyto:
> Adding this flag is to _enable_ a functionality that requires modifying the base TrueNAS system in an unsupported way. If you really want to run this with that functionality is up to you, but this flag won't be shipped in the catalog app.
> 
> Thanks

well truenas has said sysexts are the official way to do this, they do it with the nvidia driver and honeybadger has said this multiple times in the forum, so color me confused, but ok, the script is easy enough to make the change / users can use a custom app

stavros-k:
What I mean by supported is what can be done via the WebUI.

PR #5173, opened 2026-06-15
calagopus: fix wings temp data path

### LINKED ISSUE #5159: Calagopus broken temp mount
As explained in https://github.com/truenas/apps/pull/5140#issuecomment-4660412953, Calagopus wings relies on the existence of `/tmp/wings` in order to run its server installation scripts.

Unfortunately there was a slight miscommunication from my side. `/tmp/wings` should exist on the host, and be bound at the same location to the container. Why? Wings communicates with the host docker socket, and tries to bind `/tmp/wings/{uuid}` to a container. This will create errors, because `/tmp/wings` does not exist on the host. 

So this
Image
should look more like
Image

stavros-k:
Closes #5159


Jelcoo:
Can confirm this works as expected

Qubad786 (approved):

PR #5182, opened 2026-06-16
Crafty 4.10.5 is broken, update to 4.10.6

dberlin:
Crafty 4.10.5 was broken as released due to a missing python package (see https://gitlab.com/crafty-controller/crafty-4/-/merge_requests/1042)

Update to 4.10.6


stavros-k:
See #5183

dberlin:
Thank you!

PR #5245, opened 2026-06-25
hoppscotch: fix ws url

### LINKED ISSUE #5243: [Bug Report]: Hoppscotch app produces invalid VITE_BACKEND_WS_URL environment variable protocol
### App Name
Hoppscotch

v1.2.24

### App Issue
The app sets the environment variable `VITE_BACKEND_WS_URL` to an invalid value, like `ws:http://sub.domain.tld:12345`.  Additionally, the websocket URL is never set to use the secure `wss` protocol, even when the backend supports it.  This issue causes the front-end application to throw errors like  `Mixed Content: The page at 'https://sub.domain.tld/' was loaded over HTTPS, but attempted to connect to the insecure WebSocket endpoint 'ws://https//sub.domain.tld/graphql'. This request has been blocked; this endpoint must be available over WSS.` and `Uncaught SecurityError: Failed to construct 'WebSocket': An insecure WebSocket connection may not be initiated from a page loaded over HTTPS`.

Per the Hoppscotch [documentation](https://docs.hoppscotch.io/documentation/self-host/community-edition/install-and-build) an appropriate value would be something like `wss://localhost:3170/graphql` (note that the `http(s):` portion is properly excluded).

### Proposed Solution
The issue is caused by this line from the app's docker-compose.yaml file:

`{% do c.environment.add_env("VITE_BACKEND_WS_URL", "ws:%s/graphql" | format(values.hoppscotch.urls.backend)) %}`

The app's questions.yaml requires the `values.hoppscotch.urls.backend` variable to match a uri schema, like `http(s)://(sub.)domain.tld(:port)`.  This causes the value of `VITE_BACKEND_WS_URL` to be invalid (with two protocols) .  Additionally, this is what forces the websocket URL to use the insecure protocol `ws` even if the backend URL suppports secure connections.

I propose that instead of prepending `values.hoppscotch.urls.backend` wtih `ws:`, the protocol in values.hoppscotch.urls.backend (`http` or `https`) simply be replaced with the appropriate `ws` or `wss` protocol.  This would turn `http://sub.domain.tld:12345` into `ws://sub.domain.tld:12345` and `https://sub.domain.tld:12345` into `wss://sub.domain.tld:12345`.

### Alternatives Considered
Alternatively, a `values.hoppscotch.urls.websocket` variable could be added to the questions.yaml file and that could be usedt instead.  I'd opt not to go this route as, to my understanding. the backend url and the websocket url will always be the same, with the exception of the `/graphql` path and the different protocols.

stavros-k:
Closes #5243


IgneusManus (approved):
That was phenomenally quick.  Looks good to me.  Thanks!

Qubad786 (approved):

PR #5254, opened 2026-06-26
Add restic-manager to the community train

blaubaer:
# App Addition

- [ ] I have opened an [issue](https://github.com/truenas/apps/issues) to discuss this app addition before submitting this pull request.

# AI

## Description

Adds [TrueNAS Restic Manager](https://codeberg.org/echocat/truenas-restic-manager) to the community train.

TrueNAS Restic Manager is a TrueNAS-integrated backup manager that uses restic to back up selected TrueNAS datasets from snapshots. It can pair with TrueNAS, discover datasets, create temporary ZFS snapshots, run scheduled backups, and keep persistent job and run history.

This app was developed specifically for TrueNAS and is not a generic upstream app packaging request, so there was no separate prior app-request issue for it.

The app uses the normal TrueNAS Apps storage configuration shape. User-selected dataset mounts are represented through a dedicated `storage.dataset_mounts` configuration field and are rendered as read-only host-path mounts with `rslave` propagation.

## App Information

- **Upstream**: https://codeberg.org/echocat/truenas-restic-manager
- **Documentation**: https://codeberg.org/echocat/truenas-restic-manager
- **App Version**: 0.2.0

## Testing

Tested locally with:

- [x] basic-values.yaml

Additional verification:

- local catalog render validation via `devbox run app-render community restic-manager basic-values.yaml`
- verified rendered compose contains the expected fixed environment values:
  - `TRM_APP_ID=restic-manager`
  - `TRM_APP_MOUNTS_PROPERTY=storage.dataset_mounts`
  - `TRM_STORE_MIGRATE_ON_START=true`
  - `TRM_WORKDIR=/work`
  - `TRM_STORE_PATH=/data/store.db`
  - `RESTIC_CACHE_DIR=/cache`
- verified `host.docker.internal:host-gateway` is configured for TrueNAS API access from the container
- verified dataset mount rendering forces read-only `rslave` propagation
- manual runtime validation on a TrueNAS host using the exported docker-compose.yaml

`devbox run app-test community restic-manager basic-values.yaml` was also run. The container starts, but the test environment does not provide a TrueNAS API endpoint at `https://host.docker.internal`, so the service exits during startup when it cannot connect to TrueNAS. This is expected outside a TrueNAS host environment.

## Icons and Screenshots

- Icon: https://codeberg.org/repo-avatars/55d216d8ece50dc8f9d5f4cba948251273ee557f8b8a6d9375dcc8589f94d35c
- Screenshot 1 (optional): not provided yet

## Special Notes

- This app is intentionally designed to run on the TrueNAS host.
- Users sign in with their normal TrueNAS credentials. Currently, the TrueNAS user must have the `Full Admin` role for pairing and management operations.
- The app does not maintain its own authentication database.
- TrueNAS-managed authentication such as TOTP continues to apply.
- The app connects to the TrueNAS API through the configured `TRM_TRUENAS_API_URL`.
- `host.docker.internal:host-gateway` is configured so the container can reach the TrueNAS host.
- TrueNAS commonly uses a self-signed certificate, so `TRM_TRUENAS_SKIP_TLS_CHECK` defaults to `true`.
- HTTPS is served on the app port using the TrueNAS UI certificate by default.
- Plain HTTP behavior on the same app port is controlled by `TRM_TLS_INSECURE`, defaulting to `on`.
- Dataset mounts selected for backup are rendered read-only and with `rslave` propagation.
- Persistent storage is required for `/data`.
- Persistent storage is recommended for `/cache`.
- `/work` is temporary `tmpfs` storage.
- If used behind a reverse proxy, `TRM_UPSTREAM_TRUSTED_ADDRESS` should be configured so forwarded headers and optional `X-Forwarded-Prefix` values are honored.
- The app does not require `privileged`, `SYS_ADMIN`, or Docker socket access.

## Checklist

- [x] App runs successfully locally on a TrueNAS host
- [x] Only modified files under /ix-dev/
- [x] README.md included
- [x] Render validation passes with basic-values.yaml
- [ ] All automated CI checks pass (waiting)

### DIFF
--- a/ix-dev/community/restic-manager/ix_values.yaml
+++ b/ix-dev/community/restic-manager/ix_values.yaml
@@ -1,12 +1,10 @@
 images:
   image:
     repository: codeberg.org/echocat/truenas-restic-manager
-    tag: v0.2.0
+    tag: v0.2.1
 
 consts:
   restic_manager_container_name: restic-manager
   app_id: restic-manager
-  app_mounts_property: storage.dataset_mounts
   api_user: restic-manager
   api_key_name: restic-manager
-  internal_web_port: 9080
--- a/ix-dev/community/restic-manager/ix_values.yaml
+++ b/ix-dev/community/restic-manager/ix_values.yaml
@@ -1,0 +1,12 @@
+images:
+  image:
+    repository: codeberg.org/echocat/truenas-restic-manager
+    tag: v0.2.0
+
+consts:
+  restic_manager_container_name: restic-manager
+  app_id: restic-manager
+  app_mounts_property: storage.dataset_mounts
+  api_user: restic-manager
+  api_key_name: restic-manager
+  internal_web_port: 9080
--- a/ix-dev/community/restic-manager/questions.yaml
+++ b/ix-dev/community/restic-manager/questions.yaml
@@ -39,82 +39,21 @@ questions:
             type: string
             default: https://host.docker.internal
             required: true
-        - variable: truenas_skip_tls_check
-          label: Skip TrueNAS TLS Verification
-          description: |
-            Keep this enabled when TrueNAS uses its default self-signed certificate.</br>
-            Disable it only when your TrueNAS API uses a certificate trusted by the container.
-          schema:
-            type: boolean
-            default: true
-        - variable: tls_insecure
-          label: Plain HTTP Handling
-          description: |
-            Controls plain HTTP behavior on the same app port.</br>
-            on: HTTP and HTTPS are both accepted.</br>
-            redirect: HTTP redirects to HTTPS.</br>
-            off: HTTP is rejected.
-          schema:
-            type: string
-            default: "on"
-            required: true
-            enum:
-              - value: "on"
-                description: Accept HTTP and HTTPS on the same port
-              - value: "redirect"
-                description: Redirect HTTP to HTTPS
-              - value: "off"
-                description: Reject HTTP
         - variable: upstream_trusted_address
           label: Trusted Upstream Addresses
           description: |
-            Optional comma-separated IP/CIDR ranges of trusted reverse proxies.</br>
+            Optional IP/CIDR ranges of trusted reverse proxies.</br>
             When set, Restic Manager trusts `X-Forwarded-Prefix` and `X-Forwarded-For` from these addresses.</br>
-            Example: `127.0.0.0/8,::1/128,172.16.0.0/16`
+            Example: `172.16.0.0/16`
           schema:
-            type: string
-            default: ""
-        - variable: log
-          label: Logging
-          schema:
-            type: dict
-            attrs:
-              - variable: level
-                label: Log Level
+            type: list
+            default: []
+            items:
+              - variable: address
+                label: Address
                 schema:
                   type: string
-                  default: info
                   required: true
-                  enum:
-                    - value: trace
-                      description: trace
-                    - value: debug
-                      description: debug
-                    - value: info
-                      description: info
-                    - value: warn
-                      description: warn
-                    - value: error
-                      description: error
-                    - value: fatal
-                      description: fatal
-              - variable: color
-                label: Log Color
-                description: Colored logs can cause display issues in the TrueNAS app log UI.
-                schema:
-                  type: boolean
-                  default: false
-              - variable: format
-                label: Log Format
-                schema:
-                  type: string
-                  default: text
-                  required: true
-                  enum:
-                    - value: text
-                      description: Text
-                    - value: json
-                      description: JSON
         - variable: additional_envs
           label: Additional Environment Variables
           schema:
@@ -168,7 +107,7 @@ questions:
                 label: Port Number
                 schema:
                   type: int
-                  default: 9080
+                  default: 30446
                   min: 1
                   max: 65535
                   required: true
@@ -366,7 +305,7 @@ questions:
           label: Dataset Mounts
           description: |
             Dataset host-path mounts used by Restic Manager.</br>
-            These entries are normally managed from inside Restic Manager during pairing/setup and should not be edited manually in the TrueNAS app configuration.</br>
+            Restic Manager uses these entries to verify that selected backup source datasets are mounted for snapshot-based backups.</br>
             Entries are rendered read-only and with `rslave` bind propagation.
           schema:
             type: list
--- a/ix-dev/community/restic-manager/questions.yaml
+++ b/ix-dev/community/restic-manager/questions.yaml
@@ -1,0 +1,477 @@
+groups:
+  - name: Restic Manager Configuration
+    description: Configure TrueNAS Restic Manager
+  - name: Network Configuration
+    description: Configure Network for TrueNAS Restic Manager
+  - name: Storage Configuration
+    description: Configure Storage for TrueNAS Restic Manager
+  - name: Labels Configuration
+    description: Configure Labels for TrueNAS Restic Manager
+  - name: Resources Configuration
+    description: Configure Resources for TrueNAS Restic Manager
+
+questions:
+  - variable: TZ
+    group: Restic Manager Configuration
+    label: Timezone
+    schema:
+      type: string
+      default: Etc/UTC
+      required: true
+      $ref:
+        - definitions/timezone
+  - variable: restic_manager
+    label: ""
+    group: Restic Manager Configuration
+    schema:
+      type: dict
+      attrs:
+        - variable: truenas_api_url
+          label: TrueNAS API URL
+          description: |
+            URL used by Restic Manager to reach the TrueNAS API.</br>
+            Sign in to Restic Manager with your own TrueNAS user credentials. Currently, the TrueNAS user must have the `Full Admin` role for pairing and management operations.</br>
+            The default `https://host.docker.internal` works when TrueNAS listens on the default HTTPS port.</br>
+            If TrueNAS listens on another port, include it, for example `https://host.docker.internal:1234`.</br>
+            If TrueNAS only listens on selected interfaces, use the reachable host or IP instead, for example `https://192.168.1.2:1234`.</br>
+            This app always binds `host.docker.internal:host-gateway` for the container.
+          schema:
+            type: string
+            default: https://host.docker.internal
+            required: true
+        - variable: truenas_skip_tls_check
+          label: Skip TrueNAS TLS Verification
+          description: |
+            Keep this enabled when TrueNAS uses its default self-signed certificate.</br>
+            Disable it only when your TrueNAS API uses a certificate trusted by the container.
+          schema:
+            type: boolean
+            default: true
+        - variable: tls_insecure
+          label: Plain HTTP Handling
+          description: |
+            Controls plain HTTP behavior on the same app port.</br>
+            on: HTTP and HTTPS are both accepted.</br>
+            redirect: HTTP redirects to HTTPS.</br>
+            off: HTTP is rejected.
+          schema:
+            type: string
+            default: "on"
+            required: true
+            enum:
+              - value: "on"
+                description: Accept HTTP and HTTPS on the same port
+              - value: "redirect"
+                description: Redirect HTTP to HTTPS
+              - value: "off"
+                description: Reject HTTP
+        - variable: upstream_trusted_address
+          label: Trusted Upstream Addresses
+          description: |
+            Optional comma-separated IP/CIDR ranges of trusted reverse proxies.</br>
+            When set, Restic Manager trusts `X-Forwarded-Prefix` and `X-Forwarded-For` from these addresses.</br>
+            Example: `127.0.0.0/8,::1/128,172.16.0.0/16`
+          schema:
+            type: string
+            default: ""
+        - variable: log
+          label: Logging
+          schema:
+            type: dict
+            attrs:
+              - variable: level
+                label: Log Level
+                schema:
+                  type: string
+                  default: info
+                  required: true
+                  enum:
+                    - value: trace
+                      description: trace
+                    - value: debug
+                      description: debug
+                    - value: info
+                      description: info
+                    - value: warn
+                      description: warn
+                    - value: error
+                      description: error
+                    - value: fatal
+                      description: fatal
+              - variable: color
+                label: Log Color
+                description: Colored logs can cause display issues in the TrueNAS app log UI.
+                schema:
+                  type: boolean
+                  default: false
+              - variable: format
+                label: Log Format
+                schema:
+                  type: string
+                  default: text
+                  required: true
+                  enum:
+                    - value: text
+                      description: Text
+                    - value: json
+                      description: JSON
+        - variable: additional_envs
+          label: Additional Environment Variables
+          schema:
+            type: list
+            default: []
+            items:
+              - variable: env
+                label: Environment Variable
+                schema:
+                  type: dict
+                  attrs:
+                    - variable: name
+                      label: Name
+                      schema:
+                        type: string
+                        required: true
+                    - variable: value
+                      label: Value
+                      schema:
+                        type: string
+
+  - variable: network
+    label: ""
+    group: Network Configuration
+    schema:
+      type: dict
+      attrs:
+        - variable: web_port
+          label: WebUI Port
+          schema:
+            type: dict
+            attrs:
+              - variable: bind_mode
+                label: Port Bind Mode
+                description: |
+                  The port bind mode.</br>
+                  - Publish: The port will be published on the host for external access.</br>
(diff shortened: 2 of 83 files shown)


ix-dev/community/restic-manager/templates/library/base_v2_3_8/container.py
```diff
@@ -332,6 +332,14 @@ def add_docker_socket(self, read_only: bool = True, mount_path: str = "/var/run/
...
         self._storage._add_udev(read_only, mount_path)
 
+    def add_truenas_mnt(self, read_only: bool = True, mount_path: str = "/mnt"):
+        self._storage._add_truenas_mnt(read_only, mount_path)
+
+    def add_truenas_middleware_socket(
+        self, read_only: bool = False, mount_path: str = "/var/run/middleware/middlewared.sock"
+    ):
+        self._storage._add_truenas_middleware_socket(read_only, mount_path)
+
```
stavros-k: There are no plans allowing catalog apps mount those at the moment.

stavros-k: Also you are editing an existing library version, versions are immutable.

stavros-k:
> * It uses the TrueNAS API/middleware directly for authentication and system integration, including user login, dataset discovery, snapshot creation, and certificate access.

You can ask the user to provide an api key. Direct access to the socket is not going to be exposed in catalog apps at the moment

blaubaer:
@stavros-k, thanks for pointing that out. You were right about the library version; mutating `library/2.3.7` was a mistake.

I fixed that part now. The PR no longer modifies the existing `library/2.3.7` in place. Instead, I bumped the catalog library to `2.3.8` in a separate commit and regenerated the app-local library copy for `restic-manager`.

The relevant commits are split intentionally:

- [`lib: bump to 2.3.8`](https://github.com/truenas/apps/tree/cfc9a795665843487312301bf5cef594925b758f)
- [`restic-manager: use lib 2.3.8`](https://github.com/truenas/apps/tree/21912dfc6982eeb27d22a4b181a6fa1bec7ed4f0)

Local validation passed:

- `python -m pytest library/2.3.8/tests`
- `./.github/scripts/ci.py --app restic-manager --train community --test-file basic-values.yaml --render-only=true`

On the mount/socket policy point: I understand this would establish a high-trust precedent. That is intentional for this app class.

Restic Manager is not a generic web app. It is a TrueNAS-host-integrated backup manager: it discovers datasets, creates temporary ZFS snapshots, backs up from snapshot paths under `/mnt`, and uses TrueNAS middleware for system integration.

Using the middleware socket is not just convenience. For this app it avoids requiring users to configure local TrueNAS hostname/port/TLS settings and to create/store a long-lived high-privilege API key. API-key mode would not materially reduce the privileges required for snapshot-based backups; it would mostly replace local host trust with persistent secret management and more user-facing configuration failure modes.

The same applies to `/mnt`: restricting this to manually mapped dataset paths would make the app much less TrueNAS-native and duplicate TrueNAS dataset selection in app setup. The mount is read-only and implemented through explicit library helpers instead of weakening generic host path validation.

So I think the real policy question is whether the catalog can allow explicitly declared high-trust TrueNAS-native apps. If the answer is no, I understand that this app cannot fit the catalog in its intended form. If the answer is yes, I am happy to adjust wording, warnings, metadata, or guardrails around this app class.

stavros-k:
At this moment there are no plans to allow access to the socket on pre-packaged catalog apps.
And /mnt is explicitly disallowed (you just bypassed the validation with the lib helper).
So, no as a prepackaged app is not going to happen now.
If a user wants, they can install it as a custom app themselves.

blaubaer:
Hi @stavros-k,

thanks again for the previous feedback. I reworked the app to fit the constraints you pointed out.

Main changes:

- no middleware socket access
- no broad `/mnt` mount
- no shared library changes
- TrueNAS access now goes through the configured API URL, defaulting to `https://host.docker.internal`
- dataset access is represented through `storage.dataset_mounts` and rendered as read-only host-path mounts with forced `rslave` propagation
- README/questions now mention that users sign in with their own TrueNAS credentials and currently need `Full Admin`

Validation:

- `devbox run app-render community restic-manager basic-values.yaml` passes
- manually tested on a TrueNAS host using the exported compose output
- local `app-test` was attempted, but it cannot fully pass outside TrueNAS because no TrueNAS API is available at `https://host.docker.internal`

Please let me know if there is another catalog constraint I missed.

ix-dev/community/restic-manager/questions.yaml
```diff
@@ -0,0 +1,477 @@
...
+                      schema:
+                        type: int
+                        default: 256
+                        required: true
+        - variable: dataset_mounts
+          label: Dataset Mounts
+          description: |
+            Dataset host-path mounts used by Restic Manager.</br>
+            These entries are normally managed from inside Restic Manager during pairing/setup and should not be edited manually in the TrueNAS app configuration.</br>
+            Entries are rendered read-only and with `rslave` bind propagation.
```
stavros-k, line 309: What does that mean?

ix-dev/community/restic-manager/questions.yaml
```diff
@@ -0,0 +1,477 @@
...
+            default: "on"
+            required: true
+            enum:
+              - value: "on"
+                description: Accept HTTP and HTTPS on the same port
+              - value: "redirect"
+                description: Redirect HTTP to HTTPS
+              - value: "off"
+                description: Reject HTTP
+        - variable: upstream_trusted_address
```
stavros-k, line 42: this should be a list

ix-dev/community/restic-manager/questions.yaml
```diff
@@ -0,0 +1,477 @@
...
+          label: Trusted Upstream Addresses
+          description: |
+            Optional comma-separated IP/CIDR ranges of trusted reverse proxies.</br>
+            When set, Restic Manager trusts `X-Forwarded-Prefix` and `X-Forwarded-For` from these addresses.</br>
+            Example: `127.0.0.0/8,::1/128,172.16.0.0/16`
+          schema:
+            type: string
+            default: ""
+        - variable: log
```
stavros-k: log config settings shouldnt be exposed, if user wants to tune it, can use additonal envs

ix-dev/community/restic-manager/questions.yaml
```diff
@@ -0,0 +1,477 @@
...
+            required: true
+        - variable: truenas_skip_tls_check
+          label: Skip TrueNAS TLS Verification
+          description: |
+            Keep this enabled when TrueNAS uses its default self-signed certificate.</br>
+            Disable it only when your TrueNAS API uses a certificate trusted by the container.
+          schema:
+            type: boolean
+            default: true
+        - variable: tls_insecure
```
stavros-k: can probably be removed and user can set it via additional env

ix-dev/community/restic-manager/questions.yaml
```diff
@@ -0,0 +1,477 @@
...
+                      description: Publish port on the host for external access
+                    - value: "exposed"
+                      description: Expose port for inter-container communication
+                    - value: ""
+                      description: None
+              - variable: port_number
+                label: Port Number
+                schema:
+                  type: int
+                  default: 9080
```
stavros-k: 30k+

ix-dev/community/restic-manager/templates/docker-compose.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+{% set c1 = tpl.add_container(values.consts.restic_manager_container_name, "image") %}
+
+{% do c1.set_user(0, 0) %}
+{% do c1.healthcheck.set_custom_test(["CMD", "/usr/local/bin/truenas-restic-manager", "healthcheck"]) %}
+{% do c1.add_extra_host("host.docker.internal", "host-gateway") %}
+
+{% do c1.environment.add_env("TRM_TRUENAS_API_URL", values.restic_manager.truenas_api_url) %}
+{% do c1.environment.add_env("TRM_TRUENAS_SKIP_TLS_CHECK", values.restic_manager.truenas_skip_tls_check) %}
+{% do c1.environment.add_env("TRM_LISTEN", ":%d" | format(values.consts.internal_web_port)) %}
```
stavros-k: should use the configured port.

ix-dev/community/restic-manager/templates/docker-compose.yaml
```diff
@@ -0,0 +1,52 @@
...
+{% for store in values.storage.dataset_mounts %}
+  {% set dataset_path = store.host_path_config.path or store.mount_path %}
+  {% set dataset_store = {
+    "type": "host_path",
+    "read_only": true,
+    "mount_path": dataset_path,
+    "host_path_config": {
+      "acl_enable": false,
+      "path": dataset_path,
+      "propagation": "rslave",
```
stavros-k, line 39: I don't think I want to use rslave.
I can't remember currently, but I had to look into it a while ago, and wasn't working right / causing issues.

ix-dev/community/restic-manager/ix_values.yaml
```diff
@@ -0,0 +1,12 @@
+images:
+  image:
+    repository: codeberg.org/echocat/truenas-restic-manager
+    tag: v0.2.0
+
+consts:
+  restic_manager_container_name: restic-manager
+  app_id: restic-manager
+  app_mounts_property: storage.dataset_mounts
```
stavros-k: what is this?!

ix-dev/community/restic-manager/questions.yaml
blaubaer: Reworded this section to make it clearer.

The field is used by Restic Manager to verify that the selected backup source datasets are mounted for snapshot-based backups. Without those mounts, backup jobs would fail later with less useful file access errors.

ix-dev/community/restic-manager/questions.yaml
blaubaer: Changed this from a comma-separated string to a list in `questions.yaml`.

The template now joins the list back into the comma-separated format expected by `TRM_UPSTREAM_TRUSTED_ADDRESS`.

ix-dev/community/restic-manager/questions.yaml
blaubaer: Removed the exposed log settings.

Users can still override `TRM_LOG_LEVEL`, `TRM_LOG_COLOR`, or `TRM_LOG_FORMAT` through `additional_envs` if needed.

ix-dev/community/restic-manager/questions.yaml
blaubaer: Removed this setting from the catalog questions and template.

The upstream app default is sufficient now, and users can still override `TRM_TLS_INSECURE` through `additional_envs` if needed.

ix-dev/community/restic-manager/questions.yaml
blaubaer: Changed the default WebUI port to `30446`, selected from the catalog port range.

ix-dev/community/restic-manager/templates/docker-compose.yaml
blaubaer: Updated this to use the configured port everywhere.

`TRM_LISTEN`, the container target port, the published port, and the portal now all use `values.network.web_port.port_number`.

ix-dev/community/restic-manager/templates/docker-compose.yaml
blaubaer: `rslave` is intentional here.

The app can create a ZFS snapshot via RPC API `pool.snapshot.create`, back up from the snapshot path, and then delete it via `pool.snapshot.delete`. With a plain read-only bind mount, accessing the ZFS snapshot from inside the container can leave the snapshot mounted in a way that prevents clean deletion (also shutting down the container does not solve the issue).

Using `{defer: true}` for `pool.snapshot.delete` only hides the immediate error. The snapshot may still remain mounted, and TrueNAS can later report that it could not be deleted.

`rslave` makes the host-side snapshot mount visible inside the container without propagating container-side mount events back to the host. Since the app only reads immutable snapshot contents (and nothing else), this avoids the cleanup issue without requiring read-write mounts or a host-side `umount` workaround.

ix-dev/community/restic-manager/ix_values.yaml
blaubaer: Removed this from `ix_values.yaml`.

The value is still used by the app as `TRM_APP_MOUNTS_PROPERTY=storage.dataset_mounts`. It tells Restic Manager which app configuration field contains the dataset mounts it should verify for selected backup sources. See: https://github.com/truenas/apps/pull/5254#discussion_r3520153869

blaubaer:
Thanks for the review. I cleaned up the exposed settings and adjusted the port handling.

The app now uses upstream `v0.2.1`, removes the catalog-level TLS/log options, uses a list for trusted upstream addresses, defaults to port `30446`, and uses the configured port consistently for the listener, compose port mapping, and portal.

ix-dev/community/restic-manager/templates/docker-compose.yaml
stavros-k: Yea, I don't want such kind of actions coming from an app in the catalog.

blaubaer: @stavros-k, thanks for clarifying.

Is the concern specifically `rslave`, or the app calling TrueNAS API methods like `pool.snapshot.create` / `pool.snapshot.delete`?

The snapshot lifecycle is the core of this app: it creates a temporary ZFS snapshot before a backup, backs up from that immutable snapshot path, and deletes the snapshot afterwards. This avoids backing up a live, changing filesystem tree.

`rslave` is only used so that this snapshot path is visible inside the container and can be cleaned up correctly afterwards.

If this is not acceptable for catalog apps, do you have a recommended alternative for consistent ZFS snapshot-based backups from a catalog app? I have not found a cleaner one so far.

Thanks for your help and for guiding me through the catalog constraints.

stavros-k: rslave is not the main reason, but it is one, sadly I cannot remember what the issues I've seen caused by that. But I remember I was debugging a ticket for 3 days.

But an app that is frankly *just* appeared on the internet, requiring Full Admin privileges (wether this is privileged container or api), is hard to accept on the catalog.
I'm not saying that this is a malicious app. But given the huge uptick of vibe coded apps, the only defense the catalog provides is limiting access on what each app gets access to.

And sadly a big portion of the users, just hit install and provide any info requested.

blaubaer: @stavros-k, thanks for taking the time to explain the reasoning, I really do appreciate it.

I understand your concerns, and I hope you can also see it a bit from my side: an app whose whole purpose is to back up the system it runs on naturally needs the corresponding access to that system. That's inherent to what a backup tool does, not a design shortcut I picked.

Concretely, Restic Manager needs read-only filesystem access to the datasets a user selects as backup sources, so it can back up from consistent ZFS snapshot paths. The backup runs inside the container and has to read those files, so this access really can't be removed without removing the whole point of the app.

To be transparent, I've already put quite a lot of my free time into this, and I'd really rather not spend more of it polishing individual details if the actual answer lies somewhere else entirely. So I'd be genuinely grateful for a clear steer: do you have a suggestion for how I could address this within the catalog's constraints, or are backup apps of this kind simply not a fit for the catalog as a matter of policy?

If it's the latter, that's completely fair and I fully respect it, I'll just document a clean custom-app path for the users who want it; outside the catalog.

Thanks again, and thanks for bearing with me on this one.

stavros-k: At this moment, there are no plans to allow such access for apps in the catalog.
What containers users run (ie custom app etc) are up to them, but via the catalog this access is not currently something I can accept. 
So if this app main purpose require those, then yea it cannot be accepted on the catalog as of today.

blaubaer:
@stavros-k, Thanks for the clear answer (https://github.com/truenas/apps/pull/5254#discussion_r3528969867). That gives me the guidance I needed.

Given that snapshot-based backups are the main purpose of the app, and that they inherently require read-only access to selected datasets plus TrueNAS API operations, I don't think there is a meaningful catalog-compatible version of this app right now.

I do think this leaves an awkward outcome: users who want this functionality now have to run it outside the catalog, rather than through a reviewed and explicitly documented catalog path. But I understand that this is the current boundary.

I'll close this PR and focus on documenting a clean custom-app installation path instead. That way users who explicitly want this kind of high-trust backup integration can choose to run it outside the catalog.

Thanks for reviewing this and for clarifying the current catalog boundary.

Issue #5229, opened 2026-06-23
[Enhancement]: bump wger to 2.6 because of breaking changes in mobile app


sebastian-muthwill:
### App Name

wger

### Current Version (TrueNAS Catalog App Version)

1.2.26

### Enhancement Description

Bump image from 2.5-dev to 2.6

The wger app (mobile) has a new release 2.6 that is incompatible with the current App Version: 2.5-dev resulting in not working mobile app.

### Use Case

Mobile app not working with older server versions. 

### Proposed Solution

Bump server image to wger/server:2.6

### Alternatives Considered

Downgreade mobile version to 2.5 or not upgrade to 2.6 until server version has been bumped.

stavros-k:
It seems there are breaking changes so its not just an image bump.

Etregin:
@stavros-k I can start working on it unless you already started

Etregin:
@sebastian-muthwill I originally came in to open the same issue

sebastian-muthwill:
@Etregin thar would be awsome. I had no chance to look into it.

Etregin:
PR https://github.com/truenas/apps/pull/5255 completed but a reinstall of the app will be needed due to the changes, you can reuse the old storage and it "should" work

dj21d:
Question on this. I reinstalled my app based on the PR instructions, but now my mobile app returns the following:

Sync Service Unreachable
Your wger server is reachable, but the (separate) service that handles offline sync is not responding. If you self-host wger, check your configuration and make sure the PowerSync service is running correctly.

I can create a separate issue, but just wanted to check quickly if this is something I overlooked related to the new upgrade. I did try connecting directly to the IP/port of the app, to make sure it wasn't something in my NGINX blocking it

Etregin:
@dj21d I think this has to do with services that needs to be started within wger itself according to the documentation

https://wger.readthedocs.io/en/latest/administration/powersync.html

Not sure if this document is related to docker installation though as I see references for the service in the docker-compose.yaml example on their docker repo so I'm looking deeper into it as I facing the same issue

Phyxsius7:
hi @Etregin ,

I noticed that the current TrueNAS wger 2.6.0 app starts the regular wger containers, but no PowerSync container is created.

Current TrueNAS containers:

- ix-wger-server-1
- ix-wger-nginx-1
- ix-wger-worker-1
- ix-wger-beat-1
- ix-wger-postgres-1
- ix-wger-redis-1

[...]

PR #5255, opened 2026-06-26
Update wger from 2.5dev to 2.6

Etregin:
* feat(wger): upgrade wger from 2.5-dev to 2.6 with env var renames

* test(wger): add upgrade-from-2.5dev test values for 2.6 migration

---------

# App Addition

- [x] I have opened an [issue](https://github.com/truenas/apps/issues) to discuss this app addition before submitting this pull request.

# AI

## Description

Adds wger Version 2.6to the community train.

[Brief description of what the app does]

## App Information

- **Upstream**: https://github.com/wger-project/wger/
- **Documentation**: https://wger.readthedocs.io/en/latest/installation/docker.html
- **App Version**: 2.6

## Testing

Tested locally with:

- [x] basic-values.yaml

## Special Notes

- Tested migrating from 2.5dev to 2.6 using ix-dev/community/wger/templates/test_values/upgrade-from-25dev-values.yaml and CI workflow passed with no issues

## Checklist

- [x] README.md included


Etregin:
Issue https://github.com/truenas/apps/issues/5229

ix-dev/community/wger/templates/test_values/upgrade-from-25dev-values.yaml
stavros-k, line 1: There is no upgrade test harness in place, so I'm not sure what this file tests here?

ix-dev/community/wger/app.yaml
```diff
@@ -1,4 +1,4 @@
-app_version: 2.5-dev
+app_version: 2.6
```
stavros-k: should be x.y.z (ie 2.6.x)

ix-dev/community/wger/questions.yaml
```diff
@@ -68,13 +68,28 @@ questions:
...
+            type: string
+            default: ""
+            required: true
+            private: true
+        - variable: jwt_public_key
+          label: JWT Public Key
+          description: |
+            Base64url-encoded RSA public key (JWK format) used for verifying JWT tokens.</br>
+            Generate a key pair by running:</br>
+            <code>docker compose exec web ./manage.py generate-jwt-keys</code></br>
```
stavros-k: This command wont produce anything. As it requires a running wger instance and also a compose file existing.

ix-dev/community/wger/README.md
stavros-k, line 1: what exactly are these doing here? 99% of it does not apply on the app package.

ix-dev/community/wger/migrations/upgrade_to_26
```diff
@@ -0,0 +1,35 @@
...
+    wger = values.get("wger", {})
+    if isinstance(wger, dict) and "signing_key" in wger:
+        wger["jwt_private_key"] = wger.pop("signing_key")
+        wger.setdefault("jwt_public_key", "")
+        values["wger"] = wger
+
+    # wger 2.6: SITE_URL and REFRESH_TOKEN_LIFETIME are new required/recommended
+    # fields. Set safe empty defaults so existing users are prompted to fill them in.
+    if isinstance(wger, dict):
+        wger.setdefault("site_url", "")
```
stavros-k: site url is required. `""` will fail that validation. 
same for the jwt

stavros-k: This migration will make the upgrade fail.

Etregin: I will remove the site_url defaults but how do you believe we should approach this? so the migration wouldn't make the upgrade fail, should this be added as a Separate app instead and then remove the old one futher down the line or is there any other approach that can be taken?

ix-dev/community/wger/questions.yaml
Etregin: Updated with the following command that can be run without a wger stack

```
docker run --rm wger/server:2.6 python manage.py generate-jwt-keys
```

ix-dev/community/wger/README.md
Etregin: Reverted readme to the basic output

ix-dev/community/wger/migrations/upgrade_to_26
stavros-k: Given the changes **requires** values to start that we can't derive, I dont think there is a way to migrate it.
We can only do the changes, and push a major version (ie breaking change).
Users will have to re-install, and point to the existing storage(s).

It _should_ continue where it left off.

Etregin: @stavros-k Thanks for the update, if thats the case, please review my latest updates to the files and let me know if there is anything else that needs adjustments, since it will be considered a "breaking change" either way i guess i should remove the migration script?

ix-dev/community/wger/templates/docker-compose.yaml
```diff
@@ -50,7 +50,10 @@
 {# -- Common Environment Variables -- #}
 {% for c in containers %}
   {% do c.environment.add_env("SECRET_KEY", values.wger.secret_key) %}
-  {% do c.environment.add_env("SIGNING_KEY", values.wger.signing_key) %}
+  {% do c.environment.add_env("JWT_PRIVATE_KEY", values.wger.jwt_private_key) %}
+  {% do c.environment.add_env("JWT_PUBLIC_KEY", values.wger.jwt_public_key) %}
+  {% do c.environment.add_env("SITE_URL", values.wger.site_url) %}
+  {% do c.environment.add_env("REFRESH_TOKEN_LIFETIME", values.wger.refresh_token_lifetime) %}
```
stavros-k, line 56: Is this needed? (ie doesnt upstream have a sane default set?) 
Seems like a tunable thing, I dont think we shuold expose it

ix-dev/community/wger/ix_values.yaml
```diff
@@ -1,8 +1,7 @@
 images:
   image:
     repository: wger/server
-    # TODO: Pin to stable once 2.5 is released
-    tag: 2.5-dev
+    tag: 2.6
```
stavros-k: use 2.6.x..

ix-dev/community/wger/questions.yaml
```diff
@@ -68,13 +68,28 @@ questions:
...
             required: true
             private: true
-        - variable: signing_key
-          label: Signing Key
+        - variable: jwt_private_key
+          label: JWT Private Key
+          description: |
+            Base64url-encoded RSA private key (JWK format) used for signing JWT tokens.</br>
+            Generate a key pair by running the following command from the TrueNAS shell:</br>
+            <code>docker run --rm wger/server:2.6 python manage.py generate-jwt-keys</code></br>
```
stavros-k: Does not have to be truenas shell, it can be anywhere that has docker.
Also no need for `code`, backticks work

ix-dev/community/wger/app.yaml
```diff
@@ -58,4 +58,4 @@ sources:
 - https://github.com/wger-project/wger
 title: Wger
 train: community
-version: 1.2.26
+version: 1.3.0
```
stavros-k: 2.0.0
And drop the migration

ix-dev/community/wger/questions.yaml
```diff
@@ -68,13 +68,28 @@ questions:
...
+            Copy the value of JWT_PRIVATE_KEY from the output and paste it here.
+          schema:
+            type: string
```
stavros-k, line 79: default string max_length is 1024 if I remember correctly.
From tests values I see its much longer, so it will probabl yneeds to be bumped up

stavros-k (commented):
.

ix-dev/community/wger/templates/docker-compose.yaml
Etregin: @stavros-k According to the release notes for 2.6.0 its recommended to set the REFRESH_TOKEN_LIFETIME to a minimum of 3000, i dont think its set upstream 

```
JWT refresh tokens now rotate on every use. Combined with blacklist-after-rotation this makes refresh tokens single-use and lets a leaked token live only until the next refresh call. It's recommended to update your prod.env and set REFRESH_TOKEN_LIFETIME to a value like 3000
```

ix-dev/community/wger/ix_values.yaml
Etregin: My Apologies, forgot to adjust it here

stavros-k (approved):

Etregin:
@stavros-k I believe the last check is failing due to the timeout set

stavros-k:
> @stavros-k I believe the last check is failing due to the timeout set

Does it actually take more than 5min to start? If thats the case, it will cause more issues on actual systems out there. I've entire racks that boot faster than that :P

Etregin:
I've tested even with the old config on 2.5Dev and it was failing with the exact error so yea I guess it takes more than 5 minutes :D it does seem odd but on initial setup it extracts and provisions allot of excersises so it takes a while, the command to add the ingridiants after you spin up the stack downloaded in 15 minutes and then spent 6 hours extracting them

stavros-k:
> I've tested even with the old config on 2.5Dev and it was failing with the exact error so yea I guess it takes more than 5 minutes :D it does seem odd but on initial setup it extracts and provisions allot of excersises so it takes a while, the command to add the ingridiants after you spin up the stack downloaded in 15 minutes and then spent 6 hours extracting them

Thats very weird for an app. So you cant use the app at all until all that is done?

stavros-k:
Huh, even a restart that supposedly have all the assets downloaded and should just check takes same time...
Yea we should remove the startup sync variables, add 2 words in the notes on how to trigger the sync within the container. 

I dont think anyone would like their app to take > 20min to start.

PR #5262, opened 2026-06-29
palword: allow configuring the api port

### LINKED ISSUE #5253: [Enhancement]: Palworld App REST API Port Cannot Be Exposed
### App Name

Palworld

1.3.6

### Enhancement Description

[RCON is technically deprecated](https://docs.palworldgame.com/api/rcon/) and the [REST API](https://docs.palworldgame.com/api/rest-api/palwold-rest-api/) will instead be supported going forward.

The Network Configuration section has a spot for a server port and a RCON port, but nowhere to specify a REST API port (see the screenshot I took):
Image

### Use Case

Adding a field for the port would allow it to be exposed, which is helpful so that Python3 code and/or curl commands from outside the container can access the REST API if the user desires.

### Proposed Solution

ix-dev/community/palworld/questions.yaml needs to be modified to include a field for the REST API Port, and ideally this would be inserted into the Pal/Saved/Config/LinuxServer/PalWorldSettings.ini file via the following settings:
RESTAPIEnabled=True
RESTAPIPort=[Selected_Port_Number]

### Alternatives Considered

I tried to use a docker exec command to execute a REST API call inside of the container itself, but the container is very stripped down and does not have Python3 or curl in it. Even the following BASH script:
```
dockerid=$(sudo docker ps -q --filter="name=palworld")
sudo docker exec -it $dockerid bash -c '
exec 3<>/dev/tcp/127.0.0.1/[REST_API_PORT_MANUALLY_ADDED_PalWorldSettings.ini]
auth=$(printf "admin:[ADMIN_PASSWORD]" | base64)
printf "GET /v1/api/players HTTP/1.1\r\nHost: 127.0.0.1:[REST_API_PORT_MANUALLY_ADDED_PalWorldSettings.ini]\r\nAuthorization: Basic $auth\r\nConnection: close\r\n\r\n" >&3
cat <&3
'
```
...didn't work despite successfully hitting the REST API.

### Additional Context

_No response_

stavros-k:
Closes #5253

### DIFF
--- a/ix-dev/community/palworld/templates/macros/setup.sh.jinja
+++ b/ix-dev/community/palworld/templates/macros/setup.sh.jinja
@@ -50,6 +50,10 @@ set_ini_value "ServerDescription" "{{ values.palworld.server.description }}" tru
 set_ini_value "ServerPassword" '{{ values.palworld.server.password }}' true false
 set_ini_value "AdminPassword" '{{ values.palworld.admin_password }}' true false
 set_ini_value "AllowConnectPlatform" '{{ values.palworld.allow_platform }}' false
+set_ini_value "RESTAPIEnabled" '{{ values.networks.api_port.bind_mode != "" }}' false
+{% if values.networks.api_port.bind_mode %}
+set_ini_value "RESTAPIPort" '{{ values.networks.api_port.port_number }}' false
+{% endif %}
 
 echo "Setting user defined ini keys..."
 {%- for item in values.palworld.ini_keys %}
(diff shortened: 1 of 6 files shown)


ix-dev/community/palworld/templates/macros/setup.sh.jinja
```diff
@@ -50,6 +50,10 @@ set_ini_value "ServerDescription" "{{ values.palworld.server.description }}" tru
 set_ini_value "ServerPassword" '{{ values.palworld.server.password }}' true false
 set_ini_value "AdminPassword" '{{ values.palworld.admin_password }}' true false
 set_ini_value "AllowConnectPlatform" '{{ values.palworld.allow_platform }}' false
+set_ini_value "RESTAPIEnabled" '{{ values.networks.api_port.bind_mode != "" }}' false
+{% if values.networks.api_port.bind_mode %}
+set_ini_value "RESTAPIPort" '{{ values.networks.api_port.port_number }}' false
+{% endif %}
```
stavros-k:
```suggestion
set_ini_value "RESTAPIEnabled" '{{ values.network.api_port.bind_mode != "" }}' false
{% if values.network.api_port.bind_mode %}
set_ini_value "RESTAPIPort" '{{ values.network.api_port.port_number }}' false
{% endif %}
```

Qubad786 (approved):

PR #5304, opened 2026-07-04
Adding PowerSync container to wger

### LINKED ISSUE #5298: [Enhancement]: Add Powersync container to wger
### App Name

wger

2.0.2

### Enhancement Description

Need to add the PowerSync container / service upon deploying wger

### Use Case

This is required for the mobile app to work properly

### Proposed Solution

Add all the required config files and add PowerSync container to the config

### Alternatives Considered

_No response_

### Additional Context

_No response_

Etregin:
# App Addition

- [x] I have opened an [issue](https://github.com/truenas/apps/issues/5298) to discuss this app addition before submitting this pull request.

# AI

## Description

Adds powersync container to wger as required in V2.6

## App Information

- **Upstream**: https://github.com/wger-project/docker
- **Documentation**: Documentation: https://wger.readthedocs.io/en/latest/installation/docker.html
- **App Version**: 2..6.0

## Testing

Tested locally with:

- [x] basic-values.yaml

## Special Notes

- First-time setup instructions: user needs to run `python3 manage.py setup-powersync-storage` from within the server container 
- Adjusted first run for server to 300 to avoid CI/CD failing to start the container (As set upstream)
- Added powersync.yaml and sync_rules.yaml as macros as set upstream and to avoid crouting docker-compose.yaml

## Checklist

- [x] README.md included

Closes #5298

### DIFF
--- a/ix-dev/community/wger/app.yaml
+++ b/ix-dev/community/wger/app.yaml
@@ -13,8 +13,8 @@ keywords:
 - nutrition
 - tracker
 - workout
-lib_version: 2.3.4
-lib_version_hash: 2e3a8847308fb2eb0da046018f287c73822c094b5950a10377c3235794ff1242
+lib_version: 2.3.8
+lib_version_hash: cd75c897a1e8fef54b5bd00d0d8849f240bc50db2ef650eccc0ee74f3b2b2dc1
 maintainers:
 - email: dev@truenas.com
   name: truenas
@@ -58,4 +58,4 @@ sources:
 - https://github.com/wger-project/wger
 title: Wger
 train: community
-version: 2.0.5
+version: 2.0.6
--- a/ix-dev/community/wger/templates/docker-compose.yaml
+++ b/ix-dev/community/wger/templates/docker-compose.yaml
@@ -1,4 +1,5 @@
 {% from "macros/nginx.conf" import nginx_config %}
+{% from "macros/powersync.conf" import powersync_yaml %}
 {% set tpl = ix_lib.base.render.Render(values) %}
 
 {# {% if values.wger.postgres_image_selector == "postgres_17_image" %}
@@ -24,6 +25,7 @@
   "password": values.wger.db_password,
   "database": values.consts.db_name,
   "volume": values.storage.postgres_data,
+  "additional_options": {"wal_level": "logical"},
 } %}
 {% set postgres = tpl.deps.postgres(values.consts.postgres_container_name, values.wger.postgres_image_selector, pg_config, perm_container) %}
 {% do postgres.container.add_network(wger_net) %}
@@ -35,6 +37,26 @@
 } %}
 {% set redis = tpl.deps.redis(values.consts.redis_container_name, "redis_image", redis_config, perm_container) %}
 {% do redis.container.add_network(wger_net) %}
+
+{# -- PowerSync (Conditional) -- #}
+{% if values.wger.powersync_enabled %}
+  {# PowerSync service — both configs injected at render time via Docker configs #}
+  {% set powersync = tpl.add_container(values.consts.powersync_container_name, "powersync_image") %}
+  {% do powersync.set_user(values.consts.run_as_user, values.consts.run_as_group) %}
+  {% do powersync.set_command(["start", "-r", "unified"]) %}
+  {% do powersync.healthcheck.set_test("node", {"port": values.consts.powersync_port, "path": "/probes/readiness"}) %}
+  {% do powersync.healthcheck.set_start_period(120) %}
+  {% do powersync.healthcheck.set_retries(10) %}
+  {% do powersync.depends.add_dependency(values.consts.postgres_container_name, "service_healthy") %}
+  {% do powersync.depends.add_dependency(values.consts.server_container_name, "service_healthy") %}
+  {% do powersync.add_network(wger_net) %}
+
+  {# Both configs injected directly — no storage volume or init container needed #}
+  {% do powersync.configs.add("powersync.yaml", powersync_yaml(values, postgres), "/config/powersync.yaml") %}
+  {% do powersync.configs.add("sync_rules.yaml", values.wger.powersync.sync_rules_content, "/config/sync_rules.yaml") %}
+
+  {% do powersync.environment.add_env("POWERSYNC_CONFIG_PATH", "/config/powersync.yaml") %}
+{% endif %}
 
 {% set containers = [server, worker, beat] %}
 {% for c in containers + [nginx] %}
@@ -88,7 +110,19 @@
 
 {# -- Server -- #}
 {% do server.healthcheck.set_test("wget", {"port": values.consts.internal_server_port}) %}
+{% do server.healthcheck.set_start_period(300) %}
+{% do server.healthcheck.set_retries(10) %}
 {% do server.environment.add_env("WGER_PORT", values.consts.internal_server_port) %}
+{% if values.wger.powersync_enabled %}
+  {# PS_STORAGE_PG_URI is read by setup-powersync-storage to create the role with the correct password #}
+  {% do server.environment.add_env("PS_STORAGE_PG_URI",
+    "postgres://powersync_storage:%s@%s:%s/%s" | format(
+      values.wger.powersync_storage_password,
+      values.consts.postgres_container_name,
+      postgres.get_port(),
+      values.consts.db_name
+    )) %}
+{% endif %}
 {% do server.depends.add_dependency(values.consts.postgres_container_name, "service_healthy") %}
 {% do server.depends.add_dependency(values.consts.redis_container_name, "service_healthy") %}
 {% do server.add_storage("/home/wger/static", values.storage.static) %}
@@ -96,12 +130,16 @@
 
 {# -- Worker -- #}
 {% do worker.healthcheck.set_custom_test(["CMD", "celery", "-A", "wger", "inspect", "ping"]) %}
+{% do worker.healthcheck.set_timeout(15) %}
+{% do worker.healthcheck.set_start_period(300) %}
 {% do worker.set_command(["/start-worker"]) %}
 {% do worker.depends.add_dependency(values.consts.server_container_name, "service_healthy") %}
 {% do worker.add_storage("/home/wger/media", values.storage.media) %}
 
 {# -- Beat -- #}
 {% do beat.healthcheck.set_custom_test(["CMD", "celery", "-A", "wger", "inspect", "ping"]) %}
+{% do beat.healthcheck.set_timeout(15) %}
+{% do beat.healthcheck.set_start_period(60) %}
 {% do beat.set_command(["/start-beat"]) %}
 {% do beat.depends.add_dependency(values.consts.worker_container_name, "service_healthy") %}
 {% do beat.add_storage("/home/wger/beat", values.storage.beat) %}
--- a/ix-dev/community/wger/templates/macros/powersync.conf
+++ b/ix-dev/community/wger/templates/macros/powersync.conf
@@ -1,0 +1,31 @@
+{% macro powersync_yaml(values, postgres) -%}
+telemetry:
+  disable_telemetry_sharing: true
+  prometheus_port: 9090
+
+replication:
+  connections:
+    - type: postgresql
+      uri: "postgres://{{ values.consts.db_user }}:{{ values.wger.db_password }}@{{ values.consts.postgres_container_name }}:{{ postgres.get_port() }}/{{ values.consts.db_name }}"
+      sslmode: disable
+
+storage:
+  type: postgresql
+  uri: "postgres://powersync_storage:{{ values.wger.powersync_storage_password }}@{{ values.consts.postgres_container_name }}:{{ postgres.get_port() }}/{{ values.consts.db_name }}"
+  sslmode: disable
+
+port: {{ values.consts.powersync_port }}
+
+api:
+  parameters:
+    max_parameter_query_results: 2000
+
+sync_rules:
+  path: sync_rules.yaml
+
+client_auth:
+  allow_local_jwks: true
+  jwks_uri: "http://{{ values.consts.server_container_name }}:{{ values.consts.internal_server_port }}/api/v2/powersync-keys"
+  audience:
+    - "powersync"
+{%- endmacro %}
--- a/ix-dev/community/wger/ix_values.yaml
+++ b/ix-dev/community/wger/ix_values.yaml
@@ -14,6 +14,9 @@ images:
   postgres_18_image:
     repository: postgres
     tag: 18.4-trixie
+  powersync_image:
+    repository: journeyapps/powersync-service
+    tag: "1.23.2"
   redis_image:
     repository: valkey/valkey
     tag: 9.1.1
@@ -29,6 +32,8 @@ consts:
   postgres_container_name: postgres
   redis_container_name: redis
   perms_container_name: permissions
+  powersync_container_name: powersync
+  powersync_port: 8080
   db_user: wger
   db_name: wger
 
@@ -45,3 +50,41 @@ consts:
     python3 manage.py download-exercise-images
     python3 manage.py download-exercise-videos
     ```
+
+    ## PowerSync (mobile app sync)
+
+    PowerSync requires a one-time setup before it can start successfully.
+    Until this is done, the `powersync` container will log an authentication
+    error (`password authentication failed for user "powersync_storage"`) and
+    restart — this is expected.
+
+    Shell into the `server` container and run:
+
+    ```sh
+    python3 manage.py setup-powersync-storage
+    ```
+
+    This creates the required Postgres role, schema, and replication slot.
+    It is idempotent and safe to re-run. If you change `powersync_storage_password`,
+    update it in the app settings first and then re-run this command.
+
+    ### Sync rules
+
+    The `sync_rules.yaml` file controls which data is replicated to the mobile app.
+    Its content is provided via the **Sync Rules Content** field in the app settings
+    and is injected directly into the container on every deploy. To change the sync
+    rules, update the field in the UI and redeploy the app.
+
+    The default content can be found at:
+    https://github.com/wger-project/docker/blob/master/services/config-powersync/sync_rules.yaml
+
+    ### Compaction
+
+    PowerSync's replication log grows without bound unless compacted periodically.
+    Run compaction manually with:
+
+    ```sh
+    docker compose run --rm powersync compact
+    ```
+
+    A daily cron or systemd timer is recommended for production deployments.
--- a/ix-dev/community/wger/questions.yaml
+++ b/ix-dev/community/wger/questions.yaml
@@ -158,6 +158,53 @@ questions:
                       label: Value
                       schema:
                         type: string
+        - variable: powersync_enabled
+          label: Enable PowerSync
+          description: Enable offline sync support for the mobile app
+          schema:
+            type: boolean
+            default: false
+        - variable: powersync_storage_password
+          label: PowerSync Storage Password
+          description: Password for the PowerSync storage database user
+          schema:
+            type: string
+            private: true
+            required: false
+            show_if: [["powersync_enabled", "=", true]]
+        - variable: powersync
+          label: PowerSync Settings
+          schema:
+            type: dict
+            show_if: [["powersync_enabled", "=", true]]
+            attrs:
+              - variable: ps_url_path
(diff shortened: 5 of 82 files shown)


ix-dev/community/wger/templates/macros/powersync.yaml
stavros-k, line 1: yea, that wont work. you are taking ownership of the file, any edits a user do will get discarded.

ix-dev/community/wger/templates/macros/sync_rules.yaml
```diff
@@ -0,0 +1,132 @@
...
+# https://docs.powersync.com/maintenance-ops/compacting-buckets
+
+config:
+  edition: 3
+
+streams:
+  # Global reference data, shared by all users, changes rarely enough
+  core:
+    auto_subscribe: true
+    queries:
```
stavros-k: What is this thing here? 
I'm not going to have around sql code that will need babysitting on every version in schema changes

ix-dev/community/wger/templates/test_values/basic-values.yaml
```diff
@@ -17,6 +17,10 @@ wger:
   site_url: http://localhost:8080
   refresh_token_lifetime: 3000
   max_upload_size_mb: 100
+  powersync:
+    ps_storage_password: "wger_ps_storage"
```
stavros-k: Why does it needs different password?

ix-dev/community/wger/templates/docker-compose.yaml
```diff
@@ -24,6 +26,7 @@
   "password": values.wger.db_password,
   "database": values.consts.db_name,
   "volume": values.storage.postgres_data,
+  "additional_options": {"wal_level": "logical"},
```
stavros-k, line 28: Why is that needed?

ix-dev/community/wger/app.yaml
```diff
@@ -36,6 +36,11 @@ run_as_context:
   group_name: Host group is [docker]
   uid: 999
   user_name: Host user is [netdata]
+- description: Container [powersync] runs as root user and group.
+  gid: 0
```
stavros-k: Why does it need root?

ix-dev/community/wger/templates/docker-compose.yaml
```diff
@@ -96,16 +102,68 @@
...
+{% do powersync.healthcheck.set_test("tcp", {"port": ps_port}) %}
+{% do powersync.healthcheck.set_start_period(120) %}
+{% do powersync.healthcheck.set_retries(10) %}
+
+{# Pre-create the powersync_storage role on fresh postgres installs via an init script.
+   docker-entrypoint-initdb.d runs automatically on the first-ever DB initialization. #}
+{% set ps_init_sql %}
+-- Create the powersync_storage role used by the PowerSync replication service.
+-- This script runs once on a fresh database via docker-entrypoint-initdb.d.
+CREATE ROLE powersync_storage WITH LOGIN PASSWORD '{{ values.wger.powersync.ps_storage_password }}';
```
stavros-k: I don't think this will persist a postgres upgrade. Which is another failure point that I don't want to maintain.

stavros-k (changes requested):

ix-dev/community/wger/templates/macros/sync_rules.yaml
Etregin: These are the rule for the PowerSync Container, this is maintained upstream by wger and just ported here

ix-dev/community/wger/templates/test_values/basic-values.yaml
Etregin: ps_storage_password is the password for a separate dedicated Postgres role (powersync_storage) that PowerSync uses for replication

ix-dev/community/wger/templates/docker-compose.yaml
Etregin: PowerSync uses logical replication to stream database changes from Postgres in real time. The wal_level = logical setting is a hard requirement for logical replication slots — without it, PowerSync cannot create a replication slot and will fail to start. This is set in the upstream wger-project/docker repo's postgres.yaml service via command: ["postgres", "-c", "wal_level=logical"]. The additional_options key maps to the same Postgres startup parameter via the TrueNAS library's postgres helper.

ix-dev/community/wger/templates/macros/sync_rules.yaml
stavros-k: My assumptions is that this is user configurable, which is not possible with this.

ix-dev/community/wger/templates/test_values/basic-values.yaml
stavros-k: But why does it need separate role?

ix-dev/community/wger/templates/docker-compose.yaml
stavros-k: What exactly does it replicate? Its a single db, no?

Etregin: Yes its a single DB, its not replicating to another postgress DB but instead it replicates to users mobile SQLite DB to allow them to view their information in offline mode (hence its required for the mobile app to work)

ix-dev/community/wger/templates/test_values/basic-values.yaml
Etregin: I configured the file according to what was configured upstream, from my understanding this user / role has different privileges from the other wger user and has its own password to avoid confusion same as any other role in the app

ix-dev/community/wger/templates/macros/sync_rules.yaml
Etregin: Ill adjust it so it becomes more user configruable with this file as the default value

stavros-k: Just reminding, that if it wants us to maintain the list of queries, thats a non-starter.

Etregin: I dont think truenas will be maintaining anything, this file will be the "default" entry, if the user wants to adjust it then they can do so at their own risk, and even the "default" entry would be maintained upstream not by truenas

ix-dev/community/wger/app.yaml
Etregin: I tested with Other values but was facing other issues, i forgot to test without root after resolving all the issues, Will test and update you accordingly

ix-dev/community/wger/templates/docker-compose.yaml
Etregin: This was done to pass the CI/CD test, the user can configure this with a wger provided script after setup but the CI/CD test will show the PowerSync container failed to start, if you are OK with that then ill remove this part

stavros-k: That results in a non working setup. 
Also the init script won't run on existing databases, so it will still broken for most users that upgrade existing installations.

Etregin: makes sense, ill remove it and test the results but i suspose we should expect some errors with powersync container on ci/cd until the user sets this up manually within the server container

https://wger.readthedocs.io/en/latest/administration/powersync.html

Already added in the notes in the ix_values

Etregin: or even better ill add powersync as toggable option and if the user has it enabled then to add a note to run the command after the server starts with the link to the documentation and if not then it will not be setup and the container wont start

stavros-k: I'm not convinced that we can ship this without making both sides (maintainer and user) miserable.

Whats the point of having it built in if the user has to do all the work?
But in this case, we can't reliably ship it due to how it needs to be setup.

Etregin: Thats why i thought about keeping it as a toggable option, if the user knows what hes doing and willing to go through with the config to use the mobile app then they should enable the toggle, if not then its excluded and the user can use the App as is without the mobile app

stavros-k: But what does the toggle offer? A bare container that nothing works.
User have to start looking what is plumbed in and what is not, and then start configuring.

Doing half of it is almost always worse that not doing it at all.

Besides that, it still not tested that all this required things for db (wal changes, new roles, etc) survive a postgres upgrade.

Etregin: The toggle will provide the user with the container, storage for the container, the "default" config file provided by upstream with the option to edit it later, the user will just need to run 1 command to get the PowerSync service authenticated and usable on postgress 

Everything will work fine even if the user toggles the PowerSync option but doesnt run the command in the server container, the only thing that will remain broken would be the mobile app which wouldnt be working either way

As for the postgres upgrade we already have it at a fixed version, if at any point postgres version changed and upstream they updated then everything will be there for the next migration

stavros-k: There is not just 1 command.

It is:
- Few commands on the postgres container to setup the role (init dir is only run only on the first start of postgres)
- 2 config files that will also require values FROM the template.
- A command to setup powersync on the server container

In order for config files to be user servicable, we would need another init container to _patch_ those files on every startup, ensuring port, db uri, db pass, etc stay in sync.

Etregin:
@stavros-k I adjusted the The code to have powersync as a toggle option and adjusted the permissions, tested with both basic-values.yaml and powersync-values.yaml 

basic-values.yaml with powersync not toggled works with no issues, powersync-values.yaml shows that the powersync container did not start properly in the CI/CD since its restarting, once i ran the command in the server container as per wger instructions it became healthy with no issues as shown below

```
# docker ps
CONTAINER ID   IMAGE                                  COMMAND                  CREATED         STATUS                             PORTS                                         NAMES
dbf7f83c5485   wger/server:2.6.0                      "/start-beat"            4 minutes ago   Up 48 seconds (healthy)            8000/tcp                                      2bac423e3f1b7c7ccd6c69d44467120b-beat-1
07ffcdd6eca0   nginxinc/nginx-unprivileged:1.31.2     "/docker-entrypoint.…"   4 minutes ago   Up About a minute (healthy)        0.0.0.0:8080->8080/tcp, [::]:8080->8080/tcp   2bac423e3f1b7c7ccd6c69d44467120b-nginx-1
56ceb5256368   wger/server:2.6.0                      "/start-worker"          4 minutes ago   Up About a minute (healthy)        8000/tcp                                      2bac423e3f1b7c7ccd6c69d44467120b-worker-1
984fdefd8ed8   journeyapps/powersync-service:1.23.2   "node service/lib/en…"   4 minutes ago   Up 21 seconds (health: starting)                                                 2bac423e3f1b7c7ccd6c69d44467120b-powersync-1
4497a2a58857   wger/server:2.6.0                      "/home/wger/entrypoi…"   4 minutes ago   Up 4 minutes (healthy)             8000/tcp                                      2bac423e3f1b7c7ccd6c69d44467120b-server-1
00aeea37ac5b   postgres:18.4-trixie                   "docker-entrypoint.s…"   4 minutes ago   Up 4 minutes (healthy)             5432/tcp                                      2bac423e3f1b7c7ccd6c69d44467120b-postgres-1
6b9104a26a2f   valkey/valkey:9.1.0                    "docker-entrypoint.s…"   4 minutes ago   Up 4 minutes (healthy)             6379/tcp                                      2bac423e3f1b7c7ccd6c69d44467120b-redis-1
# docker exec -it 4497a2a58857 sh
$ python3 manage.py setup-powersync-storage
level=INFO ts=2026-07-08 14:54:18,050 module=apps path=/home/wger/.local/lib/python3.12/site-packages/axes/apps.py line=53 message=AXES: BEGIN version 8.3.1, blocking by ip_address
PowerSync storage ready: role 'powersync_storage', schema 'powersync' in 'wger'.
$ exit
# docker restart 984fdefd8ed8
984fdefd8ed8
# docker ps
CONTAINER ID   IMAGE                                  COMMAND                  CREATED         STATUS                        PORTS                                         NAMES
dbf7f83c5485   wger/server:2.6.0                      "/start-beat"            6 minutes ago   Up 3 minutes (healthy)        8000/tcp                                      2bac423e3f1b7c7ccd6c69d44467120b-beat-1
07ffcdd6eca0   nginxinc/nginx-unprivileged:1.31.2     "/docker-entrypoint.…"   6 minutes ago   Up 3 minutes (healthy)        0.0.0.0:8080->8080/tcp, [::]:8080->8080/tcp   2bac423e3f1b7c7ccd6c69d44467120b-nginx-1
56ceb5256368   wger/server:2.6.0                      "/start-worker"          6 minutes ago   Up 3 minutes (healthy)        8000/tcp                                      2bac423e3f1b7c7ccd6c69d44467120b-worker-1
984fdefd8ed8   journeyapps/powersync-service:1.23.2   "node service/lib/en…"   6 minutes ago   Up About a minute (healthy)                                                 2bac423e3f1b7c7ccd6c69d44467120b-powersync-1
4497a2a58857   wger/server:2.6.0                      "/home/wger/entrypoi…"   6 minutes ago   Up 6 minutes (healthy)        8000/tcp                                      2bac423e3f1b7c7ccd6c69d44467120b-server-1
00aeea37ac5b   postgres:18.4-trixie                   "docker-entrypoint.s…"   6 minutes ago   Up 6 minutes (healthy)        5432/tcp                                      2bac423e3f1b7c7ccd6c69d44467120b-postgres-1
6b9104a26a2f   valkey/valkey:9.1.0                    "docker-entrypoint.s…"   6 minutes ago   Up 6 minutes (healthy)        6379/tcp                                      2bac423e3f1b7c7ccd6c69d44467120b-redis-1
```

ix-dev/community/wger/templates/macros/powersync.conf
```diff
@@ -0,0 +1,147 @@
...
+
+client_auth:
+  allow_local_jwks: true
+  jwks_uri: "http://{{ values.consts.server_container_name }}:{{ values.consts.internal_server_port }}/api/v2/powersync-keys"
+  audience:
+    - "powersync"
+{%- endmacro %}
+
+
+{% macro sync_rules_yaml() -%}
```
stavros-k: As I've said previously. I don't want these here.
These queries are something that can change on a patch/minor version, and we never know cause it will not even break the installation.

stavros-k:
> @stavros-k I adjusted the The code to have powersync as a toggle option and adjusted the permissions, tested with both basic-values.yaml and powersync-values.yaml
> 
> basic-values.yaml with powersync not toggled works with no issues, powersync-values.yaml shows that the powersync container did not start properly in the CI/CD since its restarting, once i ran the command in the server container as per wger instructions it became healthy with no issues as shown below

I don't think this is viable. 
1. Requires carrying around files that have nothing to do with the deployment part, but rather with application, that need to be reviewed and babysitted even on minor or patch releases.
2. Requires customization on the postgres database beyond what is currently supported on the library.
3. due to the above, can no longer know if a future postgres upgrade will work and if it will break or continue silently and cause more issues.
4. have no ability to reliably test it on CI. (yes there are other apps that are not tested on CI, like vpn clients, but those are stateless apps).

ix-dev/community/wger/templates/macros/powersync.conf
Etregin: Without these here the PowerSync will fail to start, these are only written if the file does not exist and PowerSync is enabled , should not be breaking and this is part of the config upstream

If the version is at any point updated whoever is updating it would be checking all the config changes anyway so not sure what the issue is here

stavros-k: > If the version is at any point updated whoever is updating it would be checking all the config changes anyway so not sure what the issue is here

What do you mean? Do you expect on every wger update to go fish all the commits to see if the file is changed?

Etregin:
> > @stavros-k I adjusted the The code to have powersync as a toggle option and adjusted the permissions, tested with both basic-values.yaml and powersync-values.yaml
> > basic-values.yaml with powersync not toggled works with no issues, powersync-values.yaml shows that the powersync container did not start properly in the CI/CD since its restarting, once i ran the command in the server container as per wger instructions it became healthy with no issues as shown below
> 
> I don't think this is viable.
> 
> 1. Requires carrying around files that have nothing to do with the deployment part, but rather with application, that need to be reviewed and babysitted even on minor or patch releases.
> 2. Requires customization on the postgres database beyond what is currently supported on the library.
> 3. due to the above, can no longer know if a future postgres upgrade will work and if it will break or continue silently and cause more issues.
> 4. have no ability to reliably test it on CI. (yes there are other apps that are not tested on CI, like vpn clients, but those are stateless apps).

1- its part of the config, as far as i understand on version changes configs are revisited to ensure they are up to date and the application would work, doesnt really matter if they are just config file for initial querys that needs to run

2- Yes we cant test with CI but if you recall wger was failing the CI without powersync, to resolve this i adjusted the start time for the server to match the upstream for it to pass the CI

3- If we will not accomodate for the PowerSync configs even a little then having the app on the platform would be a hassle because one of the main functions which is the mobile app will not function at all

ix-dev/community/wger/templates/macros/powersync.conf
Etregin: No but if there is a breaking update the maintainer updates the release notes that there are breaking changes and guiding on what needs to be updated / changed like this did on version 2.6

stavros-k:
> 1- its part of the config, as far as i understand on version changes configs are revisited to ensure they are up to date and the application would work, doesnt really matter if they are just config file for initial querys that needs to run

This does not contain just application config, ie a port or whatever. This contains queries that are based on the apps db schema. A schema can be automatically migrated on a minor version and leave these queries broken. That schema change usually doesnt even hit the changelog.

> 2- Yes we cant test with CI but if you recall wger was failing the CI without powersync, to resolve this i adjusted the start time for the server to match the upstream for it to pass the CI

That's not true. It was previously passing https://github.com/truenas/apps/pull/4676

> 3- If we will not accomodate for the PowerSync configs even a little then having the app on the platform would be a hassle because one of the main functions which is the mobile app will not function at all

"Even a little" is an nginx config, maybe an init container to run some migrations.
But having to actively watch every change even on every release, its not little.

Etregin:
> > 1- its part of the config, as far as i understand on version changes configs are revisited to ensure they are up to date and the application would work, doesnt really matter if they are just config file for initial querys that needs to run
> 
> This does not contain just application config, ie a port or whatever. This contains queries that are based on the apps db schema. A schema can be automatically migrated on a minor version and leave these queries broken. That schema change usually doesnt even hit the changelog.
> 
> > 2- Yes we cant test with CI but if you recall wger was failing the CI without powersync, to resolve this i adjusted the start time for the server to match the upstream for it to pass the CI
> 
> That's not true. It was previously passing #4676
> 
> > 3- If we will not accomodate for the PowerSync configs even a little then having the app on the platform would be a hassle because one of the main functions which is the mobile app will not function at all
> 
> "Even a little" is an nginx config, maybe an init container to run some migrations. But having to actively watch every change even on every release, its not little.

When I started working on updating wger to V2.6 V2.5Dev was already failing CI/CD but that doesnt really matter right now.

How do you suggest we proceed with this?

stavros-k:
Given the requirements this has, I don't have something to propose at the current instant.

Etregin:
> Given the requirements this has, I don't have something to propose at the current instant.

I think its better to have it as optional and "might" be broken that not to have it at all, those queries can be adjusted by the user if its broken and this part requires manual maintenance by whoever uses it anyway so it has nothing to do with the truenas maintainers in my opinion.

Eventually its up to you

Etregin:
> Given the requirements this has, I don't have something to propose at the current instant.

I just got an idea, how about while having the PowerSync As a toggle (optional) when toggled the user needs to provide the sync-rules.yaml content with a info note directing to the wger repo for them to configure it, would that work?

Etregin:
@stavros-k Please let me know if my last suggestion is valid or not, ill wait for your update before starting to work on it as i dont want to waste time on this and then have it rejected again

innocentiv:
@stavros-k , do you have any feedback on this? I am one of the users getting blocked from using the app by the lack of this feature. Is there any way I can help you?

Etregin:
@stavros-k any update on this? We are still pending an update on this as if this is not implemented that 50% of the benefit of the app is no longer functional unless the user instead installs it as a 3rd party app and if thats the case then wger should be removed from the catalog

Etregin:
@stavros-k any update on this? its been almost 2 months now

Etregin:
@stavros-k do we plan on getting any sort of response on this or just going to ignore the PR forever?

PR #5307, opened 2026-07-04
feat(nextcloud): add ability to install custom PHP extensions

cedricziel:
## Summary

Adds a **Custom PHP Extensions** option to the Nextcloud app so users can install additional PHP extensions into the custom-built image, without having to wait for them to be added to the curated `apt_packages`/`consts.packages` list.

## Changes

- **questions.yaml**: new `nextcloud.php_extensions` list. Each entry has a `name` and an `installer` type:
  - `pecl` — installed with `pecl install` and enabled with `docker-php-ext-enable` (e.g. `ds`, `xdebug`)
  - `core` — compiled with `docker-php-ext-install -j"$(nproc)"` (e.g. `calendar`, `sockets`)
- **templates/docker-compose.yaml**: extensions are appended as `RUN` steps to the same inline Dockerfile that already handles `apt_packages`, so they apply to the Nextcloud and cron containers alike. Two safeguards:
  - Every step is guarded with `php -m | grep -qix <name> ||` so extensions already shipped in the image (`apcu`, `bcmath`, `memcached`, `redis`, `imagick`, …) are skipped — `pecl install` exits non-zero for already-installed extensions and would otherwise break the build.
  - Extension names are validated at render time against a strict character whitelist (`a-z A-Z 0-9 . _ -`); anything else fails the render, so user input cannot break out of the generated Dockerfile.
- **test_values**: `php_extensions: []` added to existing scenarios, plus a new `php-extensions-values.yaml` scenario covering all three paths: `apcu` (pecl, preinstalled → skipped), `ds` (pecl, compiled), `calendar` (core, compiled).
- `cspell.config.yaml`: added `apcu`, `bcmath`.
- App version bumped to 2.3.46.

## Testing

- All five test scenarios render successfully with `ghcr.io/truenas/apps_validation:latest` and pass `docker compose config` validation.
- The generated Dockerfile layers were built for real against `nextcloud:34.0.1`: the build succeeds, `apcu` is skipped, `ds` and `calendar` compile, and all three extensions are loaded in the resulting image.
- The injection guard was verified: a name like `apcu; rm -rf /` fails the render with `contains an invalid character [;]`.

### DIFF
--- a/cspell.config.yaml
+++ b/cspell.config.yaml
@@ -11,6 +11,7 @@ words:
   - altran
   - anki
   - anythingllm
+  - apcu
   - apikey
   - arcadiatechnology
   - archisteamfarm
@@ -30,6 +31,7 @@ words:
   - bambuddy
   - baserow
   - bazarr
+  - bcmath
   - bearbinary
   - bentopdf
   - beszel
--- a/ix-dev/stable/nextcloud/app.yaml
+++ b/ix-dev/stable/nextcloud/app.yaml
@@ -82,4 +82,4 @@ sources:
 - https://github.com/nextcloud/docker
 title: Nextcloud
 train: stable
-version: 2.3.45
+version: 2.3.46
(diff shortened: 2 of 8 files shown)


stavros-k:
Sorry, but there are no plans to allow installing random things.
It is not testable/reproducible on our our side.
If you want to **customize** it, you can use the custom app.

PR #5308, opened 2026-07-04
feat(nextcloud): add optional Client Push (notify_push) support

cedricziel:
## Summary

Adds an opt-in **Client Push (notify_push)** integration to the Nextcloud app. notify_push replaces the constant polling of desktop/mobile clients with server push notifications, significantly reducing server load — one of the most requested Nextcloud performance features. Enabling the toggle gives a fully working push setup with no manual `occ` commands or sidecar containers.

## Design

- **`notifypush` container** — reuses the existing `nextcloud` image (unbuilt) and runs the push daemon binary that ships **inside the notify_push app** on the shared html volume. This guarantees the daemon version always matches the installed app version (the same approach Nextcloud AIO uses). The entrypoint waits until the app is installed, then `exec`s the binary.
- **Automatic app setup** — a `before-starting` hook on the nextcloud container runs `occ app:enable notify_push || occ app:install notify_push` and sets `base_endpoint` to `http(s)://<host>/push` (scheme follows the certificate setting). All failures are non-fatal so Nextcloud always starts even if the app store is unreachable.
- **`/push` route via Apache** — `proxy`, `proxy_http` and `proxy_wstunnel` are enabled in the image build (only when the feature is on) and a conf proxies `/push/ws` (websocket) and `/push/` to the daemon. This works identically in plain-HTTP mode and behind the app's nginx/HTTPS container, with no additional published ports.
- **Least privilege** — the daemon runs as `568:568` (the release binaries are world-executable), is configured purely via environment variables (`DATABASE_URL`/`REDIS_URL` with URL-encoded credentials, `NEXTCLOUD_URL`, `PORT`), and needs no added capabilities.
- Healthcheck is a TCP-accept probe on port 7867 — the daemon has no dedicated health endpoint (`/test/cookie` returns 400 unless a test cookie is staged).
- Requires the `Host` field to be set (validated at render time), since clients must be able to reach the push endpoint.

## Testing

- All six test scenarios render with `ghcr.io/truenas/apps_validation:latest` and pass `docker compose config`; new `notify-push-values.yaml` scenario included.
- **Full local smoke deployment** of the rendered compose: the hook installed notify_push 1.3.3 from the app store, the daemon started as uid 568 and connected to Redis (`occ notify_push:self-test`: redis ✓), `/push` was reachable through Apache (websocket module path verified), and the container reported **healthy**. The self-test's remaining warnings (plain HTTP, localhost URL) are artifacts of the test values, not the implementation.
- The smoke test caught and fixed a healthcheck bug (`/test/cookie` → TCP probe) before CI.
- Container enums in the labels/networks sections, `run_as_context`, and the version bump are included.

### DIFF
--- a/cspell.config.yaml
+++ b/cspell.config.yaml
@@ -330,6 +330,7 @@ words:
   - nocopy
   - noip
   - notifiarr
+  - notifypush
   - nowdns
   - npmplus
   - ntfy
@@ -569,6 +570,7 @@ words:
   - wiki-js
   - wizarr
   - woodpeckerci
+  - wstunnel
   - wtfismyip
   - wyze
   - xattr
--- a/ix-dev/stable/nextcloud/app.yaml
+++ b/ix-dev/stable/nextcloud/app.yaml
@@ -63,6 +63,11 @@ run_as_context:
   group_name: Host group is [root]
   uid: 0
   user_name: Host user is [root]
+- description: Container [notifypush] can run as any non-root user and group.
+  gid: 568
+  group_name: Host group is [apps]
+  uid: 568
+  user_name: Host user is [apps]
 - description: Container [postgres] runs as non-root user and group.
   gid: 999
   group_name: Host group is [docker]
@@ -82,4 +87,4 @@ sources:
 - https://github.com/nextcloud/docker
 title: Nextcloud
 train: stable
-version: 2.3.45
+version: 2.3.46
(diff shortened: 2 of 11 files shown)


stavros-k:
Notify push (among other things) exists in the nextcloud app in the test train.
I'll probably move it this week to the community train.
Which has many advantages (https://forums.truenas.com/t/nextcloud-test-train-feedback-request/37296/19).

For the stable one there are no plans to add other integrations at the moment

PR #5320, opened 2026-07-07
Fix repo link for Actual Budget

infinitytec:
Repo for Actual Budget moved to https://github.com/actualbudget/actual some time ago so the link given was to an archived repository.


stavros-k:
https://github.com/truenas/apps/blob/0742eeb7c54cf28115971efac12395a5320fb5b2/CONTRIBUTIONS.md?plain=1#L139

PR #5322, opened 2026-07-07
bambuddy: add Tailscale integration support (socket mount, host netwo…

### LINKED ISSUE #5321: [Enhancement]: bambuddy: fix Tailscale integration — privileged port bind failures and missing socket
### App Name

bambuddy

1.0.1

### Enhancement Description

When a Tailscale IP is configured for a virtual printer in Bambuddy, two issues occur:

  1. Binding to ports 990 (FTPS) and 322 (RTSP) on the Tailscale IP fails with errno 13 EACCES. The Tailscale interface (tailscale0) only exists in the host network namespace, so the container cannot reach it without host_network: true.
  2. tailscaled.sock is not accessible inside the container, producing a warning that the socket is not mounted. On TrueNAS, Tailscale runs as an app rather than a host OS service, so the socket path differs from the standard /var/run/tailscale/tailscaled.sock.

I'm working on this

PR to follow.

### Use Case

Users running Bambu Studio on a remote machine (e.g. a laptop away from home) can connect to Bambuddy's virtual printer over their Tailscale network instead of requiring direct LAN access.

### Proposed Solution

  - Add a Tailscale Configuration section to questions.yaml with an optional socket_path field. When set, the socket is bind-mounted into the container at /var/run/tailscale/tailscaled.sock.
  - Users who need Tailscale IP binding should enable Host Network (already available in Network Configuration).
  - Bump version to 1.0.2 and add a migration for existing installs.

### Alternatives Considered

_No response_

### Additional Context

_No response_

hahny93:
Closes #5321

### DIFF
--- a/ix-dev/community/bambuddy/app.yaml
+++ b/ix-dev/community/bambuddy/app.yaml
@@ -33,4 +33,4 @@ sources:
 - https://github.com/maziggy/bambuddy
 title: Bambuddy
 train: community
-version: 1.0.1
+version: 1.0.2
--- a/ix-dev/community/bambuddy/app_migrations.yaml
+++ b/ix-dev/community/bambuddy/app_migrations.yaml
@@ -1,0 +1,6 @@
+migrations:
+- file: add_tailscale_config
+  from:
+    max_version: 1.0.1
+  target:
+    min_version: 1.0.2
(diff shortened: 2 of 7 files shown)


stavros-k:
you can already do that using additional storage.

PR #5342, opened 2026-07-09
seafile: add X-Forwarded-Proto header when TLS is terminated

mreid-tt:
## Description

Adds `X-Forwarded-Proto $scheme` to all three nginx proxy locations (`/`, `/socket.io`, `/sdoc-server/`), rendered only when a TrueNAS certificate is configured (`certificate_id` is set). This informs the backend (Django/seahub and the sdoc-server) that requests arrived over HTTPS, which is required for correct CSRF validation when accessing Seafile directly via HTTPS.

This does not address the external reverse proxy case — users behind a reverse proxy who encounter CSRF 403 errors will need to manually add `CSRF_TRUSTED_ORIGINS` and `SECURE_PROXY_SSL_HEADER` to their instance's `seahub_settings.py`.


mreid-tt:
Hey @stavros-k, now that the app has been published, I was able to test it with real SSL certificates and behind a reverse proxy. That uncovered a few issues that weren't reproducible during the initial testing. I've proposed a fix here.

ix-dev/community/seafile/templates/docker-compose.yaml
```diff
@@ -77,6 +77,17 @@
 
 {% do seafile.add_caps(["CHOWN", "DAC_OVERRIDE", "FOWNER", "SETGID", "SETUID"]) %}
 {% do seafile.add_extra_host(seafile_host_no_port, "host-gateway") %}
+{% set seafile_csrf_origin = values.seafile.server_url.rstrip("/") %}
+{%- set seafile_settings_script -%}
+#!/bin/bash
+TARGET="/shared/seafile/conf/seahub_settings.py"
+if [ -f "$TARGET" ]; then
+  sed -i "/^CSRF_TRUSTED_ORIGINS/d;/^SECURE_PROXY_SSL_HEADER/d" "$TARGET"
```
stavros-k: sed is going to cause all sorts of issues. even produce a garbage file

ix-dev/community/seafile/templates/macros/nginx.conf
```diff
@@ -17,6 +17,9 @@ server {
         proxy_set_header Host $http_host;
         proxy_set_header X-Real-IP $remote_addr;
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+        {% if values.network.certificate_id %}
+        proxy_set_header X-Forwarded-Proto $scheme;
+        {% endif %}
```
stavros-k, line 22: isnt this required for the other locations?

mreid-tt: did not seem to make a difference but it couldn't hurt so it has been added

ix-dev/community/seafile/templates/docker-compose.yaml
mreid-tt: i'm going to try a different approach, hope this is better. if not I'd really love to hear your suggestion on how best accomplish this

ix-dev/community/seafile/templates/docker-compose.yaml
```diff
@@ -77,6 +77,19 @@
 
 {% do seafile.add_caps(["CHOWN", "DAC_OVERRIDE", "FOWNER", "SETGID", "SETUID"]) %}
 {% do seafile.add_extra_host(seafile_host_no_port, "host-gateway") %}
+{% set seafile_csrf_origin = values.seafile.server_url.rstrip("/") %}
+{%- set seafile_settings_script -%}
+#!/bin/bash
+TARGET="/shared/seafile/conf/seahub_settings.py"
```
stavros-k: I don't like this. This file is meant to be edited by users, which can contain duplicate variables, variables can even be inline lambas, function calls, arrays and whatnot. This naive find/replace will inevitably break it.

If seafile does not have a way to define them via envs or something more structured, we cannot do this thing.

mreid-tt: Following up on your feedback — I looked at how the upstream seafile-mc image's `bootstrap.py` handles settings. It appends `TIME_ZONE` to `seahub_settings.py` during initial setup using `a+` mode (no find/replace, just `>>` to the end of the file):

```python
with open(join(topdir, 'conf', 'seahub_settings.py'), 'a+') as fp:
    fp.write("\nTIME_ZONE = '{time_zone}'".format(
        time_zone=os.getenv('TIME_ZONE', default='Etc/UTC')))
    fp.write('\n')
```

I've updated the startup script to follow the exact same pattern — `grep -q` to check if the setting exists, then `printf >>` to append if absent. No find/replace, no sed, no temp files. This avoids the concerns about breaking Python syntax while matching how the upstream image itself handles configuration.

stavros-k: This is done only at init time. This does not work for us. Changing the url will not update it. Leaving it broken again.

mreid-tt: I'm not sure where to go from here. At this point, it seems like we have two options:

1. Go back to the `grep`/`printf` approach to rewrite the file, ensuring there's only a single instance of each line we need to add.
2. Drop the startup script entirely and accept that reverse proxy users need to manually add `CSRF_TRUSTED_ORIGINS` to `seahub_settings.py` — the same way they'd configure any other advanced Django setting.

I'm out of ideas at this stage and would really appreciate your guidance on the best way forward.

stavros-k: Given than Reverse proxy is not configured within _our_ app, it should not be handled by it.

stavros-k (approved):

PR #5408, opened 2026-07-18
Add basic Authntication support for the Hermes-Agent Dashboard

### LINKED ISSUE #5407: [Enhancement]: Hermes-Agent Authentication adjustment
### App Name

Hermes-Agent

1.0.6

### Enhancement Description

Hey,

since the new version of Hermes-Agent it is no longer allowed to run the Dashboard insecure.
Since the current version it is mandatory to have at least one authentication method, for example the new username + password method, if the admin do not want to configure OAuth or OIDC.
It is not possible to start the app, without configuring at least one of them.

Here are the official Docs: https://hermes-agent.nousresearch.com/docs/user-guide/docker/#running-the-dashboard

Kind regards

### Use Case

Authntication

### Proposed Solution

New Env-Variables and remove deprecated one.

### Alternatives Considered

_No response_

### Additional Context

_No response_

HubGon:
Hi,

this PR adds basic Authentication for the Hermes Dashboard. Since the latest update it is mandatory to have at least one authentication method.
Also removed the option to enable insecure dashboard, since it is deprecated.

Kind regards
Closes #5407


ix-dev/community/hermes-agent/questions.yaml
```diff
@@ -17,6 +17,51 @@ questions:
...
+            required: true
+            private: true
+        - variable: hmac_key
+          label: HMAC App Key
+          description: |
+            Signing the basic provider's stateless<br>
+            session tokens, so login stays valid after<br>
+            a restart.</br>
+            Must be 32+ bytes, base64/hex/raw encoded.<br>
+            Generate a strong secret with `openssl rand -hex 32`.<br>
```
yuk7hi, line 49: Since this is rendered as HTML, let's wrap the code segment using `<code>` tags.
```suggestion
            Generate a strong secret with <code>openssl rand -hex 32</code>.
```

yuk7hi: There's a similar line at line:70 (added by me), could you please update this line as well? TIA

yuk7hi:
The related test values used in CI needs to be updated as well, to omit the deprecated config and include the auth configs.
https://github.com/truenas/apps/blob/f733713ecfda1d683043775e6d9cc8f09545e1b3/ix-dev/community/hermes-agent/templates/test_values/basic-values.yaml#L6-L9

ix-dev/community/hermes-agent/questions.yaml
```diff
@@ -38,16 +83,6 @@ questions:
           schema:
             type: boolean
             default: false
-        - variable: enable_insecure_dashboard_access
```
stavros-k, line 41: This change breaks all installations. It requires migration.

ix-dev/community/hermes-agent/questions.yaml
```diff
@@ -17,6 +17,51 @@ questions:
     schema:
       type: dict
       attrs:
+        - variable: basic_auth_enabled
```
stavros-k: Should be an enum, so it can easily be expanded if required.

ix-dev/community/hermes-agent/questions.yaml
HubGon: The dashboard of the app broke for all user since the latest update.
But we can do a migration if you want to. What do you think about enabling the dashboard by default and set standard username + password for everyone who checked the old setting "enable_insecure_dashboard_access"?

stavros-k: Removing something from schema will error out when you click upgrade. It will require reinstall.

What would a user see if there is no auth method configured?

HubGon: Ok, good to know.

The Web-Dashboard is not showing up at all and in the log the admin is hinted, that the old env variable for insecure access is deprecated and there should be at least one configured authentication methode.

It does not have to be basic auth, it can also be OIDC or OAuth.

stavros-k: Ok lets default to `""` in the auth type enum for existing installations (ie in the migration) but on the questions, lets default to basic.
Whoever has already upgraded and its on a "broken" webui, has probably already used additional envs to fix it, so if we migrate it to basic, it will error out for duplicate variables.

We can check tho if basic auth variables are defined and migrate them to the fields too.

ix-dev/community/hermes-agent/questions.yaml
HubGon: I don't get why this should be an enum. The user can choose to use the basic authentication. He is also allowed to choose to configure OIDC or OAuth to get access to the Dashboard.

stavros-k: You can have both basic AND oidc? or is it one at a time/?

ix-dev/community/hermes-agent/migrations/auth_migration
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/python3
+
+import os
+import sys
+import yaml
+
+def migrate(values):
+    values["hermes-agent"].pop("enable_insecure_dashboard_access", None)
+
+    if values["hermes-agent"].get("additional_envs", {}).get("HERMES_DASHBOARD_BASIC_AUTH_USERNAME") and values["hermes-agent"].get("additional_envs", {}).get("HERMES_DASHBOARD_BASIC_AUTH_PASSWORD"):
```
stavros-k: additional envs is a list. You can't just pop keys

stavros-k: minecraft should have a migration for additional envs to check

HubGon: rewrote the migration-script, to be more like the minecraft one. Thanks for the hint, didn't found one yesterday by myself...
hope this time its ok :)

ix-dev/community/hermes-agent/questions.yaml
yuk7hi: You can setup multiple authentication flows at the same time (i.e. Basic + OAuth + OIDC).

stavros-k (approved):

PR #5410, opened 2026-07-19
Add Tailscale Services configuration

sbquinlan:
## Summary

- add optional Tailscale Services and backend endpoint settings to the existing Tailscale app
- generate a native `TS_SERVE_CONFIG` JSON file while keeping the stock image and a single Tailscale container
- default Services to Tailscale-managed HTTPS on port 443 while allowing HTTP, HTTPS, or raw TCP endpoints on custom ports
- add a Tailnet DNS Suffix setting for HTTP and HTTPS Service hostnames; raw TCP endpoints do not require it
- validate host networking, target hosts, backend protocols, and duplicate Service or port definitions

Related to #3937.

## AI

- [x] Part or all of this PR was generated with LLM assistance.

## Validation

- rendered HTTPS, HTTP, raw TCP, IPv6, empty-managed, disabled, and missing-DNS-suffix cases
- verified the generated HTTPS configuration maps a Service hostname on port 443 to a different local backend port
- confirmed the rendered Compose contains only the stock Tailscale container
- passed YAML parsing, `git diff --check`, and `docker compose config` for the installable test Compose
- inspected the Tailscale v1.98.9 Serve configuration and Service hostname lookup paths

### DIFF
--- a/ix-dev/community/tailscale/README.md
+++ b/ix-dev/community/tailscale/README.md
@@ -3,3 +3,6 @@
 [Tailscale](https://tailscale.com) Secure remote access to shared resources
 
 - When `Userspace` is **disabled**, `Tailscale` will run with `/dev/net/tun` device mounted from the host.
+- Tailscale Services can optionally forward TCP ports to applications published on the NAS. Define each Service
+  and its TCP ports in the Tailscale admin console first, authenticate this node with a tag-based identity, and
+  approve the Service advertisement or configure an auto-approver policy.
--- a/ix-dev/community/tailscale/README.md
+++ b/ix-dev/community/tailscale/README.md
@@ -5,6 +5,7 @@
 - When `Userspace` is **disabled**, `Tailscale` will run with `/dev/net/tun` device mounted from the host.
 - Tailscale Services can optionally expose applications published on the NAS. HTTPS endpoints terminate
   Tailscale-managed TLS and proxy to the application's backend port, so a Service on port 443 can reach an app
-  published on a different port. Define each Service and its ports in the Tailscale admin console first, enable
-  HTTPS for the tailnet when using HTTPS endpoints, authenticate this node with a tag-based identity, and approve
-  the Service advertisement or configure an auto-approver policy.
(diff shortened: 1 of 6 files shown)


stavros-k:
I don't understand why we have to move the json generation here? This is something application specific and can change at any point, or add new features.
Having it here, prohibits ANY change to that file from the user. So as tailscale adds more features we are forced to add support for those. This is not something that would need to be maintained here.

From what I can see it is just a JSON + env var, which can already be done as explained in the linked issue.

Thanks

sbquinlan:
@stavros-k, I don’t think pointing to Additional Storage and Additional Environments addresses the purpose of this PR:

- Transmission already translates wizard settings into `settings.json`, and several other apps generate JSON, YAML, nginx, Mosquitto, rsync, or JavaScript configuration from wizard values.
- Managed Services is disabled by default. Users can still supply arbitrary JSON through Additional Storage and `TS_SERVE_CONFIG`; the two approaches are not mutually exclusive.

“Users can manually write and mount the configuration” could equally be used to argue against providing most app wizard options—they could just write Compose YAML themselves. Technical possibility is not the same as an accessible or convenient user experience. This PR adds a wizard for the common case without removing the advanced escape hatch.

Would you reconsider and reopen the PR?

stavros-k:
All those apps are __patching__ the config files AND only things that are deployment varying, ports, db creds, stuff like that. Maybe one or two extra settings.
Those above are mostly mandatory to do, because most values are cross cutting, ie both on docker compose definition (which user cannot edit) and on the app, so this is done to keep them in sync. 

rsync, webdav, ddns-updater are more special apps. Those replace previously host baked services.

What this PR does, is brings all the burden of keeping the wizard aware of ALL the tailscale options. Cause you are taking ownership of the file.
And given that the file is not really a deployment concern, ie does not link to any docker compose details like ports, or other dependencies that the user does not have option to configure, what I said still stands.

Otherwise we will have to start managing every config file under the sun.
Homepage's yaml files, meshcentral's json file, random app ini file, conf file, another custom format file.

Thanks

sbquinlan:
alright, fair enough. thanks

PR #5423, opened 2026-07-20
Fix PlainNVR recording host path permissions

endless1233214:
## Summary
- bump PlainNVR to app image 0.1.1 / catalog version 1.0.1
- enable automatic permissions for PlainNVR data and recordings host paths
- keeps host-path installs writable by the configured non-root PlainNVR UID/GID

## Why
A real TrueNAS install using host-path recordings could start live view successfully while recording stayed in recovery because camera segment directories were not writable by UID/GID 568. The permission helper skipped host paths unless `auto_permissions` was set, so this exposes the checkbox and defaults it on for PlainNVR's writable storage mounts.

## Testing
- Ruby YAML parser loaded app.yaml, ix_values.yaml, and questions.yaml
- git diff --check
- Verified on a running TrueNAS install that fixing directory ownership restored segment writes

### DIFF
--- a/ix-dev/community/plainnvr/questions.yaml
+++ b/ix-dev/community/plainnvr/questions.yaml
@@ -425,6 +425,17 @@ questions:
                         type: hostpath
                         show_if: [["acl_enable", "=", false]]
                         required: true
+                    - variable: auto_permissions
+                      label: Automatic Permissions
+                      description: |
+                        Automatically set permissions for the host path.
+                        PlainNVR runs as the configured User ID and Group ID, so
+                        enabling this lets TrueNAS correct ownership when the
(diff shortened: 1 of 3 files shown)


ix-dev/community/plainnvr/questions.yaml
```diff
@@ -425,6 +425,17 @@ questions:
                         type: hostpath
                         show_if: [["acl_enable", "=", false]]
                         required: true
+                    - variable: auto_permissions
```
stavros-k, line 428: Thats only for postgres datasets.

The permissions apply only on non-empty directories and never touched again.

stavros-k:
Version is bumped in #5425

PR #5439, opened 2026-07-22
Configure Renovate versioning for AE NetScope

WhiteAssassins:
## Description

Adds custom versioning for AE NetScope so Renovate can detect its versioned alpha image tags correctly.

## Changes

- Adds a `customVersioning()` rule scoped to `ghcr.io/whiteassassins/ae-netscope`.
- Supports `vX.Y.Z-alpha` tags and optional numeric alpha hotfix suffixes such as `vX.Y.Z-alpha.1`.
- Replaces the previous manual app update approach; no app files are changed.

## Validation

- Validated `.github/renovate-config.js` with the official Renovate config validator.
- Confirmed the regex accepts versioned alpha and alpha hotfix tags.
- Confirmed it rejects floating `alpha`, tags without the `v` prefix, stable tags, and other prerelease suffixes.

### DIFF
--- a/.github/renovate-config.js
+++ b/.github/renovate-config.js
@@ -73,6 +73,13 @@ module.exports = {
       labels: ["patch"],
     },
     {
+      // AE NetScope publishes stable, alpha, and alpha hotfix tags.
+      matchDatasources: ["docker"],
+      matchPackageNames: ["ghcr.io/whiteassassins/ae-netscope"],
+      versioning: "semver",
+      allowedVersions: "/^v\\d+\\.\\d+\\.\\d+(?:-alpha(?:\\.\\d+)?)?$/",
+    },
+    {
       matchDatasources: ["docker"],
       labels: ["enterprise"],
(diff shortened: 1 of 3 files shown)


stavros-k:
Please fix the renovate-config **instead**.

.github/renovate-config.js
```diff
@@ -72,6 +72,13 @@ module.exports = {
       groupName: "updates-patch-minor",
       labels: ["patch"],
     },
+    {
+      // AE NetScope publishes stable, alpha, and alpha hotfix tags.
```
stavros-k: There is like 200 lines of examples for custom versioning, why did you made up your own... ?

WhiteAssassins: i used SemVer because AE NetScope has alpha, alpha.N, and stable tags. But yeah, I should use the existing customVersioning() pattern

stavros-k (approved):

PR #5477, opened 2026-07-29
Add YunKan to community train

### LINKED ISSUE #5476: [App Request]: YunKan (self-hosted NVR with on-device AI)
### App Name

YunKan

### App Description

Self-hosted NVR for RTSP/ONVIF IP cameras with on-device AI detection (motion, person, vehicle, face, license plate, pose/fall, package, baby cry). Everything runs in a single container - media server, recorder, detection workers, web admin and reverse proxy - and video never leaves the local network.

### App Website / Github Repo

https://yun-kan.com/

### Docker Installation Documentation

https://yun-kan.com/docs

### App Icon URL

https://raw.githubusercontent.com/mrtian2016/yunkan-unraid-templates/main/icon.png

### Docker Image

ghcr.io/mrtian2016/yunkan-cpu (variants: yunkan-openvino / yunkan-cuda / yunkan-trt)

### App Category

Security

### Requirements Checklist

- [x] The app is actively maintained
- [x] The app has official Docker images available
- [x] The app has versioned image tags available
- [x] The app is open source or has a free tier
- [x] I have tested the Docker installation locally

### Additional Information

I am the developer of YunKan and I have the app definition for the community train ready (modeled on the Frigate app: single app with an `image_selector` for the CPU / OpenVINO / CUDA / TensorRT images). Both test value files render and the CPU image deploys healthy with the repo's `ci.py`. A PR follows this issue.

Three points I would like to flag up front:

1. **Licensing model** - the app is proprietary with a free tier that works without any payment or account. Comparable apps already in the catalog are UniFi Controller, Tailscale and Plex.
2. **Host identity mounts** - license binding needs a stable hardware fingerprint, so the template mounts `/etc/machine-id` and `/sys/class/dmi/id/product_uuid` read-only, both declared in `host_mounts`. Nothing is written and nothing leaves the machine.
3. **Host network default** - the app defaults to host networking (WebRTC media, LAN camera discovery, RTSP). Bridge mode is fully supported via the usual port questions. Happy to flip the default if you prefer.

Images are published for linux/amd64. Similar app in the catalog: Frigate.

mrtian2016:
# App Addition

- [x] I have opened an [issue](https://github.com/truenas/apps/issues/5476) to discuss this app addition before submitting this pull request.

# AI

## Description

Adds YunKan to the community train.

YunKan is a self-hosted NVR for RTSP/ONVIF IP cameras with on-device AI detection (motion, person, vehicle, face, license plate, pose/fall, package, baby cry). Everything runs in a single container - media server, recorder, detection workers, web admin and reverse proxy, supervised by s6-overlay - and video never leaves the local network. I am the developer.

The app definition is modeled on Frigate: one app, with an `image_selector` for the CPU / OpenVINO / CUDA / TensorRT images.

## App Information

- **Upstream**: https://yun-kan.com/
- **Documentation**: https://yun-kan.com/docs
- **App Version**: 0.9.32

## Testing

Tested locally with:

- [x] basic-values.yaml - bridge network, ix_volume storage, CPU image; deployed with `ci.py`, container reaches healthy (`/healthz` 200)
- [x] host-network-values.yaml - host network path; rendered and deployed with the CPU image

## Icons and Screenshots

- Icon (512x512 PNG): https://raw.githubusercontent.com/mrtian2016/yunkan-unraid-templates/main/icon.png
- Screenshot 1 (1600x900): https://raw.githubusercontent.com/mrtian2016/yunkan-unraid-templates/main/screenshots/dashboard.jpg
- Screenshot 2 (1600x900): https://raw.githubusercontent.com/mrtian2016/yunkan-unraid-templates/main/screenshots/events.jpg

`app.yaml` currently points at the expected `media.sys.truenas.net` paths; I will update it if the final CDN URLs differ.

## Special Notes

- **Licensing**: proprietary with a free tier that works without any payment or account. Comparable apps already in the catalog: UniFi Controller, Tailscale, Plex.
- **Host identity mounts** (declared in `host_mounts`): `/etc/machine-id` and `/sys/class/dmi/id/product_uuid`, both read-only. They are only used to compute a stable hardware fingerprint locally for offline license binding - nothing is written and nothing leaves the machine.
- **Host Network defaults to enabled**: WebRTC media (23515 TCP/UDP), LAN camera discovery and RTSP need direct host networking to work out of the box. Bridge mode is fully supported (web + WebRTC ports published, `WebRTC External IPs` question for ICE candidates) - happy to flip the default if you prefer.
- **Runs as root**: the image is supervised by s6-overlay which starts the bundled media server, web server and worker processes; capabilities are limited to the six listed in `app.yaml`.
- **Docker socket is deliberately not mounted**: the app has an in-app self-upgrade feature on other platforms; without the socket it detects this and hides that path, so TrueNAS stays the only thing that recreates the container (no image-tag drift).
- **GPU variants**: the OpenVINO / CUDA / TensorRT images cannot be exercised in CI (no GPU on runners), so both test files use the CPU image. GPU passthrough goes through the standard `Resources -> GPU Configuration` section.
- Images are published for **linux/amd64** only.
- First start opens a setup wizard (SQLite or MySQL, then admin account creation) - covered in the app notes.

## Checklist

- [x] README.md included


ix-dev/community/yunkan/templates/test_values/host-network-values.yaml
```diff
@@ -0,0 +1,35 @@
...
+    cpus: 2.0
+    memory: 4096
+
+yunkan:
+  image_selector: image
+  webrtc_extern_ips: "192.0.2.10"
+  mount_product_uuid: false
+  devices: []
+  additional_envs:
+    - name: SKYVIEW_PUBLIC_PORT
```
stavros-k: I cant find this env in the docs

ix-dev/community/yunkan/ix_values.yaml
```diff
@@ -0,0 +1,53 @@
...
+consts:
+  yunkan_container_name: yunkan
+  internal_web_port: 23406
+  internal_webrtc_port: 23515
+  machine_id_path: /etc/machine-id
+  product_uuid_path: /sys/class/dmi/id/product_uuid
+  notes_body: |
+    ## First Start
+
+    Open the Web UI and complete the setup wizard (pick SQLite or MySQL, then create
```
stavros-k: Why is this a runtime config and not via env vars?

ix-dev/community/yunkan/questions.yaml
```diff
@@ -0,0 +1,723 @@
...
+                                        description: Indicates in which order Compose connects the service's containers to its networks.
+                                        schema:
+                                          type: int
+                                          "null": true
+        - variable: host_network
+          label: Host Network
+          description: |
+            Bind to the host network.</br>
+            Enabled by default: WebRTC live view and camera discovery on the LAN need
+            direct access to the host network. Clients use 23406 (web interface) and
```
stavros-k, line 336: Are the internal ports configured? 
Having host network option without configurable ports can cause port conflicts.

ix-dev/community/yunkan/questions.yaml
```diff
@@ -0,0 +1,723 @@
...
+                schema:
+                  type: int
+                  default: 4
+                  required: true
+              - variable: memory
+                label: Memory (in MB)
+                description: Memory limit for YunKan.
+                schema:
+                  type: int
+                  default: 8192
```
stavros-k: Dont deviate from defaults

ix-dev/community/yunkan/README.md
stavros-k, line 1: readme files are on purpose short and link upstream.
There is no reason to maintain copies of upstream docs here

ix-dev/community/yunkan/ix_values.yaml
```diff
@@ -0,0 +1,53 @@
...
+    repository: ghcr.io/mrtian2016/yunkan-trt
+    tag: 0.9.32
+
+consts:
+  yunkan_container_name: yunkan
+  internal_web_port: 23406
+  internal_webrtc_port: 23515
+  machine_id_path: /etc/machine-id
+  product_uuid_path: /sys/class/dmi/id/product_uuid
+  notes_body: |
```
stavros-k: I dont think those notes needs to be here.
Isn't all this shown once you open the webui anyway?

Also user does not care where it is mounted inside the container (it is still listed in the volumes section in the UI), same for ports

stavros-k (changes requested):
Hello, did a quick review, nothing deep here.

ix-dev/community/yunkan/templates/test_values/host-network-values.yaml
mrtian2016: Removed - it was only there to exercise `additional_envs` in CI. The test file now uses `additional_envs: []`.

ix-dev/community/yunkan/ix_values.yaml
mrtian2016: That's how the upstream app works: a first-run wizard in the web UI writes the app's own config store, so the database choice and the admin account can't be provisioned via env vars. The notes section describing it has been trimmed away anyway.

ix-dev/community/yunkan/questions.yaml
mrtian2016: The internal ports are fixed in the image (the web proxy, media server and stream-token auth are wired to them) and were deliberately picked in an uncommon high range (23406, 23515, 23880, 24214) to avoid clashing with common services. Users who do hit a conflict can turn off host networking and use the bridge config, where the host-side ports are fully configurable. Host networking is the default because WebRTC live view and ONVIF camera discovery then work out of the box.

ix-dev/community/yunkan/questions.yaml
mrtian2016: Reverted to the library defaults (2 CPUs / 4096 MB).

ix-dev/community/yunkan/README.md
mrtian2016: Trimmed to the short form, linking upstream.

ix-dev/community/yunkan/ix_values.yaml
mrtian2016: Trimmed to two short notes (first-run wizard + pointing recordings at a dataset with enough capacity); dropped the mount paths and port details.

ix-dev/community/yunkan/questions.yaml
mrtian2016: To add some context on why it's the default: camera discovery is WS-Discovery (UDP multicast), and the app also advertises itself via mDNS (`.local` hostname for the mobile apps) and exposes cameras to HomeKit through zeroconf. All of these are multicast-based, so they only work with host networking.

ix-dev/community/yunkan/templates/test_values/host-network-values.yaml
stavros-k: If this is a valid env (or there are more valid envs) it might be good to have them somewhere documented/listed.

ix-dev/community/yunkan/ix_values.yaml
stavros-k: That sucks, most if not all apps here are "locked" to either sqlite or some db (with the db shipped as a side container). So what I'm testing is what user also gets.
And having it able for each user to just do random, makes it hard to maintain/debug.

ix-dev/community/yunkan/questions.yaml
stavros-k: Right, but host network and mDNS will cause issues on TrueNAS's mDNS.

ix-dev/community/yunkan/ix_values.yaml
stavros-k: Well that note for capacity is a bit "too late". Those notes are shown _after_ installation.

This section is mostly for things like default password, or some management cli command reminder (ie for nextcloud `occ blah`) etc.

ix-dev/community/yunkan/templates/test_values/host-network-values.yaml
mrtian2016: Agreed. The image intentionally keeps the env surface small - WebRTC extern IPs and the media secret (both already exposed as questions here) plus a couple of TLS/port overrides for reverse-proxy setups. I'll add a reference list for them to the upstream docs.

ix-dev/community/yunkan/ix_values.yaml
mrtian2016: Fair concern - to clarify how it actually behaves: the wizard silently provisions the bundled SQLite on its own as soon as it loads; the user only creates the admin account. MySQL is an advanced path for people who already operate their own DB server - the app never ships or manages a database and there is no side container in any deployment, so the container topology is identical either way. What CI deploys here is exactly what a default user gets (SQLite).

stavros-k: Well even if the setup stays the same, the user can change the "deployment" from the shipped one.

For example:

User configures some database, which might for example fail to connect, tries to use the app, it crashes somewhere and because of some failed operation it surfaces in the logs as a directory not found error for example. (ie a directory might be created after a succesful commit which never does on a not connected db).

User reports here that app has permissions problems (ie fails to create the dir). But thats not the real cause.
Sure weird scenario, not saying that this will happen, but I've seen similar things many times. (ie an error on X to surface as error of Y).

---

We prefer to "lock" down whatever it has to do with "infra", to avoid weird cases like these.

ix-dev/community/yunkan/questions.yaml
mrtian2016: The app's advertisement is a separate responder (python-zeroconf) under its own distinct name (`yun-kan.local`) - it never claims or answers for the NAS's own hostname, and name probing (RFC 6762) plus SO_REUSEPORT socket sharing is the same pattern Home Assistant uses alongside Avahi. That said, if you have seen concrete interference on SCALE, we can add an env switch in an upcoming release to turn the app's advertisement off and set it in this chart.

ix-dev/community/yunkan/ix_values.yaml
mrtian2016: We'd actually rather not lock this one down: the app is multi-process (recorder, detection and automation workers all share the database), and while the bundled SQLite is what everyone gets by default and is fine at typical home scale, heavier multi-camera deployments genuinely benefit from MySQL/MariaDB - that's why the advanced path exists. It is a deliberate escalation for people who already run a DB server, not free-form infra.

On the error-masking concern: the wizard probes the connection (and auto-creates the database) before the choice is committed - an unreachable or misconfigured database fails right there with an explicit error, so the app never ends up half-configured on a dead DB.

ix-dev/community/yunkan/ix_values.yaml
mrtian2016: Right - and the capacity hint was already in the [Recordings Storage] question description, where it's actually actionable, so I dropped it from the notes. They now only carry the first-start pointer (open the web UI, run the first-run wizard).

ix-dev/community/yunkan/ix_values.yaml
stavros-k: I'd prefer to ship the mariadb container in this case, but we need to be able to autoconfigure it.
Apps are meant to be easy to install and have their basic deps bundled. 
If someone has to bring his own db the moment that wants to add few cameras, then it defeats the purpose of the app.

mrtian2016: Agreed, that's a fair point - let's do it. Two parts:

1. App side: the next release adds env-based DB provisioning (`SKYVIEW_DB_TYPE` / `SKYVIEW_DB_HOST` / `SKYVIEW_DB_PORT` / `SKYVIEW_DB_USER` / `SKYVIEW_DB_PASSWORD` / `SKYVIEW_DB_NAME`). When set, they are the source of truth - the wizard skips the database step entirely and only creates the admin account, so the deployment is fully locked to what the chart ships.
2. Chart side: I'll wire the library's mariadb dep (side container, auto-generated credentials, `depends_on: service_healthy`, its own ix_volume) and pin the images to that release, so every TrueNAS install gets the bundled MariaDB out of the box.

Our release cadence is quick - I expect to update this PR within days. This also happens to match what we recommend upstream anyway: the app is multi-process and MariaDB holds up better than SQLite once camera count grows.

stavros-k: Sounds good to me

ix-dev/community/yunkan/templates/test_values/host-network-values.yaml
stavros-k: btw the extern IPs can be a list in the UI and you join them in the compose file

ix-dev/community/yunkan/questions.yaml
stavros-k: I won't say that I'm proficient with mDNS, so I can't speak about it.
I'll try to lookup older mdns discussions, maybe I have more notes there, and I'll come back to this

ix-dev/community/yunkan/ix_values.yaml
mrtian2016: Done - the PR now ships a bundled MariaDB (side container, own volume, `depends_on: service_healthy`) and the app is locked to it through `SKYVIEW_DB_*`, so nothing about the database is configurable by the user anymore.

Verified end to end against the released image with the repo's own ci.py (real deploy, not just render):
- both containers reach healthy, the app creates its 37 tables in the bundled MariaDB, and no database config file is written on the app's data volume;
- `/api/setup/status` reports `{phase: admin_missing, database: mysql}` - the wizard skips the database step entirely and only asks for the admin account;
- the app's own database endpoints refuse to change anything (403), and online migration in its settings is disabled.

One detail worth flagging: with host networking every container shares the host namespace, so the database binds 127.0.0.1 on an uncommon port rather than 3306, to avoid clashing with anything else on the NAS.

ix-dev/community/yunkan/templates/test_values/host-network-values.yaml
mrtian2016: Followed up on this too - the env vars the image accepts are now documented upstream (WebRTC extern IPs, the public web port, the single-port TLS overrides, the media secret, an mDNS off switch, and the SKYVIEW_DB_* set used here). They are all optional; a standard install needs none of them.

ix-dev/community/yunkan/questions.yaml
mrtian2016: Also delivered the mDNS off switch mentioned above: the image now honours `SKYVIEW_MDNS_ENABLED=0`, so if the app's advertisement ever conflicts with the NAS's own mDNS stack it can be turned off without touching anything else. Happy to set it in this chart if you'd prefer it off by default on SCALE.

ix-dev/community/yunkan/templates/test_values/host-network-values.yaml
mrtian2016: Here it is: https://yun-kan.com/en/docs/install#environment-variables

stavros-k:
Ugh, Right. When using host network docker DNS breaks. 
That's a problem. I don't want to bind such deps on the host, which is why the internal networks was introduced.

I've seen many times 127.0.0.1 bind break due to users running some VPN app with host network. Or having custom routes defined on the host.

mrtian2016:
You're right, and it's worse than that: a `network_mode: host` container can't join any docker network at all, so an internal network isn't reachable from the app either way.

So I dropped TCP entirely - the bundled MariaDB now runs with `--skip-networking` and the app talks to it over a unix socket on a shared `temporary` volume. No host binding, no docker DNS, unaffected by VPN apps or custom routes, and it works the same whether host networking is on or off.

Verified locally against the released image: MariaDB starts with `--network none`, the app connects over the socket and creates its full schema (37 tables), and the healthcheck runs `mariadb-admin --socket=...`. I'll push the chart change together with the app release that adds socket support - a couple of days.

stavros-k:
> You're right, and it's worse than that: a `network_mode: host` container can't join any docker network at all, so an internal network isn't reachable from the app either way.
> 
> So I dropped TCP entirely - the bundled MariaDB now runs with `--skip-networking` and the app talks to it over a unix socket on a shared `temporary` volume. No host binding, no docker DNS, unaffected by VPN apps or custom routes, and it works the same whether host networking is on or off.
> 
> Verified locally against the released image: MariaDB starts with `--network none`, the app connects over the socket and creates its full schema (37 tables), and the healthcheck runs `mariadb-admin --socket=...`. I'll push the chart change together with the app release that adds socket support - a couple of days.

Socket means that it will have to be a file in a shared _host_ path. A path that we cannot pick ourselves.

I'm not sure I want to go down that route. People have been building scripts and tools to automatically detect databases based on the _naming_ patterns and use the internal docker networks we create to join them and automatically backup.

mDNS is one of the reason that I regret going with postgres with home assistant. Makes is impossible to use host network AND having the setup ease.

This repo stricly prohibits host network when an app needs a dependency for this and many other reasons.

ix-dev/community/yunkan/ix_values.yaml
mrtian2016: Thought about this some more in light of the host-network policy. Since the dependency container and host networking can't be combined here, and host networking is what makes the camera side work out of the box (WS-Discovery multicast, WebRTC), here's where I landed:

- The bundled SQLite is back as the default - single container, zero setup, and it's what the CI deploys. It genuinely covers typical home scale.
- For the heavier setups, the database can now be declared at install time instead: an optional [External Database] section in the form (for example pointing at the MariaDB app from this catalog), provisioned through the same `SKYVIEW_DB_*` env vars. So when it's used it's still fixed infrastructure: the wizard skips the database step and nothing about the database is configurable inside the app.

Not as neat as shipping the dependency, I know - but within the constraints it keeps the install one-click for the common case and keeps the declared-up-front behaviour for the rest. Happy to adjust if you see a better shape for it.

ix-dev/community/yunkan/templates/test_values/host-network-values.yaml
mrtian2016: Good idea - done. [WebRTC External IPs] is now a list in the form, joined into the env var in the template.

ix-dev/community/yunkan/questions.yaml
```diff
@@ -49,26 +49,78 @@ questions:
                 description: CUDA (NVIDIA)
               - value: "tensorrt_image"
                 description: TensorRT (NVIDIA)
-        - variable: db_password
-          label: Database Password
+        - variable: external_db
+          label: External Database (Advanced)
```
stavros-k: Advanced use cases, will have to convert their app to custom-app and manage it from there.

So lets drop the config here, and lock to sqlite via env var.

mrtian2016: Agreed - done. The chart now just sets `SKYVIEW_DB_TYPE=sqlite` and the external database section is gone: single container, nothing about the database is configurable anywhere, and advanced setups can convert to a custom app as you said.

I'll push it together with the next app release - while testing the sqlite-locked mode I found and fixed a path-resolution issue in it, so the image pin needs to move one release forward. Within a day or so.

mrtian2016: Done and pushed. The chart now pins `SKYVIEW_DB_TYPE=sqlite`: the external database section is gone, the first-run wizard skips the database step (admin account only), and nothing about the database is configurable inside the app. Advanced setups can convert to a custom app as you said.

Also addressed from earlier in the thread: [WebRTC External IPs] is now a list in the form, joined in the template.

Images are pinned to 0.9.39 - while testing the sqlite-locked mode I found and fixed a path-resolution issue in it, so this release is the first one it works correctly on. Re-verified with a real ci.py deploy: single container reaches healthy, the database file lands on the data volume, and no database config file is written anywhere.

ix-dev/community/yunkan/ix_values.yaml
```diff
@@ -0,0 +1,20 @@
+images:
+  image:
+    repository: ghcr.io/mrtian2016/yunkan-cpu
+    tag: 0.9.39
```
stavros-k, line 4: Upstream site says that software updates are only during active subscription. What does that mean? If we push lets say a version 0.10.0 some users might end up on a setup that no longer works when their subscription is ended?

mrtian2016: Good question - that wording is about feature entitlement, not a kill switch. The app never checks the license against the software version: any release runs with any license state (or none at all).

When a subscription lapses, the app degrades in place to the free tier - recording, live view, playback, export and basic motion detection with push keep working (with a camera-count cap); only the paid AI features pause until renewal. That happens on whatever version they are already running, so pushing 0.10.0 changes nothing for an expired user: they end up in exactly the same state whether they update or not. Never a broken setup.

stavros-k (approved):

PR #5485, opened 2026-07-30
fix(nextcloud): include port in X-Forwarded-Host header for correct dual-access redirects

### LINKED ISSUE #2507: Nextcloud: Support IP address access with TLS without needing OVERWRITEHOST
I have the Nextcloud app installed with TLS enabled. It generally works, but in my environment, client devices must access the server using its IP address because they cannot resolve the domain name. This is common for older devices, such as legacy Android and Windows systems, which don’t support multicast DNS (e.g., *.local).

Because of this limitation, getting the Nextcloud app to work out-of-the-box wasn’t straightforward. To work around the issue, I modified the Nginx configuration template:
https://github.com/shimies/truenas-apps/commit/5670fe66ca6247fdb4ccbdc16470e5343422a7c0

I'm opening this issue to ask whether this change (or something similar) can be incorporated upstream. Currently, I need to manually reapply the patch after every update to the Nextcloud app.

#### Background
Looking at [Nextcloud app compose.yaml template](https://github.com/truenas/apps/blob/master/trains/stable/nextcloud/1.7.2/templates/docker-compose.yaml#L119), it seems that TLS support requires parameters of (`nextcloud.host` and `nginx.use_different_port`, `nginx.external_port`). These settings ensure that the `OVERWRITEHOST` is set (e.g., `${nextcloud.host}:${nginx.external_port}`), so Nextcloud knows which host/port to use in redirect URLs.
Without those parameters, nextcloud tells (through HTTP `Location` header) nginx to redirect to `https://${nextcloud.host}:443` which, in my case, is the port the Nextcloud web UI app is listening to not Nextcloud. (in my case `${nextcloud.host}` is set to `truenas.local`)

So, while using `OVERWRITEHOST` is fine for DNS-resolving clients, it breaks for clients that access Nextcloud via IP address. This led me to think that using `OVERWRITEHOST` might not be appropriate if we want to support both domain and IP access flexibly.

#### Workaround
I found that Nextcloud can generate the redirect URL that respects theoriginal request's host (IP address or domain name) if the nginx is configured appropriately.

Nextcloud works as follows:
1. [getServerHost()](https://github.com/nextcloud/server/blob/v30.0.5/lib/private/AppFramework/Http/Request.php#L810) is called to get hostname for redirect urls
1. which then call [getInsecureServerHost()](https://github.com/nextcloud/server/blob/v30.0.5/lib/private/AppFramework/Http/Request.php#L780) (if `OVERWRITEHOST` is not set) that gets it from `HTTP_X_FORWARDED_HOST` HTTP header.

Currently even though `X-Forwarded-Host` is set in the app, it does not include the port number, which is probably the problem and why it needs `OVERWRITEHOST` to support TLS in the first place.

---

Can we update the Nginx configuration so that it includes the port number in the `X-Forwarded-Host` header?

If so, we may no longer need to rely on `OVERWRITEHOST`, and clients could access Nextcloud using either the domain name or the IP address even with TLS enabled. The parameters (`use_different_port` and `external_port`) in the app setting page could be potentially removed as well.

### LINKED ISSUE #5484: Nextcloud: X-Forwarded-Host header missing port number (regression in 2.3.50)
## Problem
The nginx config in the Nextcloud app template sends `X-Forwarded-Host` using `$host` which does **not** include the port number:

```
proxy_set_header X-Forwarded-Host  $host;
```

Since the nginx container IP is in the `trusted_proxies` range (`172.16.0.0/12` for Docker bridge), Nextclouds `getInsecureServerHost()` prefers `X-Forwarded-Host` over the actual `Host` header. Without the port, redirect URLs are generated without it:

- `https://192.168.0.165:30027/` → `https://192.168.0.165/login` (port lost → hits TrueNAS instead of Nextcloud)

## Fix
Replace `$host` with `$http_host` (which includes the port):

```
proxy_set_header X-Forwarded-Host  $http_host;
```

After this change both access methods work correctly:
- LAN via IP: `https://192.168.0.165:30027/` → `https://192.168.0.165:30027/login` ✅
- External via FQDN: `https://nextcloud.domain.com/` → `https://nextcloud.domain.com/login` ✅

With this fix, `OVERWRITEHOST` is no longer needed for dual-access setups.

## Location
`ix-dev/stable/nextcloud/templates/macros/nc.jinja.conf` line 60

## Related
Closes #2507 (that issue identified the root cause but was closed without applying the `$host` → `$http_host` fix)

FerrumLogic:
## Problem
The nginx config sends `X-Forwarded-Host` using `$host` which does **not** include the port number:

```
proxy_set_header X-Forwarded-Host  $host;
```

When the nginx container IP is in `trusted_proxies` range (Docker bridge `172.16.0.0/12`), Nextcloud uses `X-Forwarded-Host` instead of the `Host` header. Without the port, all redirect URLs are generated without it, breaking LAN IP-based access.

**Result:** `https://192.168.0.165:30027/` → redirects to `https://192.168.0.165/login` (port 443/TrueNAS instead of Nextcloud on :30027).

## Fix
Changed `$host` to `$http_host` (which includes the port):

```
proxy_set_header X-Forwarded-Host  $http_host;
```

After this change both access methods work without `OVERWRITEHOST`:
| Access | Before | After |
|--------|--------|-------|
| LAN via IP | `https://192.168.0.165/login` ❌ | `https://192.168.0.165:30027/login` ✅ |
| External via FQDN | `https://nextcloud.domain.com/login` ✅ | `https://nextcloud.domain.com/login` ✅ |

## Files changed
- `ix-dev/stable/nextcloud/templates/macros/nc.jinja.conf`
- `trains/stable/nextcloud/2.3.50/templates/macros/nc.jinja.conf`

## Related issues
- Closes #2507 (original issue, same root cause but was closed without applying this fix)
- Closes #5484


FerrumLogic:
Hi @stavros-k! Since you are the active maintainer reviewing community PRs, could you take a look at this one?

This fixes the **regression introduced in 2.3.50** (reported in #5484): the nginx template uses `$host` for `X-Forwarded-Host`, which drops the port, so Nextcloud generates redirects without it when accessed via LAN IP (`https://192.168.0.165/login` instead of `https://192.168.0.165:30027/login`).

The fix is minimal — switch to `$http_host` in both the ix-dev and train templates. Both LAN (IP:port) and external (FQDN) access verified working without `OVERWRITEHOST`.

Thanks!

stavros-k:
Hello,

https://github.com/truenas/apps/blame/0096a655f691d9a9187322f4bc87a712aeb79e3c/CONTRIBUTIONS.md#L139

FerrumLogic:
Hi @stavros-k! Thank you for the pointer.

I have updated the PR: removed the manual edit to `trains/stable/nextcloud/2.3.50/templates/macros/nc.jinja.conf` (auto-generated, per CONTRIBUTIONS.md), and kept only the source fix in `ix-dev/stable/nextcloud/templates/macros/nc.jinja.conf` (`$host` -> `$http_host`).

The trains file will pick up the change automatically on the next catalog regeneration.

Thanks!

stavros-k (approved):

PR #5462, opened 2026-07-25
Update updates-patch-minor

truenasbot:
> ℹ️ **Note**
> 
> This PR body was truncated due to platform limits.

Closes #5461 

This PR contains the following updates:

| Package | Update | Change |
|---|---|---|
| actualbudget/actual-server | minor | `26.7.0` → `26.8.1` |
| alexta69/metube | minor | `2026.07.24` → `2026.08.04` |
| [amir20/dozzle](https://redirect.github.com/amir20/dozzle) | minor | `v10.6.11` → `v10.7.1` |
| [arcadiatechnology/crafty-4](https://craftycontrol.com/) ([source](https://gitlab.com/crafty-controller/crafty-4)) | patch | `4.10.7` → `4.10.8` |
| [automaticrippingmachine/automatic-ripping-machine](https://redirect.github.com/automatic-ripping-machine/automatic-ripping-machine) | patch | `2.24.1` → `2.24.3` |
| [binwiederhier/ntfy](https://ntfy.sh/) ([source](https://redirect.github.com/binwiederhier/ntfy)) | minor | `v2.26.3` → `v2.27.0` |
| chatwoot/chatwoot | patch | `v4.16.1-ce` → `v4.16.2-ce` |
| chatwoot/chatwoot | patch | `v4.16.1` → `v4.16.2` |
| chocobozzz/peertube | patch | `v8.2.3` → `v8.2.4` |
| [chrislusf/seaweedfs](https://redirect.github.com/seaweedfs/seaweedfs) | minor | `4.40` → `4.41` |
| [clamav/clamav](https://redirect.github.com/Cisco-Talos/clamav.git) ([source](https://redirect.github.com/Cisco-Talos/clamav)) | patch | `1.5.3` → `1.5.4` |
| [codeberg.org/forgejo/forgejo](https://forgejo.org) ([source](https://codeberg.org/forgejo/forgejo)) | patch | `16.0.1-rootless` → `16.0.2-rootless` |
| [codewithcj/sparkyfitness](https://redirect.github.com/CodeWithCJ/SparkyFitness) | patch | `v1.6.0` → `v1.6.1` |
| [codewithcj/sparkyfitness_server](https://redirect.github.com/CodeWithCJ/SparkyFitness) | patch | `v1.6.0` → `v1.6.1` |
[...]

PR #5493, opened 2026-08-01
ci: render environment-dependent app tests without installation

richardsolomou:
# Pull Request

## Description

Add per-scenario render-only support to the app test matrix for applications that require external credentials or unavailable hardware.

Use render-only validation for NetBird, Tailscale, stdapi.ai, and vLLM ROCm. Also correct the stdapi.ai region fixture to use its declared list shape and bump its catalog revision.

**Why:** These scenarios can validate rendering and Docker Compose configuration without predictably failing installation checks on generic CI runners.

## Testing

- Verified render-only and normal matrix classification
- Rendered the stdapi.ai fixture successfully
- Regenerated stdapi.ai metadata
- `uvx flake8 .github/scripts/changed_apps.py`
- `actionlint .github/workflows/app-test-suite.yaml`
- `git diff --check`
- All GitHub Actions checks pass

---
*Created with [PostHog Code](https://posthog.com/code?ref=pr)*


richardsolomou:
@stavros-k this should help unblock https://github.com/truenas/apps/pull/5462

stavros-k:
The failing tests do not block the PRs. All PRs require manual approval.

But I still dont want to skip install even for things that require external credentials.
Because logs can help catch an "invalid config" vs "invalid/missing credentials"

PR #5530, opened 2026-08-09
BookOrbit uploads fail on larger file sizes

pho3nixf1re:
# Issue
BookOrbit file uploads fail when the file size is > 20mb. This can also happen with multiple simultaneous smaller file uploads or when other processes are writing data to `/tmp`.

# Fix
This provides a configuration value that will allow a configurable tmpfs value used by the application. The default is a bit above the application's built-in file size limit to allow for some headroom on the `/tmp` path since that is used across processes and not just for the file upload.

I checked the available contribution templates but there only seems to be one for new apps, not changes to existing ones. I am glad to make any updates necessary.


stavros-k:
Removed limit all together to match upstream.
Added gid/uid for tmpfs, just to be on the "safe/explicit" side.

stavros-k (approved):

PR #5557, opened 2026-08-11
backrest: expose mount propagation for host path storage

### LINKED ISSUE #5556: [Enhancement]: Backrest: expose mount propagation (rslave) for host paths — replication remounts silently produce empty backups
### App Name

backrest

1.1.11

### Enhancement Description

Backrest's host-path storage items render their bind mounts as rprivate, and the app's questions.yaml exposes no way to change that. When a host path is a ZFS dataset tree whose children get remounted — most obviously a replication target — the running container's view of those children goes stale and never recovers until the app is restarted.

I'd like the host-path storage schema to expose mount propagation, so it can be set to rslave for paths that need to follow host mount changes.

The base library already types the field. In library/2.3.8/storage.py, IxStorageHostPathConfig declares:

`propagation: NotRequired[Literal["shared", "slave", "private", "rshared", "rslave", "rprivate"]]`
What's missing is the plumbing to reach it: Backrest's questions.yaml has no propagation field, and app config validation is pydantic with extra="forbid", so supplying the key by hand is rejected rather than passed through.

### Use Case

Backrest backing up a ZFS replication target — which is close to the app's central use case, since a backup app is very often pointed at exactly this kind of dataset.

Host layout, one dataset per app under the replication target:

```
data/backups/nexus/apps          <- bind-mounted into the Backrest container
data/backups/nexus/apps/plex     <- child dataset
data/backups/nexus/apps/immich   <- child dataset
...

```
zettarepl receives into every child dataset nightly (zfs recv -F), unmounting each target before the receive and remounting the subtree afterwards — the unmount in zettarepl/replication/run.py is unconditional. On this system that is roughly 40 unmount/remount steps per night.

Because the bind is rprivate, none of those remounts propagate into the container. The parent path stays visible; every child resolves to the parent's empty stub directory:

#### on the host
```
$ stat -c %d /mnt/data/backups/nexus/apps/plex
62                      # the child dataset
```
#### inside the Backrest container, same path
```
$ stat -c %d /backup/backups/nexus/apps/plex
60                      # the parent's stub — an empty directory
```
The failure is silent, and that's what makes it serious for a backup app. There is no error and no log line. restic walks an empty tree, captures zero files, and the plan reports success. On 2026-07-29 a nightly plan here reported success having backed up nothing; it only became visible because the deployment had an independent guard checking that the backup source was non-empty. Without such a guard, the symptom is a backup that has been green for weeks and contains nothing.

Restarting the app fixes it for exactly one cycle — until the next replication run.

### Proposed Solution

Expose mount propagation on host-path storage items in Backrest's questions.yaml, defaulting to rprivate so existing deployments are unchanged.

The host's own mounts are shared, so a slave bind has something to receive from — rslave is sufficient here and is the value I'd set.

If exposing all six propagation values is undesirable, a single boolean — "follow host mount changes", mapping to rslave — would cover this case and is probably easier to document.

### Alternatives Considered

- Convert Backrest to a Custom App, where the bind can be declared directly. This works, but gives up catalog updates entirely.
- A host-side cron job that compares each dataset's device number on the host against what the container resolves, and restarts the app on divergence. This is what I'm running now. It's a poll rather than a fix: it has to run frequently enough to catch off-schedule replication runs, and restarting the app mid-backup can strand an exclusive lock in the restic repository, so the workaround introduces a failure mode of its own.
- Reading from .zfs/snapshot instead of the live path. This cannot work from inside a container — the snapshot automount is performed via call_usermodehelper, which always mounts in the initial namespace. The container lists every snapshot name correctly and every one of them is empty, so this is strictly worse: it fails every time rather than only after replication.
- Setting canmount=noauto or mountpoint=legacy to stop the remounts. This stops the churn by making the tree invisible to everything else too.

### Additional Context

Verified on TrueNAS 25.10, Backrest catalog version 1.1.11, lib_version 2.3.4.

I've scoped this request to Backrest since that's where I hit it and where the silent-empty-backup consequence is worst, but the underlying gap isn't Backrest-specific — any catalog app pointed at a dataset tree that gets remounted has the same exposure, so a shared fix may be more useful to you than a per-app one.

ArLab1:
Closes #5556

Exposes mount propagation on Backrest's additional_storage host path config, so a host path can be mounted rslave and follow host mount changes.

### DIFF
--- a/ix-dev/community/backrest/app.yaml
+++ b/ix-dev/community/backrest/app.yaml
@@ -30,4 +30,4 @@ sources:
 - https://github.com/garethgeorge/backrest
 title: Backrest
 train: community
-version: 1.1.11
+version: 1.1.12
--- a/ix-dev/community/backrest/questions.yaml
+++ b/ix-dev/community/backrest/questions.yaml
@@ -524,6 +524,29 @@ questions:
                               type: hostpath
                               show_if: [["acl_enable", "=", false]]
                               required: true
+                          - variable: propagation
+                            label: Mount Propagation
+                            description: |
+                              Controls whether mount changes made on the host after the app starts
+                              are visible inside the container.</br>
+                              Private: Changes on the host are not propagated. (Default)</br>
+                              Slave: Mounts and unmounts performed on the host are propagated into
+                              the container. Use this when the host path is a ZFS dataset tree whose
+                              children are remounted while the app runs, such as a replication
+                              target. Without it, those datasets appear empty inside the container
+                              until the app is restarted.
+                            schema:
+                              type: string
+                              default: "rprivate"
+                              enum:
+                                - value: "rprivate"
+                                  description: Private (Recursive)
+                                - value: "rslave"
+                                  description: Slave (Recursive)
+                                - value: "private"
+                                  description: Private
+                                - value: "slave"
+                                  description: Slave
                     - variable: ix_volume_config
                       label: ixVolume Configuration
                       description: The configuration for the ixVolume dataset.
--- a/ix-dev/community/backrest/templates/test_values/basic-values.yaml
+++ b/ix-dev/community/backrest/templates/test_values/basic-values.yaml
@@ -44,4 +44,5 @@ storage:
       host_path_config:
         path: /opt/tests/mnt/backrest/backup-source
         acl_enable: false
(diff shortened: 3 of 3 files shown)


stavros-k:
There are no plans to expose that at this time.

PR #5587, opened 2026-08-13
Add t3monitoring

### LINKED ISSUE #5586: [App Request]: t3monitoring
### App Name

T3 Monitoring

### App Description

A self-hosted TYPO3 instance running the t3monitoring extension, which tracks the core version, extension versions and known security advisories of other TYPO3 installations from one dashboard.

### App Website / Github Repo

https://github.com/georgringer/t3monitoring

### Docker Installation Documentation

https://github.com/undkonsorten/pkgs/container/t3monitoring

### App Icon URL

_No response_

### Docker Image

ghcr.io/undkonsorten/t3monitoring

### App Category

Monitoring

### Requirements Checklist

- [x] The app is actively maintained
- [x] The app has official Docker images available
- [x] The app has versioned image tags available
- [x] The app is open source or has a free tier
- [x] I have tested the Docker installation locally

### Additional Information

Use case: agencies and teams running several TYPO3 sites use t3monitoring as a central
place to see which installations are behind on core/extension updates or affected by a
published TYPO3 security advisory. The extension is a well-established part of the TYPO3
ecosystem (t3monitor/t3monitoring on Packagist).

One thing to flag up front, so it is not a surprise during review: the upstream project
(georgringer/t3monitoring) does not publish a Docker image. t3monitoring is a TYPO3
extension, not a standalone application — it has to be installed into a TYPO3 instance.
The image at ghcr.io/undkonsorten/t3monitoring is therefore built and maintained by us
(undkonsorten): a TYPO3 13.4 instance with the extension pre-installed, tagged to match
the TYPO3 version it ships. It is public, versioned, and built reproducibly from a
committed Dockerfile.

If the catalog requires images published by the upstream project itself, we would rather
know now than after submitting the PR. We are happy to discuss alternatives.

The app is ready and validated locally; we will open the pull request once this issue has
been triaged.

Starkmann:
# App Addition

- [x] I have opened an [issue](https://github.com/truenas/apps/issues) to discuss this app addition before submitting this pull request.

Closes #5586

# AI

## Description
Adds **T3 Monitoring** to the community train.

t3monitoring is a TYPO3 extension that monitors *other* TYPO3 installations from one
place: their core version, their extension versions, and whether any of them are
affected by a published TYPO3 security advisory. This app packages it as a ready-to-run
instance — TYPO3 13.4 with the extension pre-installed, MariaDB, and a scheduler
container that performs the periodic checks.

Three containers plus the standard `permissions` init container:

| Container | Role |
| --- | --- |
| `t3monitoring` | TYPO3 + Apache, serves the backend at `/typo3/` |
| `t3monitoring-scheduler` | runs `typo3 scheduler:run` once a minute; the monitoring checks are scheduler tasks, so nothing is collected without it |
| `mariadb` | database |

## App Information

- **Upstream**: https://github.com/georgringer/t3monitoring
- **Documentation**: https://docs.typo3.org/p/t3monitor/t3monitoring/main/en-us/
- **Packagist**: https://packagist.org/packages/t3monitor/t3monitoring
- **Image**: `ghcr.io/undkonsorten/t3monitoring` (public, `linux/amd64`)
- **App Version**: 13.4.34 (tracks the TYPO3 version shipped)

## Testing

Tested locally with:

- [x] basic-values.yaml

Verified on a genuinely empty dataset (test mounts cleared between runs, since
`ci.py` reuses `/opt/tests/mnt` and would otherwise skip first-boot setup):

- `ci.py` render + install passes, no failed containers, no warnings in any container log
- the web container reaches `healthy`
- `/typo3/` returns **200** with the TYPO3 login form
- 58 tables created; the configured administrator exists and its password verifies
- scheduler logs zero errors
- `generate_metadata.py` produces no diff; `port_validation.py` and
  `apps_dev_charts_validate` both pass against current `master`

## Icons and Screenshots

- Icon: **[attach `truenas-app-icon.svg`]** — `app.yaml`/`item.yaml` already reference
  `https://media.sys.truenas.net/apps/t3monitoring/icons/icon.svg`

  This is the t3monitoring extension's own icon, taken unmodified from
  `Resources/Public/Icons/Extension.svg` in the upstream extension (GPL-2.0+, same
  licence as the extension itself): a TYPO3-orange rounded square with a pulse line.
  Note the pulse is a transparent cutout rather than a white shape, so it picks up
  whatever background it is composited on — legible on both light and dark cards. Happy
  to supply a variant with an opaque white pulse if the catalog prefers icons that
  render identically in both themes.

- Screenshots: none for now; happy to add backend screenshots if useful.

## Special Notes

**First-time setup is automatic.** A fresh install gets an empty database, so the image
creates the schema on first boot (`database:updateschema` → `extension:setup` →
`backend:createadmin` → `cache:flush`) using the administrator credentials entered in the
installation wizard. This takes a minute or two, which is why the web container has a
300s healthcheck `start_period`. Only the web container performs setup —
`T3M_DB_AUTO_SETUP=1` is set on it alone, because both containers run the same image
against the same database and concurrent DDL would corrupt the schema; the scheduler
waits for the tables it needs.

**The public frontend returns 404 on a fresh install** (`No site configuration found`).
This is expected and documented in the app notes: the bundled TYPO3 site configuration
points at a root page that only exists once the operator creates content. t3monitoring is
used entirely from the backend, so it does not affect the app's function.

**The Install Tool password is entered as plain text** and hashed inside the container on
startup (argon2i). TYPO3 compares that setting against a stored hash, so it would
otherwise force operators to paste a pre-hashed `$argon2i$…` string into the wizard.

**Port**: the published default is **30469**, verified unique against current `master`.

**About the image**: t3monitoring is a TYPO3 *extension*, so the upstream project does not
publish a Docker image — there is nothing standalone to containerise. The image here is
built and maintained by undkonsorten from a committed Dockerfile, tagged to match the
TYPO3 version, and contains no Composer dev dependencies (enforced at build time). This
was raised in the issue before opening this PR.

## Checklist

- [x] README.md included


ix-dev/community/t3monitoring/ix_values.yaml
```diff
@@ -0,0 +1,60 @@
+images:
+  image:
+    # Built from deploy/Dockerfile in the t3monitoring-dev repo and pushed here.
+    # Update both this tag and app.yaml's `version` (and `app_version` if the
+    # upstream TYPO3/t3monitoring version changed) whenever you push a new build.
+    repository: ghcr.io/undkonsorten/t3monitoring
+    tag: "13.4.34"
+  mariadb_image:
+    repository: mariadb
+    tag: "10.11"
```
stavros-k: it should use current version of mariadb, as the rest of the apps.

ix-dev/community/t3monitoring/ix_values.yaml
```diff
@@ -0,0 +1,60 @@
...
+  container_utils_image:
+    repository: ixsystems/container-utils
+    tag: 1.0.2
+
+consts:
+  web_container_name: t3monitoring
+  scheduler_container_name: t3monitoring-scheduler
```
stavros-k: that will create a stuttering container name repeateing the app name

ix-dev/community/t3monitoring/ix_values.yaml
```diff
@@ -0,0 +1,60 @@
...
+  mariadb_container_name: mariadb
+  perms_container_name: permissions
+  db_user: typo3
+  db_name: t3monitoring
```
stavros-k: db user should match db name. you should follow codebase patterns

ix-dev/community/t3monitoring/ix_values.yaml
```diff
@@ -0,0 +1,60 @@
...
+  # Matches the www-data uid/gid baked into deploy/Dockerfile.
+  run_user: 33
+  run_group: 33
+  internal_web_port: 80
+  notes_body: |
```
stavros-k: This is not what this is for. 
Notes are not a FAQ.

ix-dev/community/t3monitoring/README.md
stavros-k, line 1: follow codebase patterns

ix-dev/community/t3monitoring/templates/docker-compose.yaml
stavros-k, line 1: Your LLM, accidentally wrote his thinking in this file.

ix-dev/community/t3monitoring/templates/docker-compose.yaml
```diff
@@ -0,0 +1,126 @@
...
+  "root_password": values.t3monitoring.db_root_password,
+  "password": values.t3monitoring.db_password,
+  "database": values.consts.db_name,
+  "volume": values.storage.db_data,
+} %}
+{% set mariadb_container = tpl.deps.mariadb(values.consts.mariadb_container_name, "mariadb_image", mariadb_config, perm_container) %}
+{% do mariadb_container.container.add_network(t3_net) %}
+{# The fixture DB predates utf8mb4 and expects classic utf8/utf8_unicode_ci;
+   deps.mariadb() only sets --port by default, so re-set the full command. #}
+{% do mariadb_container.container.set_command(["--port", "3306", "--character-set-server=utf8", "--collation-server=utf8_unicode_ci"]) %}
```
stavros-k: I'm afraid that this is not allowed.

stavros-k (changes requested):
I've bailed mid-review, as it was like I was reading an LLM thinking block.
But I saw that there are many things that doesn't follow codebase pattern (Known llm artifact).

You have to review before you push your changes.

Thanks

ix-dev/community/t3monitoring/ix_values.yaml
Starkmann: Done. Used 12.3.2

ix-dev/community/t3monitoring/ix_values.yaml
Starkmann: Renamed to scheduler

ix-dev/community/t3monitoring/ix_values.yaml
Starkmann: Both now t3monitoring

ix-dev/community/t3monitoring/ix_values.yaml
Starkmann: I shortened the notes

ix-dev/community/t3monitoring/README.md
Starkmann: Adjusted.

ix-dev/community/t3monitoring/templates/docker-compose.yaml
Starkmann: Removed all the comments

ix-dev/community/t3monitoring/ix_values.yaml
stavros-k: Still this block is useless. Those are not visible until the app is up and running.

ix-dev/community/t3monitoring/templates/docker-compose.yaml
Starkmann: Removed the set_command

ix-dev/community/t3monitoring/ix_values.yaml
```diff
@@ -0,0 +1,33 @@
+images:
+  image:
+    # Built from deploy/Dockerfile in the t3monitoring-dev repo and pushed here.
+    # Update both this tag and app.yaml's `version` (and `app_version` if the
+    # upstream TYPO3/t3monitoring version changed) whenever you push a new build.
```
stavros-k: That seems a directive from LLM towards you..?

Starkmann: Ah see the problem, the truenas-apps it a subrepo for our develompent suite for t3monitoring and TYPO3.
So I wrote some comments for my colleagues. But then this is not the right place to do so.

Starkmann: Removed

ix-dev/community/t3monitoring/ix_values.yaml
Starkmann: Removed the notes block

ix-dev/community/t3monitoring/app.yaml
```diff
@@ -0,0 +1,54 @@
...
+  group_name: Host group is [root]
+  uid: 0
+  user_name: Host user is [root]
+screenshots: []
+sources:
+- https://apps.truenas.com/catalog/t3monitoring_community/
+- https://github.com/georgringer/t3monitoring
+- https://github.com/undkonsorten/pkgs/container/t3monitoring
+- https://packagist.org/packages/t3monitor/t3monitoring
+- https://undkonsorten.com
```
stavros-k: What exactly is this?

stavros-k:
I cannot push on your branch.

ix-dev/community/t3monitoring/app.yaml
Starkmann: https://github.com/georgringer/t3monitoring
https://packagist.org/packages/t3monitor/t3monitoring
The extension for TYPO3 that does the monitoring

https://github.com/orgs/undkonsorten/packages/container/package/t3monitoring (i correct this)
The image of the truenas app

https://undkonsorten.com
sponsor for this app

Starkmann:
It is a fork. Not sure what you mean.

stavros-k:
It does not let me push. If its under an organization, thats the problem.
Otherwise you have not checked the box on the right that says that you allow maintainers to push

Starkmann:
Sorry did not know the the organization is a problem, now you have write access.

ix-dev/community/t3monitoring/app.yaml
stavros-k: Sorry, but none of those provide any info regarding this container.
Even the "home" is pointing to the extension source, but docs or "source" of the container and how to use it is nowhere to be found.

stavros-k:
Does TYPO3 support having more extensions in parallel or only one at a time (ie the monitoring one)? 
How does someone add those?

ix-dev/community/t3monitoring/templates/docker-compose.yaml
```diff
@@ -0,0 +1,70 @@
...
+{% do mariadb.container.add_network(t3_net) %}
+
+{% set web = tpl.add_container(values.consts.t3monitoring_container_name, "image") %}
+{% do web.environment.add_env("T3M_DB_AUTO_SETUP", "1") %}
+{% do web.healthcheck.set_test("curl", {"port": values.consts.internal_web_port, "path": "/typo3/", "headers": [["Host", values.t3monitoring.server_name]]}) %}
+{% do web.add_port(values.network.web_port, {"container_port": values.consts.internal_web_port}) %}
+
+{% set scheduler = tpl.add_container(values.consts.scheduler_container_name, "image") %}
+{% do scheduler.set_command(["sh", "-c", "while true; do vendor/bin/typo3 scheduler:run || echo \"scheduler:run failed with exit $?\" >&2; sleep 60; done"]) %}
+{% do scheduler.healthcheck.disable() %}
```
stavros-k, line 25: This will exit on first hiccup and without healthcheck no-one will ever know.

Doesnt the image have a separate entrypoint for scheduler?

ix-dev/community/t3monitoring/questions.yaml
```diff
@@ -0,0 +1,739 @@
...
+            The public hostname this instance is served under, without scheme or
+            trailing slash (e.g. t3monitoring.example.com). TYPO3's site
+            configuration builds its base URL from this
+            (base: 'https://%env(SERVER_NAME)%/'), so leaving it wrong makes the
+            frontend fail to resolve a site and return a 500.
+          schema:
+            type: string
+            default: ""
+            required: true
+        - variable: force_backend_https
```
stavros-k, line 56: With this enabled, won't the healtcheck fail?

stavros-k:
@Starkmann ping

Starkmann:
> Does TYPO3 support having more extensions in parallel or only one at a time (ie the monitoring one)? How does someone add those?

TYPO3 supports many extensions in parallel. This is just one extension which is widely used for monitoring TYPO3 instances. The extensions can added via the TYPO3 extension repository (https://extensions.typo3.org/extension) or composer. Composer is the recommended  option.

ix-dev/community/t3monitoring/templates/docker-compose.yaml
Starkmann: TYPO3 doesn't ship a persistent scheduler daemon — scheduler:run is a one-shot CLI command meant to be triggered periodically (traditionally via a real cron entry), which is why this wraps it in a loop rather than a dedicated entrypoint. Added a pidof-based liveness healthcheck on the loop, same pattern this repo already uses for Nextcloud's own cron container, so a wedged/dead loop now actually shows up instead of looking "running" forever.

ix-dev/community/t3monitoring/questions.yaml
Starkmann: Fixed by sending X-Forwarded-Proto: https on the healthcheck request when this option is on, same as a real TLS-terminating proxy would — restores full-depth checking.

Starkmann:
> @Starkmann ping

I was on holidays for a week.

stavros-k:
> Does TYPO3 support having more extensions in parallel or only one at a time (ie the monitoring one)? How does someone add those?

It feels weird to have an app that just adds an extension. 
What happens if someone wants another extension added or swapped? 
Would we have to add t3 x number of extension apps? Or t3 x number of combos of extensions?
That doesnt sound logical.

Starkmann:
> > Does TYPO3 support having more extensions in parallel or only one at a time (ie the monitoring one)? How does someone add those?
> 
> It feels weird to have an app that just adds an extension. What happens if someone wants another extension added or swapped? Would we have to add t3 x number of extension apps? Or t3 x number of combos of extensions? That doesnt sound logical.

The purpose of this is to have a TYPO3 with monitoring possibility for other TYPO3 instances that can be hosted worldwide. That's what the extension is for.

So yeah, we could have another app which is called TYPO3. And then we have a bear TYPO3 management system without any extensions. Think of it like the WordPress app, for example. On this app people can install or manage TYPO3 as they want. And could of course also install the t3monitoring extension. Bash into the container and install via composer some extensions is possible, of course, but not handy and not the intention of this app.

stavros-k:
If its like wordpress, I dont see the reason to have a "wordpress with preinstalled extension" kinda app? 
Do I miss something?

Starkmann:
Well it actually comes down to how usable it is. You could also say: Why is there a nextcloud app when we have a debian app where you can install nextcloud into by yourself.

The extensions in TYPO3 is nothing you can click in the backend and then they are downloaded and working like in WordPress. You really have to bash into the container and do a couple of shell commands until the extension is up and installed.

So yeah, I understand your point. My goal was to have a one click install of the functionality the extension actually provides.  A tool for admins to control and monitor other TYPO3 instances.

stavros-k:
Well, no nextcloud example does not make sense.

My concern is that the moment this lands, where do you draw the line when another user say "I just want Typo3 with THAT extension as one click install?"

ix-dev/community/t3monitoring/app.yaml
Starkmann: https://github.com/undkonsorten/truenas-t3monitoring/pkgs/container/t3monitoring
I added the documentation here.

Starkmann:
You are right. It's hard to draw that line. In my head when I thought about this was that this is a rather unique extension that only has an administration purpose. And if somebody still wants to use this with another extension, it is still possible. 
So maybe as a compromise we could name the app TYPO3 and it comes with preinstalled t3monitoring extension, i could check that.

Starkmann:
Following up on plain-TYPO3-vs-t3monitoring — I checked what's actually t3monitoring-specific in this repo, and whether extensions could be added afterward.

The app is generic TYPO3 plus one composer line. composer.json is the stock typo3/cms-base-distribution set + t3monitor/t3monitoring. Everything else (Dockerfile, entrypoint, scheduler, healthchecks) is generic Composer-mode TYPO3 machinery — the T3M_ naming is cosmetic. A plain app is trivially buildable from the same pipeline.

 Whether a user can add extensions later depends on Composer vs. Classic mode:

 - This app is Composer mode, where there's no runtime extension-install path at all, for any extension — that's been TYPO3 core policy since v11, and this image doesn't even ship composer or a writable vendor/. So the WordPress-plugin comparison doesn't hold: any extension change means a new image, built by me, always.
 
- Classic mode does give real click-to-install via TER (t3monitoring is published there), but it's a different installation type, not a flag — needs a rebuilt TYPO3 layer, a writable+persisted extension volume, and permanent outbound access to extensions.typo3.org, none of which this offline/reproducible image has today. It's also the path TYPO3 core itself steers people away from for new installs.

So "plain now, add extensions later" isn't really on the table without a much bigger rearchitecture. Given that, I'd rather keep Composer mode and do the compromise I mentioned earlier: rename the app to TYPO3, ship it with t3monitoring preinstalled, and treat other extension requests as additional image tags of the same app rather than new apps. Happy to do that rename/cleanup if it resolves this for you.

stavros-k:
> ship it with t3monitoring preinstalled, and treat other extension requests as additional image tags of the same app rather than new apps.

That doesn't really work. Any extension requires _you_ to provide the tag.
And how does it work if someone was plugin A and plugin B but not plugin C because it collides?


Apps that rely too much on plugins (ie netbox) and have no built in way for the user to install those, (ie wp, nextcloud, etc), are intentionally "not added yet" because its hard problem to solve. At least for the given case here where there is just a template rendered.

Starkmann:
Okay, I see. I can understand that.

Starkmann:
Sorry, accidentally closed.

So this is the reason why I wanted to just have the functionality of monitoring the instances. We don't even have front end here, just backend. Just the monitoring feature. You can also think of it like it is a tool for monitoring WordPress which is existing in the catalog as well.

stavros-k:
I get that, but if an app is accepted with a specific plugin bundled in, then we have to start accepting the "same" app with a different plugin bundled in. And there are hundreds if not more plugins.

PR #5588, opened 2026-08-14
Zammad: Update to version 6.5.4

kunago:
# Pull Request

Update Zammad from version 6.5.2 to 6.5.4 fixing some backwards implemented security issues.

Also update Postgres helper image from 18.4 to 18.6. which is available mainstream.


stavros-k:
Updates should be handled by the bot, if they are not picked, the renovate-config.js should be updated to fix it.

PR #5596, opened 2026-08-14
golmenero/ratelog: Updates environments Variables

golmenero:
Add existing TMDB Api Key env variable.

Previously, the user had to add it manually via environment variables.


ix-dev/community/ratelog/questions.yaml
```diff
@@ -42,6 +42,13 @@ questions:
             default: ""
             required: true
             private: true
+        - variable: tmdb_api_key
+          label: TMDB Api Key
+          schema:
+            type: string
+            default: ""
+            required: true
```
stavros-k, line 50: This will prohibit anyone to upgrade, as it will require a key which they cant set during upgrade.

golmenero: Unfortunately this variables is indeed mandatory. I dont know what the orocess should be now

stavros-k: Is it failing to install without it? (if thats the case, then there wont be any installs that we will break)

golmenero: It definetly wont work. The env var is definetly needed for the app to work

stavros-k (approved):

PR #5629, opened 2026-08-20
feat(komodo) Allow adding capabilities to periphery container

nicolas-r:
# Summary

Provide to the user the possibility to add capabilities to the periphery container

## Changes

* Add a block (taken from the ix-app) to add the capabilities in questions.yaml 
* Add jinja code in docker-compose.yaml
* Add an entry for capabilities, empty, in basic-values.yaml

## Testing

Tested locally with:

- [ ] basic-values.yaml

Also tested with another file with non empty capabilities and the docker compose was good


nicolas-r:
Thank you for the new commit

stavros-k (approved):

PR #5646, opened 2026-08-23
fix(home-assistant): serve on fixed internal 8123 and map published port to it

### LINKED ISSUE #5645: [Bug Report]: home-assistant unreachable/unhealthy on any custom WebUI port (HA stays on 8123)
### App Name

home-assistant

1.8.54 (stable), HA image 2026.8.2

### App Issue

Deploying Home Assistant with any custom "WebUI Port" other than the container default `8123` makes the app **unreachable** on that port and permanently `unhealthy`. Setting the WebUI port to `8123` is the only value that works.

**Root cause:** the chart tries to move HA's listen port to the user-chosen port by writing `http.server_port` into `configuration.yaml` (see `templates/macros/config.sh.jinja`), then maps `host_port -> host_port` and healthchecks `host_port`. But recent Home Assistant stages any `http` config change (port or TLS) as **pending** and only promotes it after confirming the new endpoint is reachable. On a headless TrueNAS deploy that confirmation never happens, so:

- HA keeps serving on the **stable** default `8123`.
- The published port maps `host_port -> host_port`, where **nothing listens** -> "Unable to connect".
- The healthcheck targets `host_port` -> never succeeds -> `unhealthy`.
- The unpromoted pending config is also what triggers the earlier "Pending HTTP config was not confirmed ... exit code 100" first-boot timeout some users hit.

It is a deadlock: HA won't switch to the custom port until it's reachable, and it can't be reachable until HA switches.

### Reproduction

1. Install `home-assistant` (stable), leave the default WebUI port `30103` (or any non-8123 value), no certificate.
2. Wait for boot. App shows `Up (unhealthy)`; `http://<nas>:30103/` gives "Unable to connect".

### Evidence (clean stock install, port 30103, no cert)

```text
# docker ps -> published maps 1:1 as the chart intends
ix-home-assistant-home-assistant-1  0.0.0.0:30103->30103/tcp  Up (unhealthy)

# but HA actually listens on 8123, not 30103
$ ss -tlnp | grep python
tcp  0.0.0.0:8123  LISTEN  72/python3

# configuration.yaml has the port the chart asked for
$ cat /config/configuration.yaml | grep -A1 '^http:'
http:
  server_port: 30103

# ...yet HA ignores it, because .storage/http keeps stable=8123 and pending=30103 unpromoted
$ cat /config/.storage/http
{
  "data": {
    "stable":  { "server_port": 8123, ... },
    "pending": { "server_port": 30103, ..., "error": "not_promoted" }
  }
}

# HA serves fine on 8123
$ wget -qO- http://127.0.0.1:8123/manifest.json
{"background_color":"#FFFFFF", ... }
```

### Proposed fix

Stop reconfiguring HA's `server_port`; keep HA on its fixed default `8123` and map the published port to it (the same "fixed internal port" pattern ~118 other apps in this catalog already use):

- `ix_values.yaml`: add `consts.web_port: 8123`.
- `templates/docker-compose.yaml`: `add_port(values.network.web_port, {"container_port": values.consts.web_port})` and healthcheck `port: values.consts.web_port`.
- `templates/macros/config.sh.jinja`: remove the `http.server_port` rewrite.

This also removes the "pending not confirmed -> exit 100" risk. PR attached.

### Note (related, out of scope)

The built-in certificate option drives HA's TLS through the same `configuration.yaml` -> pending mechanism, so it is affected by the same deadlock. Recommend terminating TLS at a reverse proxy. Can be addressed separately.

julian-alarcon:
## Summary

Fixes #5645.

Home Assistant is unreachable and stays `unhealthy` when deployed on any WebUI port other than the container default `8123`.

The chart currently rewrites HA's `http.server_port` in `configuration.yaml` to the chosen host port, maps `host_port -> host_port`, and healthchecks `host_port`. But recent Home Assistant stages every `http` config change as **pending** and only promotes it once the new endpoint is confirmed reachable. On a headless TrueNAS deploy that confirmation never happens, so HA keeps listening on the **stable** default `8123` while nothing listens on the published port. Result: "Unable to connect" + failing healthcheck. (The same unpromoted-pending path is behind the "Pending HTTP config was not confirmed ... exit code 100" first-boot timeout.)

## Fix

Adopt the "fixed internal port" convention already used by ~118 apps in this catalog: keep HA on `8123` and map the published port to it.

- `ix_values.yaml`: add `consts.web_port: 8123`.
- `templates/docker-compose.yaml`: `add_port(values.network.web_port, {"container_port": values.consts.web_port})`; healthcheck `port: values.consts.web_port`.
- `templates/macros/config.sh.jinja`: remove the `http.server_port` rewrite (the pending-config trigger).
- `app.yaml`: version bump `1.8.54 -> 1.8.55`.

Portal and healthcheck scheme (proto based on `certificate_id`) are unchanged; the portal still advertises the published host port, which is correct.

## Evidence

Clean stock install, port `30103`, no cert:

```text
docker ps          -> 0.0.0.0:30103->30103/tcp  Up (unhealthy)
ss -tlnp           -> HA LISTENs on 8123, not 30103
configuration.yaml -> http.server_port: 30103   (written, but ignored)
.storage/http      -> stable=8123, pending=30103 "error": "not_promoted"
wget 127.0.0.1:8123/manifest.json -> 200 OK
```

## Testing

Relying on `app-test-suite` CI to render + boot the app.

## Note (out of scope)

The built-in certificate/TLS option drives HA TLS through the same `configuration.yaml` -> pending path, so it shares the deadlock. Recommend terminating TLS at a reverse proxy; can be handled in a follow-up.

### DIFF
--- a/ix-dev/stable/home-assistant/app.yaml
+++ b/ix-dev/stable/home-assistant/app.yaml
@@ -49,4 +49,4 @@ sources:
 - https://github.com/home-assistant/home-assistant
 title: Home Assistant
 train: stable
-version: 1.8.54
+version: 1.8.55
--- a/ix-dev/stable/home-assistant/ix_values.yaml
+++ b/ix-dev/stable/home-assistant/ix_values.yaml
@@ -26,3 +26,6 @@ consts:
   config_path: /config
   ssl_key_path: /ssl/key.pem
   ssl_cert_path: /ssl/cert.pem
+  # Home Assistant always listens on 8123 internally. The published host port is
+  # mapped to this, instead of reconfiguring HA's server_port (see config.sh.jinja).
+  web_port: 8123
--- a/ix-dev/stable/home-assistant/templates/docker-compose.yaml
+++ b/ix-dev/stable/home-assistant/templates/docker-compose.yaml
@@ -50,10 +50,10 @@
 {% endif %}
 
 {% set proto = "https" if values.network.certificate_id else "http" %}
-{% do c1.healthcheck.set_test("wget", {"port": values.network.web_port.port_number, "path": "/manifest.json", "scheme": proto, "spider": false}) %}
+{% do c1.healthcheck.set_test("wget", {"port": values.consts.web_port, "path": "/manifest.json", "scheme": proto, "spider": false}) %}
 {% do c1.environment.add_user_envs(values.home_assistant.additional_envs) %}
 
-{% do c1.add_port(values.network.web_port) %}
+{% do c1.add_port(values.network.web_port, {"container_port": values.consts.web_port}) %}
 {% for port in values.network.additional_ports %}
   {% do c1.add_port(port) %}
 {% endfor %}
--- a/ix-dev/stable/home-assistant/templates/macros/config.sh.jinja
+++ b/ix-dev/stable/home-assistant/templates/macros/config.sh.jinja
@@ -29,9 +29,6 @@ if ! yq --exit-status '.http' < "{{ config_file }}" &> /dev/null; then
   yq -i '.http = {}' "{{ config_file }}"
 fi
 
-echo "Matching server port to configured port [{{ values.network.web_port.port_number }}] in [{{ config_file }}]"
-yq -i '.http.server_port = {{ values.network.web_port.port_number }}' "{{ config_file }}"
-
 {%- if values.network.certificate_id %}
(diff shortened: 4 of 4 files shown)


stavros-k:
There is an issue now tho, existing installs that HA migrated their config file to db-stored config, will hold their custom port. Which will break by setting it to 8123.

PR #5653, opened 2026-08-24
Add Nexview to the community train

DerKezorm:
# App Addition

- [x] I have opened an [issue](https://github.com/truenas/apps/issues/5652) to discuss this app addition before submitting this pull request.

# AI

- [x] Part or All of this PR was generated by an LLM.

The app files, `questions.yaml` and the compose template were drafted with an LLM working from
existing community apps (`radarr`, `seerr`, `cantinarr`) as the reference, then tested locally
as described below. I am the author of the upstream app and have reviewed everything here.

## Description

Adds **Nexview** to the community train.

Nexview is a media discovery and request dashboard for a household. It shows new and upcoming
releases from TMDB, marks what Radarr and Sonarr already have, and turns a request into an add
on the right instance with the right quality profile and root folder. It downloads nothing
itself and stores no media — it sits in front of an *arr setup the user already runs.

Every household member has their own account. Admins approve or decline requests, set monthly
quotas per person, block genres or certifications per person, and can switch an account into a
children's view with large category tiles and an age filter. The interface is available in
English and German.

## App Information

- **Upstream**: https://github.com/DerKezorm/nexview
- **Documentation**: https://github.com/DerKezorm/nexview#readme
- **Project site**: https://nexview.nexapps.dev
- **App Version**: 0.17.0
- **Image**: `ghcr.io/derkezorm/nexview`, multi-arch (`linux/amd64`, `linux/arm64`), semver tags
- **Licence**: AGPL-3.0-or-later (the project moved from MIT to AGPL while this PR was open; versions up to 0.17.0 remain MIT)

## Testing

Tested locally with:

- [x] basic-values.yaml
- [x] advanced-values.yaml

Both render cleanly and reach `healthy`, and the web UI answers `200` on the published port
(`8080` and `8081` respectively). `advanced-values.yaml` covers every exposed option: secret key
set, log level set, extra environment variable, non-default token lifetimes and timezone, and a
second read-only storage entry — so the conditional branches in the compose template are
exercised, not just the defaults.

**One note on how it was tested.** `./.github/scripts/ci.py` does not run on a Windows host:
Docker Desktop bind mounts do not support `renameat2`, which `apps_catalog_hash_generate` needs,
so it fails before doing any work. I ran the same two steps from the same
`ghcr.io/truenas/apps_validation` image against a Docker volume instead, then brought the
rendered compose up with `docker compose` and waited for health. The host is arm64, so for the
local run only, the platform of the app container was switched to `linux/arm64`
[...]

PR #5656, opened 2026-08-25
nexview: configurable container port and caller-address setting

DerKezorm:
Follow-up to #5653, with the change @stavros-k asked for while merging it.

## The port

> It would be nice to have it configurable via env var, so when a user selects host network, can still configure it, avoiding any potential conflicts on the host.

Nexview 0.20.0 reads `NEXVIEW_PORT` for exactly that, and this exposes it as **Container Port** (default 8000).

All three places that assumed a fixed 8000 now follow the setting — the port mapping, the healthcheck and the portal. The healthcheck matters: knocking on 8000 while the app listens elsewhere would mark the container unhealthy. The `internal_web_port` const is gone, since nothing needs it any more.

## The caller address

Also new in 0.20.0: Nexview slows sign-ins down after several wrong passwords and then closes the door for fifteen minutes. The lock opens again by itself.

Attempts are always counted **per account**, which needs no configuration. Counting per **address** is optional and has to be told how Nexview can learn the real one — behind a reverse proxy every request looks like it comes from the same address, so guessing there would lock out a whole household on the first typo. Hence an enum with an explicit "per account only" default rather than a free-text field.

## Verification

Rendered both test value files against the vendored library:

| | `basic-values.yaml` | `host-network-values.yaml` |
|---|---|---|
| `NEXVIEW_PORT` | 8000 | 8123 |
| `NEXVIEW_CLIENT_IP` | *(empty)* | `proxy` |
| ports | 8080 → 8000 | none (host network) |
| healthcheck | `127.0.0.1:8000` | `127.0.0.1:8123` |

The second values file is new and covers host networking with a moved port — the branch this change is about. It uses `ix_volume`, not `host_path`.

Image pinned to 0.20.0, since both variables only exist from that version. Chart version 1.0.0 → 1.1.0.


ix-dev/community/nexview/questions.yaml
```diff
@@ -60,6 +60,45 @@ questions:
             min: 1
             max: 365
             required: true
+        - variable: port_number
```
stavros-k: please please please.
See other apps how its done.

stavros-k (changes requested):

stavros-k (approved):

PR #5660, opened 2026-08-26
hermes: gpu

### LINKED ISSUE #5657: [Enhancement]: Option to pass GPU to Hermes
### App Name

Hermes

v2026.8.19

### Enhancement Description

Hermes can run various apps and tools, based on what user wants. That means, there are tools which can use GPU acceleration as well.

### Use Case

In my case, it would help accelerate faster-whisper using GPU acceleration.

### Proposed Solution

Add into YAML definition for choosing GPU to pass into the container.

### Alternatives Considered

In mean time, I think I will need to convert it to custom app.

### Additional Context

_No response_

stavros-k:
Closes #5657

### DIFF
--- a/ix-dev/community/hermes-agent/app.yaml
+++ b/ix-dev/community/hermes-agent/app.yaml
@@ -40,4 +40,4 @@ sources:
 - https://hub.docker.com/r/nousresearch/hermes-agent
 title: Hermes-Agent
 train: community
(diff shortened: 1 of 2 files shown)


Qubad786 (approved):

gamelaster:
Thank you @stavros-k :)

PR #5665, opened 2026-08-26
home-assistant: fix port/ssl / deprecates HA app

### LINKED ISSUE #5645: [Bug Report]: home-assistant unreachable/unhealthy on any custom WebUI port (HA stays on 8123)

stavros-k:
Supersedes #5646
Closes #5645

Additionally deprecates HA app, as it is much more usable as a VM (HA OS).

### DIFF
--- a/ix-dev/stable/home-assistant/app.yaml
+++ b/ix-dev/stable/home-assistant/app.yaml
@@ -49,4 +49,4 @@ sources:
 - https://github.com/home-assistant/home-assistant
 title: Home Assistant
 train: stable
-version: 1.8.55
+version: 1.8.56
--- a/ix-dev/stable/home-assistant/app.yaml
+++ b/ix-dev/stable/home-assistant/app.yaml
@@ -47,6 +47,6 @@ screenshots:
 sources:
 - https://apps.truenas.com/catalog/home-assistant_stable/
 - https://github.com/home-assistant/home-assistant
-title: Home Assistant
+title: Home Assistant (Deprecated)
 train: stable
 version: 1.8.56
--- a/ix-dev/stable/home-assistant/deprecations.yaml
+++ b/ix-dev/stable/home-assistant/deprecations.yaml
@@ -1,0 +1,19 @@
+- scope: "full"
+
+  deprecated_date: "2026-08-26"
+  removal_date: "2026-11-26"
+  reason: >-
+    This app is deprecated and will be removed in a future release.
+    Run Home Assistant OS in a VM instead: it gets direct access to the host network,
+    so mDNS/zeroconf, DHCP, SSDP/UPnP, Bluetooth, HomeKit and Thread/Matter discovery all work,
+    and it can run the add-on ecosystem (ESPHome, Zigbee2MQTT, Mosquitto, Node-RED)
+    along with Supervisor-managed updates and backups.
+    USB radios such as Zigbee and Z-Wave sticks can be passed through to the VM.
+    Use the KVM (.qcow2) image from https://www.home-assistant.io/installation/linux
+    and move your existing setup across by creating a full backup here first, then restoring it
+    during onboarding: https://www.home-assistant.io/common-tasks/general/#restoring-a-backup.
+    Home Assistant's backup only covers its configuration directory. This app's PostgreSQL
+    database, where history and long-term statistics live, is not included in it and goes away
+    with the app. Home Assistant OS uses SQLite for the recorder (by default), so if you want to keep that
+    history, migrate the database before removing this app:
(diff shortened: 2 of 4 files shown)


Qubad786 (approved):

dori4n:
Hi, um... it just came up in the Discord server that it's deprecated. Could you put it back, please?

Based on https://www.home-assistant.io/installation/#about-installation-types the only real difference between HomeAssistant Container (Docker) and HomeAssistant OS (VM) is the support for apps (Docker containers!) within HomeAssistant itself. And that some integrations, chief among these MQTT, Z-Wave, Thread, and the like, are provided through apps. This lends itself to the issue, that HomeAssistant cannot use these *from within HomeAssistant*, which is a non-issue, because you can just install and configure said apps in TrueNAS' Docker integration as well. I have done this, and it works great. In fact, I prefer it over the VM solution, because the RAM usage is lower this way. RAM's expensive these days. ZFS likes it too. I'd rather not allocate it to HAOS needlessly so it sits idle, because HomeAssistant doesn't need it, but throws a fit on startup if it's not there.

ReFil:
Deprecating the home assistant app (a stable train app used by 10k+ users according to the catalog) in favour of spooling up a vm with the officially recommended DB migration solution being downloading CSV files and running manual SQL queries seems like a surefire way of causing all sorts of problems. Home assistant container remains an officially supported and updated installation method by the home assistant team, so I struggle to see the argument for deprecating it. As per the deprecation comment direct access to the host network is a double edged sword, there are benefits to containerisation too

PR #5703, opened 2026-08-31
fix(stdapi-ai): correct the listing description and README

JGoutin:
The catalogue text overstated and misdescribed the app:

- "80+ models" is stale. The gateway now fronts 100+ models from 20+ providers.
- "enterprise compliance" implies certifications the project does not hold.
- "pay-per-use pricing" is wrong for this listing. The image published here is the AGPL-3.0 community build, which costs nothing. AWS bills the user directly for model usage, with no markup.

Also name the other AWS services the gateway fronts (Polly, Transcribe, Comprehend) and the Cohere protocol, and mention the non-chat endpoints, so the card reflects what a reader actually gets.

Bump version 1.0.18 to 1.0.19 for the metadata change, and tag the two outbound stdapi.ai links so the maintainer can measure how much traffic this listing sends. changelog_url and sources are left untagged.

### DIFF
--- a/ix-dev/community/stdapi-ai/README.md
+++ b/ix-dev/community/stdapi-ai/README.md
@@ -1,4 +1,4 @@
 # stdapi.ai
 
-[stdapi.ai](https://stdapi.ai) is an OpenAI and Anthropic compatible API gateway for AWS Bedrock.
-Access 80+ models including Claude, Nova, Deepseek, Mistral, and Minimax with enterprise compliance and pay-per-use pricing.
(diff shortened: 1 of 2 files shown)


stavros-k:
Maybe a more generic description would be better? Otherwise we will have to update the description every other week :D

ix-dev/community/stdapi-ai/README.md
stavros-k, line 1: README is intentionally _lean_ and points to upstream (with CLEAN links). 
So we dont have to keep updating the descriptions.

JGoutin: OK, I will update this.

stavros-k (approved):

PR #5736, opened 2026-09-03
cantinarr: mention music in the description and keywords

windoze95:
Cantinarr 0.8.0 added a Lidarr music module. This updates the listing copy to match: the `description` now says "movies, TV, books, and music" and names Lidarr beside the other arrs, `keywords` gain `lidarr` and `music`, and the `CANTINARR_ARR_CALLBACK_URL` question's hint names Lidarr among the containers that call back. `version` bumped 1.0.6 → 1.0.7. No values, ports, or storage changes.
Reviewed by human developers.

PR #5739, opened 2026-09-04
cantinarr: point the push gateway default at push.cantinarr.com

windoze95:
Cantinarr's community push relay moved to `https://push.cantinarr.com`. This changes the default offered for the push gateway question and bumps the app version.

Existing installs are unaffected: their stored answer keeps pointing at `push.julian.codes`, which still answers from the same gateway, and current Cantinarr images rewrite that name to the new one at start. No variable keys change.


stavros-k:
https://github.com/truenas/apps/pull/5736#pullrequestreview-5104239915

PR #5772, opened 2026-09-10
Update ActuallyFreeCD to 1.0.1

D69wookie:
This updates ActuallyFreeCD to catalogue version 1.0.1.

Changes:
- Fixes Additional Environment Variables wiring.
- Adds configurable output UID, GID, directory mode and file mode settings.
- Adds only the FSETID capability so setgid directory permissions can persist while TrueNAS continues to drop all other container capabilities.
- Keeps the server image unchanged at ghcr.io/d69wookie/actuallyfreecd-server:0.9.0.

Hardware tested on TrueNAS with a physical optical drive using /dev/sr0 and /dev/sg8.

Verified output permissions:
- Directories: root:media 2775
- Files: root:media 0664

The app renders successfully and passes apps_dev_charts_validate.

### DIFF
--- a/ix-dev/community/actuallyfreecd/questions.yaml
+++ b/ix-dev/community/actuallyfreecd/questions.yaml
@@ -1,6 +1,8 @@
 groups:
   - name: ActuallyFreeCD Server Configuration
     description: Configure ActuallyFreeCD Server
+  - name: Output Permissions Configuration
+    description: Configure ownership and permissions for ripped music files.
   - name: Network Configuration
     description: Configure Network for ActuallyFreeCD Server
   - name: Storage Configuration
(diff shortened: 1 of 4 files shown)


ix-dev/community/actuallyfreecd/questions.yaml
```diff
@@ -75,6 +77,43 @@ questions:
                       schema:
                         type: string
 
+  - variable: output_permissions
```
stavros-k, line 80: We've been already through this. Groups are fixed. But you keep letting an LLM do the work without doing any yourself, so it keeps coming back.

D69wookie:
You're right, and I apologise — I misunderstood your earlier point about
groups being fixed and ended up reintroducing something we'd already
discussed.

To explain what I was actually trying to solve rather than defend the
implementation:

When I first installed ActuallyFreeCD from the Community catalogue and
performed a real CD rip, the output directories/files were being created as
root:root. The music dataset is also used by SMB and my other media
applications, so those files weren't inheriting the group used by the
shared dataset and couldn't be managed properly through the normal
media/SMB workflow.

That is what led me down the path of trying to make the output group
configurable.

During testing I got that working as root:media, with files being created
as 0664. However, the newly created Artist/Album/Disc directories were
still 0775 rather than 2775, so although the group was then correct, the
setgid bit was being lost as ActuallyFreeCD created the directory hierarchy.

Adding only FSETID allowed the application to preserve that bit. I verified
this with another real CD rip: the resulting directories were root:media
2775 and the files root:media 0664.

I understand now that exposing the group as a configurable value isn't the
appropriate TrueNAS Apps solution because the app groups are fixed.

What I'm trying to understand is the catalogue-supported way to solve the
original problem: ActuallyFreeCD needs to create an Artist/Album/Disc
hierarchy in a user-selected host-path music dataset while allowing that
output to remain accessible to SMB and other media applications sharing the
dataset.

Is the intended TrueNAS approach for this to be handled through the
ACL/inheritance configuration on the host-path dataset when the app is
installed, rather than having the application manage the output group
itself?

If so, that's absolutely fine. I'd rather implement this in the way TrueNAS
expects than work around the permissions model.

thanks

Darren

stavros-k:
Once you stop having an LLM respond, we can discuss.

PR #5827, opened 2026-09-17
libredb-studio: update to 0.16.0

yusuf-gundogdu:
Updates LibreDB Studio from 0.15.0 to 0.16.0 and bumps the chart version to 1.0.11.

Only files under ix-dev are touched, the trains directory is left to the catalog job.


stavros-k:
Updates handled automatically.
Thanks

PR #5842, opened 2026-09-18
gitea-act-runner: track renamed upstream image (0.6.1 -> 3.5.0)

FN-Rob:
# gitea-act-runner: track the renamed upstream image (0.6.1 -> 3.5.0)

## Problem

`gitea-act-runner` has shipped act_runner **0.6.1 since 2026-04-30**. It is not
that upstream stopped releasing; the image repository the app points at was
abandoned that day.

Upstream renamed the project in `gitea/runner` commit `13dc938`
("Rename act_runner to runner", #850) on 2026-04-30, and began publishing to a
new Docker Hub repository.

| repository | tags | newest | last push |
|---|---|---|---|
| `gitea/act_runner` (what this app pins) | 77 | **0.6.1** | **2026-04-30** |
| `gitea/runner` (current) | 197 | **3.5.0** | 2026-09-14 |

`gitea/act_runner:latest` still resolves and still pulls, it just returns the
April build, so nothing surfaces as broken.

## Why Renovate cannot catch this

Renovate is tracking the package correctly. The Dependency Dashboard (#304)
lists:

```
ix-dev/community/gitea-act-runner/ix_values.yaml (2)
- `gitea/act_runner 0.6.1`
```

There genuinely are no newer tags under that name, so "no update available" is
the right answer to the question being asked. Advancing requires changing the
repository **name**, which is a rename rather than a version bump. Meanwhile the
chart has kept moving (1.1.14, updated 2026-09-18) while `app_version` stayed at
0.6.1, which makes the app look maintained and current.

Consider a Renovate `replacementName` rule so a future upstream rename produces
a PR instead of silence.

## What this changes

- `ix_values.yaml`: `gitea/act_runner:0.6.1` -> `gitea/runner:3.5.0`
- `templates/docker-compose.yaml`: healthcheck process name (see below)
- `app.yaml`: `app_version`, `changelog_url`, `sources`, `version` 1.1.14 -> 1.2.0

## The part a plain version bump would miss

The healthcheck greps for the process by name:

```jinja
{% do c1.healthcheck.set_test("pgrep", {"process": "act_runner"}) %}
```

The rename also renamed the binary. In `gitea/runner:3.5.0` it is
`/usr/local/bin/gitea-runner`, and `run.sh:68` is `exec gitea-runner daemon`.
Bumping only the image would leave every container permanently unhealthy.
Verified against the image:

```
pgrep act_runner   -> no match
pgrep gitea-runner -> match
```

## Compatibility checks

- **Existing registrations survive.** The `.runner` `Registration` struct is
  byte-identical between v0.6.1 and v3.5.0 (same fields, same JSON tags), so a
  persisted `/data/.runner` is read unchanged. No migration needed.
- **All three env vars this app sets are still honoured** by `run.sh` in 3.5.0:
  `GITEA_INSTANCE_URL`, `GITEA_RUNNER_NAME`, `GITEA_RUNNER_REGISTRATION_TOKEN`.
- **`config.yaml` schema is a superset.** Every key valid in 0.6.1 still exists
  in 3.5.0; nothing was removed. User configs carry over as-is.
- **`HOME=/data`** is unchanged, so the action cache stays on the data volume.
- `lib_version` stays at 2.3.13, already the newest, so no `copy-lib` re-run.

## Why users should care

Among the 26 releases skipped is `cce8543` (2026-05-07, first in v1.0.2), which
fixes a race in the shared action cache: when two concurrent jobs use the same
`uses:`, one job's read can interleave with a peer's checkout and see partial
state, surfacing as `Cannot find module .../dist/index.js`. Anyone running this
app above `capacity: 1` has been exposed to that since April.

## Version bump

`1.1.14 -> 1.2.0` (minor). No breaking change to app config, user values, or
persisted state, but more than a patch. Happy to make it `2.0.0` if you would
rather signal the upstream major jump.


FN-Rob:
@stavros-k mentioning you because of the Renovate replacementName recommendation. If I remember correctly, that's in your wheelhouse (maybe Kris?).

stavros-k (approved):

PR #5843, opened 2026-09-18
fix(renovate): drop custom rustfs versioning, track stable only

tilalx:
# Pull Request

rustfs released 1.0.0 stable. Drop the custom prerelease-aware
versioning for rustfs/rustfs — it now uses Renovate's default docker
semver versioning, so the bot only offers stable tags going forward.


stavros-k (approved):

stavros-k:
switched to stable on the app as well, as renovate won't update it if it doesnt match the versioning

